PR #24, opened 2020-03-04
Set proper file modes for configuration files.

ahesford:
The default mode for files copied by `install` is 0755. Configuration files in /etc/zfsbootmenu are installed with this mode by default. This PR changes the mode for configuration files to 0644.


zdykstra:
Thanks, good catch!

Issue #23, opened 2020-03-04
[FEATURE REQUEST] Forward command-line arguments to "client" kernels


ahesford:
Currently, zfsbootmenu looks in `/etc/default/grub` to find the command-line arguments it passes to "client" kernels it will boot. It might be more convenient and more transparent to forward the arguments passed to the zfsbootmenu kernel. This allows users to specify arguments in the "natural" place: the boot manager that would boot their kernels if zfsbootmenu were not used.

There are a couple of considerations here:
1. Some arguments will need to be culled or replaced when passing to clients (*e.g.*, `zfsbootmenu` or `root` arguments);
2. If zfsbootmenu is using a different kernel version than the one selected for booting, the arguments may not be directly compatible.

The first point may not be that big a deal, because zfsbootmenu already takes responsibility for the `root` argument and removing the specific `zfsbootmenu` argument would be trivial. The second point may be more problematic, since a key feature of boot environments is the booting of different kernel versions. This admits the possibility that the zfsbootmenu and client kernels will have different versions and potentially different sets of supported arguments.

zdykstra:
Work has been started on this in [issue-23](https://github.com/zdykstra/zfsbootmenu/tree/issue-23) . I'm planning on gating this behind `inherit_cmdline=1` on the kernel command line. I'll plan on stripping out `root=*`, `zfsbootmenu=*`, `initrd=*` and `initramfs=*`. I'm not sure if there are any others that can be removed automatically.

ahesford:
With the upcoming move to the `org.zfsbootmenu:commandline` property, I no longer think forwarding command-line arguments from zfsbootmenu "client" kernels is a particularly interesting feature.

PR #30, opened 2020-05-11
Prefer ZFS property org.zfsbootmenu:commandline for command-line args

ahesford:
It bothers me to have `/etc/default/grub` to define command-line arguments when I don't have grub installed. Moving the arguments to the ZFS property `org.zfsbootmenu:commandline` is consistent with the reliance on `org.zfsbootmenu:kernel` for choosing a kernel version, and allows for nice default-and-override semantics that aren't realizable with a filesystem.

This should supersede issue #23, because this seems like a cleaner approach. I only wanted command-line inheritance because I didn't want the grub file. :)

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -105,7 +105,7 @@ kexec_kernel() {
     emergency_shell "unable to mount ${fs}"
   fi
 
-  cli_args="$( find_kernel_args "${mnt}" )"
+  cli_args="$( find_kernel_args "${fs}" "${mnt}" )"
 
   # restore kernel log level just before we kexec
   echo "${printk}" > /proc/sys/kernel/printk
@@ -246,7 +246,7 @@ find_be_kernels() {
   def_kernel_file="${mnt/mnt/default_kernel}"
   echo "${def_version}" > "${def_kernel_file}"
 
-  def_args="$( find_kernel_args "${mnt}" )"
+  def_args="$( find_kernel_args "${fs}" "${mnt}" )"
   def_args_file="${mnt/mnt/default_args}"
   echo "${def_args}" > "${def_args_file}"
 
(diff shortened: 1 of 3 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -285,14 +285,20 @@ select_kernel() {
...
-  local selected_args
+  local selected_args zfsbe_args
 
   if [ -f "${BASE}/default_args" ]
   then
     cat "${BASE}/default_args"
     return
   fi
 
+  zfsbe_args="$( zfs get -H -o value org.zfsbootmenu:commandline ${zfsbe} )"
```
zdykstra, line 296: `${zfsbe}` is perhaps badly named, in this case it's the full path to where the boot environment is mounted. An example call is at https://github.com/zdykstra/zfsbootmenu/blob/master/90zfsbootmenu/zfsbootmenu-lib.sh#L249

`find_kernel_args()` should probably accept a second, optional parameter that's the name of the ZFS file system, like `find_kernel()` takes.

ahesford: OK, the latest push addresses this with the following logic:

1. The BE dataset name is now the first argument and the mountpoint is the second argument; if the property is the preferred source for the arguments, its requirement should come first.

2. The property is only checked if the dataset argument is non-empty.

3. The `etc/default/grub` file is only checked if the mountpoint argument is non-empty.

This should robustly handle all sorts of issues discovering arguments, with the Right Thing (TM) done if the function is provided with both the dataset name and mountpoint. Both calls to this function have been updated to provide the expected arguments.

zdykstra:
I'm going to merge this in now, as everything looks good to go. If concrete testing tonight reveals any issues I'll just fix it up prior to the 1.0rc2 release.

This was an excellent idea - it very elegantly uses a great ZFS feature. Thank you!

PR #34, opened 2020-05-16
[WIP] Use dracut to generate unifiedEFI images

ericonr:
This enables the user to add more configuration options in
/etc/zfsbootmenu/dracut.conf.d, such as the location of Secure Boot keys
to sign the unifiedEFI image.

### DIFF
--- a/bin/generate-zbm
+++ b/bin/generate-zbm
@@ -116,14 +116,16 @@ if ( ( defined $runConf{kernel} ) and ( length $runConf{kernel} ) ) {
 $runConf{bootdir} = dirname( $runConf{kernel} );
 ( $runConf{kernel_prefix}, $runConf{kernel_version} ) = split( '-', basename( $runConf{kernel} ) );
 
-# We always need to create an initramfs
-printf "Creating ZFS Boot Menu %s, with %s %s\n", $runConf{version}, $runConf{kernel_prefix}, $runConf{kernel_version};
+# We need to create an initramfs for all cases other than unifiedEFI
+if ( defined( $config{Components}{Copies} ) and ( $config{Components}{Copies} gt 0 ) ) {
+  printf "Creating ZFS Boot Menu %s, with %s %s\n", $runConf{version}, $runConf{kernel_prefix}, $runConf{kernel_version};
 
-$runConf{initramfs} = createInitramfs( $tempdir, $runConf{kernel_version} );
+  $runConf{initramfs} = createInitramfs( $tempdir, $runConf{kernel_version} );
+}
 
 # Create a unified kernel/initramfs/command line EFI file
 if ( defined( $config{EFI}{Copies} ) and ( $config{EFI}{Copies} gt 0 ) ) {
-  $runConf{unified_efi} = unifiedEFI( $tempdir, $runConf{kernel}, $runConf{initramfs} );
+  $runConf{unified_efi} = unifiedEFI( $tempdir, $runConf{kernel_version} );
 
   if ( defined( $config{EFI}{Versioned} ) and ( $config{EFI}{Versioned} ) ) {
     $runConf{efi_target} =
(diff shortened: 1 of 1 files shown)


zdykstra:
Thank you for the PR - this is a great idea. I should have time to test this locally tomorrow.  The logic changes / additions look fine at first glance.

ericonr:
Glad to help! I have tested this locally, and booted from it (with Secure Boot).

bin/generate-zbm
```diff
@@ -116,14 +116,16 @@ if ( ( defined $runConf{kernel} ) and ( length $runConf{kernel} ) ) {
 $runConf{bootdir} = dirname( $runConf{kernel} );
 ( $runConf{kernel_prefix}, $runConf{kernel_version} ) = split( '-', basename( $runConf{kernel} ) );
 
-# We always need to create an initramfs
-printf "Creating ZFS Boot Menu %s, with %s %s\n", $runConf{version}, $runConf{kernel_prefix}, $runConf{kernel_version};
+# We need to create an initramfs for all cases other than unifiedEFI
```
ericonr, line 119: If you add another type of output (besides component and unified), we'd need to add another case here. Do you think there's another way to write this conditional that is more general?

bin/generate-zbm
```diff
@@ -116,14 +116,16 @@ if ( ( defined $runConf{kernel} ) and ( length $runConf{kernel} ) ) {
...
+if ( defined( $config{Components}{Copies} ) and ( $config{Components}{Copies} gt 0 ) ) {
```
ericonr, line 120: If you add another type of output (besides component and unified), we'd need to add another case here. Do you think there's another way to write this conditional that is more general?

zdykstra:
> Glad to help! I have tested this locally, and booted from it (with Secure Boot).

Can you outline how it works with Secure Boot? I'm not familiar with how that works, and it'd be nice to document that once it's possible to do.

bin/generate-zbm
zdykstra: In the event that another output type is added, I can see if there's a nicer way to write that out. For now though it's reasonably easy to read so I'm inclined to leave it as-is.

ericonr:
Secure Boot requires motherboard support to set your own keys. I believe there's an EFI application that you can boot into to add new ones without official firmware support, but there have been cases of it bricking laptops.

There is an Arch Wiki page for it (and I don't remember which keys you need to configure the firmware), but since you already recommend rEFInd, it can generate keys by itself (not sure if this is enough, I can give it a try later). Just install it with `--localkeys` and it will create them and sign itself before installation. If you already have keys, you have to put them inside `/etc/refind.d/keys/` and name them `refind_local.{crt,cer,key}`. Once you have that, you can just put it in a dracut config file (or via command line, if I'm not mistaken) like

```
uefi_secureboot_cert="/etc/refind.d/keys/refind_local.crt"
uefi_secureboot_key="/etc/refind.d/keys/refind_local.key"
```

And dracut will use them if you pass it the `--uefi` option.

dm17:
> Secure Boot requires motherboard support to set your own keys. I believe there's an EFI application that you can boot into to add new ones without official firmware support, but there have been cases of it bricking laptops.
> 
> There is an Arch Wiki page for it (and I don't remember which keys you need to configure the firmware), but since you already recommend rEFInd, it can generate keys by itself (not sure if this is enough, I can give it a try later). Just install it with `--localkeys` and it will create them and sign itself before installation. If you already have keys, you have to put them inside `/etc/refind.d/keys/` and name them `refind_local.{crt,cer,key}`. Once you have that, you can just put it in a dracut config file (or via command line, if I'm not mistaken) like
> 
> ```
> uefi_secureboot_cert="/etc/refind.d/keys/refind_local.crt"
> uefi_secureboot_key="/etc/refind.d/keys/refind_local.key"
> ```
> 
> And dracut will use them if you pass it the `--uefi` option. Btw, kudos to their man pages. Got it all from there.

You said "before installation," but I'm not sure what installation this is before.
Can you give a bit more info on where this fits into zdykstra's guide here:
https://github.com/zdykstra/zfsbootmenu/blob/master/void-install.md
Thanks.

Issue #41, opened 2020-06-16
List of boot environments needs proper refresh after emergency shell


ahesford:
In the current implementation, a list of boot environments is written to `${BASE}/env` before entering the menu input loop in `zfsbootmenu.sh`. Functions that are expected to alter this list, like `clone_snapshot` and (in the `send-recv` branch for now) `duplicate_snapshot`, modify the list by appending to `${BASE}/env` when new boot environments are created.

If the user drops to an emergency shell with `Alt+R` and renames, creates or deletes BEs manually, `${BASE}/env` will be out of sync with reality. If the generation of `${BASE}/env` is moved to a function, complete with a drop to an emergency shell if no boot environments are found, there are two good possibilities for addressing this issue:
1. Call this function in every iteration of the menu input loop. This might slow things down a bit and is often unnecessary, but the list will always be accurate and there will be no need for `clone_snapshot` or `duplicate_snapshot` to update the list when they add a new boot environment.
2. Leave the `{clone,duplicate}_snapshot` functions to update the list as they do no, and only call the BE-list-update function after entering an emergency shell. This will be more efficient, but might not be the most proper approach for some definition of proper.

zdykstra:
I'm inclined to try the first option and get a feel for added latency while navigating the menus. If it proves to seem a little slow, the function can be called only when leaving something that can possibly make changes (snapshot screen, recovery shell).

People with a small number of boot environments and kernels will be the least impacted by calling it every redraw - while those with a lot of BEs and kernels will be the most impacted.

ahesford:
Addressed on the `rescan` branch in #42.

ahesford:
#42 merged.

PR #42, opened 2020-06-17
Rescan BEs and kernels with every iteration of menu loop

ahesford:
Implements a solution to #41 but needs a more testing before merging to master.

This also required reworking the kernel detection logic to avoid continually appending the same kernels to the list of available versions. (The kernel detection is a second commit to facilitate comparison and evaluation.)

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -219,7 +219,7 @@ set_default_env() {
 # arg1: ZFS filesystem
 # arg2: mountpoint
 # prints: nothing
-# returns: 0 if kernels were found
+# returns: 0 if kernels were found, 1 otherwise
 
 find_be_kernels() {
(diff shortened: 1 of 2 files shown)


zdykstra:
I'll give this a run-through tonight! How does the lag feel when going in and out of snapshot/kernel/command line screens?

ahesford:
I don't notice a difference. In my experience, `zfs list` is really only slow if you ask it to show snapshots and you have a bunch of them.

zdykstra:
Everything here looks great. I've run this branch through a few boot cycles on my test laptop.

* Duplicating a boot environment works as expected
* Cloning a snapshot works as expected
* Cloning and promoting a snapshot works as expected

Other boot modes (menu countdown, immediate boot to bootfs+kernel) work as expected.  After creating a number of new BEs via send|recv and cloning a snapshot, I didn't notice any appreciable change in the time it takes to draw the main menu.

PR #44, opened 2020-06-18
[WIP] Run shellcheck on any push

zdykstra:
`shellcheck` can help catch a number of small issues, and otherwise point out minor improvements. This commit adds an Actions config to execute `shellcheck` on a push to any branch. 

`modules-setup.sh` and `zfsbootmenu-parse-commandline.sh` have a number of problems that will need to be fixed before this can be merged, so we have a passing badge!


zdykstra:
Thanks for the help getting master into shape, @ahesford

Issue #47, opened 2020-06-19
Kernel version selector should have "set default" abilities


ahesford:
In the kernel selection menu invoked by `ALT+K` for a specific boot environment, there should be an `ALT+D` hotkey that pins the default kernel version for that environment, much like `ALT+D` at the top level sets a default boot environment. This amounts to setting the pool writable and setting the `org.zfsbootmenu:kernel` property on the current boot environment.

zdykstra:
I can tackle this tonight, unless you have designs on it.

ahesford:
Go ahead; I have nothing started.

zdykstra:
Fixed in #48 !

PR #48, opened 2020-06-19
Initial support for setting default kernel versions

ahesford:
This implements the logic outlined in #47 and seems to work as expected, but should be tested a bit more before merging. I take the kernel path identified from `draw_kernel` and drop the `/boot/` prefix to write the full kernel file name to the `org.zfsbootmenu:kernel` property of the associated BE.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -70,11 +70,11 @@ draw_kernel() {
 
   benv="${1}"
 
-  selected="$( fzf --prompt "${benv} > " --tac \
-    --with-nth=2 --header="[ENTER] boot [ESC] back" < "${BASE}/${benv}/kernels" )"
-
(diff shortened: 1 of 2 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -219,6 +219,40 @@ clone_snapshot() {
...
+
+  pool="${fs%%/*}"
+  [ -n "${pool}" ] || return 1
+
+  # Strip /boot/ to list only the file
+  kernel="${2#/boot/}"
+
+  # Make sure the pool is writable
+  if set_rw_pool "${pool}"; then
+    key_wrapper "${pool}"
```
zdykstra, line 241: `key_wrapper` should have  `CLEAR_SCREEN=1` set before it's called, and `CLEAR_SCREEN=0` set after. If this isn't done, the password input prompt appears on the bottom of the screen, where fzf leaves the cursor for input.

This is a giant hack, but a global variable was the only reasonable way to make this happen.

PR #50, opened 2020-06-24
Show diff of snapshot against the current filesystem

zdykstra:
When figuring out which snapshot to boot to fix a problem, it can be useful to know which files have changed. When browsing snapshots, there's now an ALT-D key handler which will call the `show_diff()` function. This function takes a single argument, the name of a snapshot. The pool and origin filesystem of the snapshot are determined via bash variable manipulation.  The diff shows all the changes between the snapshot selected and the current state of the fileystem.

To handle `zfs diff` not exiting when the `fzf` browser is exited early, the pid of the process is captured and then killed when fzf exits. The pool is mounted read-write because the `zfs diff` process requires it. 

`BE_SELECTED` is reset to 1, so that you return to the list of snapshots for that boot environmentt, instead of all the way back to the boot environment list.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -88,12 +88,47 @@ draw_snapshots() {
   benv="${1}"
 
   selected="$( zfs list -t snapshot -H -o name "${benv}" |
-    fzf --prompt "Snapshot > " --tac --expect=alt-x,alt-c \
-        --header="[ENTER] duplicate [ALT+X] clone and promote [ALT+C] clone only [ESC] back" )"
+    fzf --prompt "Snapshot > " --tac --expect=alt-x,alt-c,alt-d \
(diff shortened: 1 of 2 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -88,14 +88,49 @@ draw_snapshots() {
   benv="${1}"
 
   selected="$( zfs list -t snapshot -H -o name "${benv}" |
-    fzf --prompt "Snapshot > " --tac --expect=alt-x,alt-c \
-        --header="[ENTER] duplicate [ALT+X] clone and promote [ALT+C] clone only [ESC] back" )"
+    fzf --prompt "Snapshot > " --tac --expect=alt-x,alt-c,alt-d \
+        --header="[ENTER] duplicate [ALT+X] clone and promote [ALT+C] clone only [ALT-D] show diff [ESC] back" )"
```
ahesford, line 92: Header should be `ALT+D` instead of `ALT-D` for consistency.

zdykstra: Good catch, fixed!

PR #55, opened 2020-07-05
Improve kernel/initramfs detection in dracut module, support unversioned kernels in generate-zbm

### LINKED ISSUE #20: Handle kernel files without a version suffix
Arch and derivatives do not add a version suffix to their kernels in /boot. This breaks generate-zbm when attempting to set `kver`.  Logic to handle this use case will have to be added - possibly by just not appending `--kver` to the dracut command line.

ahesford:
The dracut module now looks for a much broader range of acceptable names in pairs of kernels and initramfs/initrd images it will identify in boot environments. These improvements should allow zfsbootmenu to properly identify BEs from Arch Linux and its derivatives that do not include proper version information in kernel naming.

The `generate-zbm` script now accepts the `--kver` argument that should allow image creation on systems like Arch Linux where the version number cannot otherwise be reliably detected. This should fix #20. The `--prefix` argument is an added bonus that allows the prefix of the output zfsbootmenu components or UEFI bundle to be specified, which can improve handling in some (imagined) cases where kernel naming is even more severely broken than on Arch.

**Note**: this was rebased from earlier versions of this branch to include the `safeCopy` revert and subsequent behavior change in `generate-zbm`.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -322,8 +322,8 @@ find_be_kernels() {
   fs="${1}"
 
 
-  local kernel version kernel_records
-  local defaults def_kernel def_version def_kernel_file def_args def_args_file
+  local kernel kernel_base labels version kernel_records
+  local defaults def_kernel def_kernel_file def_args def_args_file
 
   # Check if /boot even exists in the environment
   mnt="$( mount_zfs "${fs}" )"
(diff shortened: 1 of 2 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -338,22 +338,32 @@ find_be_kernels() {
...
+    [ -e "${mnt}/boot/${kernel}" ] || continue
+
+    # Kernel "base" extends to first hyphen
+    kernel_base="${kernel%%-*}"
+    # Kernel "version" is everything after base and may be empty
+    version="${kernel#${kernel_base}}"
+    version="${version#-}"
+
+    # initramfs images can take many forms, look for a sensible one
+    labels=( "$kernel" )
```
zdykstra, line 354: What distribution does some flavor of `initramfs-linux.img` (or whatever this ends up matching in the big initramfs check defined later) ?

bin/generate-zbm
```diff
@@ -115,24 +121,44 @@ if ( ( defined $config{Global}{BootMountPoint} ) and ( length $config{Global}{Bo
...
+  # Try to determine a kernel file when one was not provided
+  if ( ! nonempty $runConf{kernel_version} ) {
+    $runConf{kernel} = latestKernel;
+  } else {
+    $runConf{kernel} = versionedKernel $runConf{kernel_version};
+  }
+
+  # Make sure a kernel was found
+  unless ( nonempty $runConf{kernel} ) {
+    print "Unable to choose a kernel file, cannot continue\n";
```
zdykstra, line 134: Should this have an exit value of 1, since it's a failure?

bin/generate-zbm
```diff
@@ -115,24 +121,44 @@ if ( ( defined $config{Global}{BootMountPoint} ) and ( length $config{Global}{Bo
...
-$runConf{bootdir} = dirname( $runConf{kernel} );
-
-if ( basename( $runConf{kernel} ) =~ m/(vmlinux|vmlinuz|linux|kernel)-(\S+)/i ) {
-  $runConf{kernel_prefix} = $1;
-  $runConf{kernel_version} = $2;
-} else {
-  printf "Unable to determine kernel prefix and version from %s", $runConf{kernel};
-  exit;
+# Try to determine kernel_prefix or kernel_version if necessary
+if ( ! ( nonempty $runConf{kernel_prefix} and nonempty $runConf{kernel_version} ) ) {
```
zdykstra, line 146: I wouldn't mind seeing `if ( ! ( ...` normalized to `unless ( ...`, just so the pattern is consistent throughout the whole tool.

90zfsbootmenu/zfsbootmenu-lib.sh
ahesford: It's been awhile since I ran Arch, but I'm pretty sure the default initramfs image for that distribution takes the forms
```
initramfs-linux.img
initramfs-linux-lts.img
initramfs-linux-zen.img
...
```
where the corresponding kernsl are `linux`, `linux-lts`, `linux-zen`, etc.

The [Arch wiki](https://wiki.archlinux.org/index.php/Mkinitcpio#Troubleshooting) seems to confirm this.

bin/generate-zbm
ahesford: That's probably a good idea. I added a few other nonzero exit codes for error conditions for consistency.

bin/generate-zbm
ahesford: That's reasonable, done.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Ahh, good to know. Would it make more sense then to reverse the ordering and set `labels=( "$version" )` first and then push `$kernel` onto the array? This way the Void naming pattern will be matched early in the cycle, rather than later.

ahesford: The advantage of keeping `$kernel` before `$version` is that it allows for kernel-specific images even when the kernels have the same version. In this totally contrived example, let's assume a user has installed `vmlinuz-5.7.7_1` and `linux-5.7.7_1` with different configurations. If we favor the default Void naming, then `initramfs-5.7.7_1.img` will be preferred for both kernels. However, if the user wants custom behavior for `linux-5.7.7_1`, preferring the `$kernel` suffix allows creation of `initramfs-linux-5.7.7_1.img` that will take priority with that kernel.

To improve matters quite a bit, I've unrolled most of the brace expansions into nested `for` loops that adjust the loop priorities. Now the Void default will hit third rather than the ~30th place (!!) it was hitting with the original expansion.

PR #56, opened 2020-07-06
Allow root prefix to be customized for other distributions

### LINKED ISSUE #21: Build command line for other systems
Obarun (Arch) is unable to boot, because they use a slightly different zfs kernel command line. 

`root=zfs=obarun/ROOT/master`

/etc/os-release
``` 
NAME="Obarun Linux"
PRETTY_NAME="Obarun Linux"
ID=obarun
ID_LIKE=arch
ANSI_COLOR="0;36"
HOME_URL="https://web.obarun.org/"
DOCUMENTATION_URL="https://wiki.obarun.org/"
SUPPORT_URL="https://forum.obarun.org/"
BUG_REPORT_URL="https://forum.obarun.org/"
```

ahesford:
Fixes #21.

When creating a `root=` kernel command-line, kexec_kernel now defers to a function that will attempt to determine the appropriate prefix to prepend to the ZFS dataset name. If the `org.zfsbootmenu:rootprefix` property is defined on the BE, that will be preferred; otherwise, if `/usr/lib/os-release` or `/etc/os-release` are available, those will be queried for OS identifiers will be searched for a preexisting default. If no suitable values are discovered, a default of "zfs:" is used.

**Needs testing**, especially on non-Void distributions.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -167,6 +167,7 @@ kexec_kernel() {
   fi
 
   cli_args="$( find_kernel_args "${fs}" "${mnt}" )"
+  root_prefix="$( find_root_prefix "${fs}" "${mnt}" )"
 
   # restore kernel log level just before we kexec
   # shellcheck disable=SC2154
@@ -174,7 +175,7 @@ kexec_kernel() {
 
   kexec -l "${mnt}${kernel}" \
     --initrd="${mnt}${initramfs}" \
-    --command-line="root=zfs:${fs} ${cli_args}"
+    --command-line="root=${root_prefix}${fs} ${cli_args}"
 
   umount "${mnt}"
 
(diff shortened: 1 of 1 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -424,6 +425,57 @@ select_kernel() {
...
+      unset ID ID_LIKE
+      for osrel in ${zfsbe_mnt}/{usr/lib,etc}/os-release; do
+        if [ -f "${osrel}" ]; then
+          # shellcheck disable=SC1090
+          . "${osrel}" >/dev/null 2>&1
+        fi
+      done
+
+      for ostype in $ID $ID_LIKE; do
+        case "$ostype" in
```
zdykstra, line 458: Maybe add the Void and Ubuntu/Debian types here so they're explicitly documented. Ubuntu/Debian seem to be able to use the Dracut method of `zfs:`- https://github.com/openzfs/zfs/blob/master/contrib/initramfs/scripts/zfs#L758-L766

ahesford: That's resonable; done.

PR #61, opened 2020-07-15
Add configuration options for kernel, version and prefix

ahesford:
Support new kernel options in the config file. Arch users, for example, should be able to add
```
[Kernel]
Path=/boot/vmlinuz-linux
Version=/current
```
and use `generate-zbm` without arguments to produce bootmenu images. Note I use `/current` instead of `current` because it's conceivable somebody may specify a Linux kernel with version string `current` and we don't want to conflict. Obviously `/current` is not a valid version name because the modules directory could not exist for that version (which means `dracut` would fail independently).

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,9 @@ Copies=0
 
 ### Kernel
 * `CommandLine` If you're making a unified EFI file, this is the command line passed to the module. Refer to [Command line options](README.md#command-line-options).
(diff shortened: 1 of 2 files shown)


zdykstra:
I overall like this, though to me using `/current` feels weird. Are there any other magic characters that could replace `/` ?

ahesford:
Merged in 5fdb8729

ahesford:
Decided to go with `%current` instead.

PR #63, opened 2020-08-09
proposed.yaml: Use better defaults, update README.

ericonr:


README.md
```diff
@@ -195,20 +195,20 @@ An example YAML configuration file follows:
...
 ---
 Global:
-  ManageImages: true
+  ManageImages: false
   BootMountPoint: /boot/efi
   DracutConfDir: /etc/zfsbootmenu/dracut.conf.d
 Components:
   ImageDir: /boot/efi/EFI/void
-  Versions: false
+  Versions: 2
```
ahesford, line 203: In the README, let's keep `Versions: false` here because it provides contrast with the `Versions: 2` in the `EFI` section.

ericonr: Good?

PR #64, opened 2020-08-09
Yaml conversion

ahesford:
I think this is ready for a close look and merge to master. Afterwards, we can do a bit more testing, including on some Project Trident config, and contemplate a new release to take advantage of the improved backup and kernel selection logic.

Barring objections, I intend to squash these commits when the merge is ready. @ericonr, please let me know if you find other critical issues.

### DIFF
--- a/bin/generate-zbm
+++ b/bin/generate-zbm
@@ -36,6 +36,7 @@ sub safeCopy;
 sub nonempty;
 sub cleanupMount;
 sub enabled;
+sub maxRevision;
 sub convertConfig;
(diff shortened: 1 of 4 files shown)


bin/generate-zbm
```diff
@@ -3,7 +3,7 @@
 use strict;
 use warnings;
 
-our $VERSION = '1.3';
+our $VERSION = '1.4';
```
ericonr, line 6: Should the version change happen here?

PR #65, opened 2020-08-11
Add explicit --migrate option to generate-zbm

ahesford:
Conversion from INI to YAML is no longer fully automatic, but is instead accomplished by specifying `--migrate` with an optional path to the INI file to migrate. Combined with `--config`, this allows totally customizable migration sources and targets.

`generate-zbm` always terminates immediately after attempted migration, with the exit code indicating success (zero) or failure (nonzero).

This provides a simple mechanism to migrate INI to YAML in XBPS package upgrades:
```
if [ -f /etc/zfsbootmenu/config.ini ] && generate-zbm --migrate; then
	mv /etc/zfsbootmenu/config.ini /etc/zfsbootmenu/config.ini.migrated
fi
```
This will even leave a backup copy of any existing YAML at `/etc/zfsbootmenu/config.yaml.backup`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,12 +189,9 @@ Your distribution should have packages for these already.
 
 ## config.yaml
 
-The YAML file `/etc/zfsbootmenu/config.yaml` is used to control the behavior of `generate-zbm`. In prior versions, an INI file was used; `generate-zbm` will convert an existing `config.ini` file if possible when no `config.yaml` is found.
-
-An example YAML configuration file follows:
+The YAML file `/etc/zfsbootmenu/config.yaml` is used to control the behavior of `generate-zbm`. An example YAML configuration file follows:
 
 ```
----
 Global:
   ManageImages: false
   BootMountPoint: /boot/efi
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -241,6 +238,23 @@ Kernel:
...
+
+```
+generate-zbm --migrate [ini-config] [--config yaml-config]
+```
+
+By default, the output YAML will be written to `/etc/zfsbootmenu/config.yaml`; use the `--config` argument to customize the output location.
+
+The argument `[ini-config]` to `--migrate` is optional. When it is not provided, `generate-zbm` will derive an input file by dropping the `.yaml` extension from the output file and appending a `.ini` extension.
+
+As a special case, if and only if `generate-zbm` is run without a `--config` option (*i.e.*, it attempts to load the default `/etc/zfsbootmenu/config.yaml`) and the default configuration does *not* exist. Under these circumstances, `generate-zbm` will behave as if it were passed the `--migrate /etc/zfsbootmenu/config.ini` option.
```
ericonr, line 254: "There is a special case: if and ..."

PR #66, opened 2020-08-12
Fix job control in recovery shell

zdykstra:
The original PR used the `cttyhack` module from `busybox`.  I've switched this to use `setsid` and some input/output redirection as noted from the URL @Piraty listed. This will enable this to be effectively the default on all installs, since every system should have `setsid` in the base install.


Piraty:
https://busybox.net/FAQ.html#job_control

zdykstra:
Better handled in https://github.com/zdykstra/zfsbootmenu/tree/skim_support , closing this out in favor of that branch.

PR #69, opened 2020-08-19
Prune everything not explicitly needed

zdykstra:
Various binaries were added over the last year, and some were no longer needed. Those binaries that were not referenced in any file in 90zfsbootmenu were removed.

Other binaries (ip, df, etc) are not explicitly needed for the operation of the module. These should be optionally installed via a so-called 'recovery mode' configuration file.

The cut binary was only used to manipulate a hostid. Those routines were switched to bash string manipulation methods.

### DIFF
--- a/90zfsbootmenu/module-setup.sh
+++ b/90zfsbootmenu/module-setup.sh
@@ -79,9 +79,7 @@ install() {
   dracut_install /usr/lib/udev/vdev_id
   dracut_install /usr/lib/udev/zvol_id
   dracut_install tac
-  dracut_install awk
   dracut_install basename
-  dracut_install cut
   dracut_install head
   dracut_install kexec
   dracut_install fzf
@@ -89,17 +87,10 @@ install() {
   dracut_install sort
   dracut_install sed
   dracut_install grep
-  dracut_install xargs
-  dracut_install clear
-  dracut_install reset
-  dracut_install lsblk
-  dracut_install cut
   dracut_install tput
   dracut_install mount
-  dracut_install df
-  dracut_install ip
-  dracut_install /usr/bin/mkdir
-  dracut_install /usr/bin/tail
+  dracut_install mkdir
+  dracut_install tail
   dracut_install mbuffer
   dracut_install tr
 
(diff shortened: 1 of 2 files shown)


90zfsbootmenu/module-setup.sh
```diff
@@ -79,27 +79,17 @@ install() {
...
-  dracut_install awk
   dracut_install basename
-  dracut_install cut
   dracut_install head
   dracut_install kexec
   dracut_install fzf
   dracut_install mktemp
   dracut_install sort
   dracut_install sed
-  dracut_install grep
```
ahesford, line 91: I've saved myself a few times using `grep` with context in place of a real pager, so I'd advocate keeping it in here. That being said, `less` appears to be in my boot image and installed by `99base`, so I'm not sure why I couldn't access it from the emergency shell.

90zfsbootmenu/module-setup.sh
```diff
@@ -121,11 +111,7 @@ install() {
...
   # Synchronize initramfs and system hostid
-  AA=$(hostid | cut -b 1,2)
-  BB=$(hostid | cut -b 3,4)
-  CC=$(hostid | cut -b 5,6)
-  DD=$(hostid | cut -b 7,8)
-
+  HOSTID="$( hostid )"
   # shellcheck disable=SC2154
-  echo -ne "\\x${DD}\\x${CC}\\x${BB}\\x${AA}" > "${initdir}/etc/hostid"
+  echo -ne "\\x${HOSTID:6:2}\\x${HOSTID:4:2}\\x${HOSTID:2:2}\\x${HOSTID:0:2}" > "${initdir}/etc/hostid"
```
ahesford, line 116: This is reasonable, if bash-only; we've already embraced bashisms like arrays where needed and it saves us the `cut`. I assume you've double-checked the ordering, even though it looks correct from here.

90zfsbootmenu/module-setup.sh
zdykstra: That's reasonable - I've added it back in.

90zfsbootmenu/module-setup.sh
zdykstra: I tested it by moving /etc/hostid out of the way, creating a new EFI bundle and setting `spl_hostid` on the command line / booting the bundle. The pool was able to be imported so it seems accurate. I can do another check tonight to make sure I'm remembering things correctly.

PR #77, opened 2020-08-25
Add support for pool status and checkpoints

zdykstra:
- On the main boot menu screen, add a hook for ALT+P which shows a list
  of discovered/imported pools.

- On the pool health screen, add key shortcuts for Rewinding a
  checkpoint or Deleting a checkpoint. Since the fzf/sk header is a
  static string at the bottom of the menu, it is shown regardless of a
  pool checkpoint being present or not.

- If a checkpoint is to be deleted, use fzf to prompt the user Yes/No -
  defaulting to No. The pool is re-imported read/write and then the
  checkpoint is deleted.

- If a checkpoint is to be rewound, use fzf to prompt. The since a
  rewind operation has to take place on pool import, export the pool and
  then set the --rewind-to-checkpoint flag on the zpool import.

- set_rw_pool() does double duty - it guards against importing the pool
  if a resume image is found, and then it re-imports the pool R/W. The
  resume guard functionality was broken out into a distinct function so
  that it can be called before an alternate export/re-import process is
  used.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -57,9 +57,9 @@ draw_be() {
   test -f "${env}" || return 130
 
   selected="$( ${FUZZYSEL} -0 --prompt "BE > " \
-    --expect=alt-k,alt-d,alt-s,alt-c,alt-r \
+    --expect=alt-k,alt-d,alt-s,alt-c,alt-r,alt-p \
     --preview-window="up:${PREVIEW_HEIGHT}" \
-    --header="[ENTER] boot [ALT+K] kernel [ALT+D] set bootfs [ALT+S] snapshots [ALT+C] cmdline" \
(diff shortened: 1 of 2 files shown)


ahesford:
I suggest two changes:
1. The comments before `delete_checkpoint` and `rewind_checkpoint` have traling spaces; remove.
2. In `import_pool`, I have two tests:
```
[ "${force_import}" ]
[ ${read_write} ]
````
while you've added
```
[ "${rewind_to_checkpoint}" ]
```
For consistency with the rest of `zfsbootmenu-lib.sh` I think the tests should be `[ -n ... ]`.

A *possible* change would be to modify the read-only check in `set_rw_pool` to allow, *e.g.*, a `force_export` environment variable:
```
[ -n "${force_export}" ] || [ "x$( zpool get -H -o value readonly "${pool}" )" = "xon" ] || return 0
```
Then, instead of duplicating the `resume_prompt` and `export_pool` logic in `rewind_checkpoint`, you can just do
```
rewind_to_checkpoint=yes force_export=yes set_rw_pool "${pool}"
```
I still like pulling `resume_prompt` into its own function, because I might hack on that if I ever get around to adding swap-clearing capabilities instead of just warning the user about possible issues.

PR #78, opened 2020-08-27
Support automatic "noresume" when importing pools R/W

ahesford:
Rather than just warn the user about the need to set the "noresume" kernel argument when booting after an active suspend partition is detected and a pool is imported R/W, ZFSBootMenu now supports rewriting the kernel command line to remove any "resume=" arguments and append a "noresume". This uses `awk` to cleanly (and, it seems, correctly) handle the possibility of double-quoted strings in the argument list, which the kernel docs say are permitted. (The docs do not mention single-quoted strings, so we don't support them.) I was able to get an almost-functional implementation with nothing but `sed`, but it was not capable of properly ignoring the `resume=` argument as the quoted value of another argument, like `somearg="a b resume=/dev/sda c d"`. If we *really* don't care about such strange cases, we might be able drop `awk` and just do
```
sed -e 's/\([[:space:]]\+\|^\)resume=\([^[:space:]"]*\("[^"]*"\)*\)*[[:space:]]*/\1/g' <<<"$cmdline"
```
but, given the importance of kernel arguments, we probably want to strive for correctness here.

The prior behavior has been preserved by allowing the user to type DANGEROUS instead of the redefined NORESUME response to the resume guard prompt, just in case there is some unexpected edge case that thwarts automatic redirection and the user wants to assume responsibility for pool corruption.

The inline `awk` script needs some stress testing, but I've thrown a few challenging examples at it and haven't caught any unexpected failures.

### DIFF
--- a/90zfsbootmenu/module-setup.sh
+++ b/90zfsbootmenu/module-setup.sh
@@ -79,6 +79,7 @@ install() {
     "tr"
     "tac"
     "blkid"
+    "awk"
   )
 
   for _exec in "${essential_execs[@]}"; do
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -188,43 +188,6 @@ kexec_kernel() {
   cli_args="$( find_kernel_args "${fs}" "${mnt}" )"
   root_prefix="$( find_root_prefix "${fs}" "${mnt}" )"
 
-  if [ -e "${BASE}/noresume" ]; then
-    # Must replace resume= arguments and append a noresume
-    cli_args="$( awk <<<"${cli_args}" '
-      BEGIN {
-        quot = 0;
-        supp = 0;
-        ORS = " ";
-      }
-
-      {
-        for (i=1; i <= NF; i++) {
-          if ( quot == 0 ) {
-            # If unquoted, determine if output should be suppressed
-            if ( $(i) ~ /^resume=/ ) {
-              # Argument starts with "resume=", suppress
-              supp = 1;
-            } else {
-              # Nothing else is suppressed
-              supp = 0;
-            }
-          }
-
-          # If output is not suppressed, print the field
-          if ( supp == 0 && length($(i)) > 0 ) {
-            print $(i);
-          }
-
-          # If an odd number of quotes are in this field, toggle quoting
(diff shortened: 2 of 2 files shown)


zdykstra:
I've re-read how kernel arguments are both discovered and then used/consumed by `zfsbootmenu-lib.sh` and `zfsbootmenu-preview.sh`. Making it so that `zfsbootmenu-preview.sh` can be made aware of `noresume` being set on the KCL is beyond the scope of this PR.

ahesford:
I attempted to move the `awk` manipulations to `find_kernel_args`, but couldn't get past consistency issues that made it unsuitable for merging. I've instead merged the original PR (with a fixup to avoid using `touch` to create the `noresume` trigger, because `touch` isn't included in the initramfs) and left the incomplete work on the `noresume` branch with commit https://github.com/zdykstra/zfsbootmenu/commit/2fc91461f1d3c99e0527e84f7abcf6f075f1edfe.

My thinking was that the `awk` transformation was idempotent, so we could keep redoing the transformation with every call and not care. This is *almost* true (`find_kernel_args` will currently append an extra `noresume` argument with each invocation, but this is a trivial fix and shouldn't change kernel behavior anyway). When booting a kernel, `kexec_kernel` generally gets the arguments it needs from the new `find_kernel_args`.

The problem is that the ZBM preview window and the command-line editor both rely on cached args at `${BASE}/default_args` and `${BASE}/${selected_be}/default_args`. If these files were created before the pool was mounted R/W and the resume guard triggered `noresume`, the preview and the KCL editor will show outdated information. What's worse, if the user edits the KCL, this will be cached as written, but subsequent calls to `find_kernel_args` (like in `kexec_kernel`) will re-apply the `noresume` transformation.

Much of this could probably be cleaned up significantly just be making `find_kernel_args` a standalone helper script that can be called where needed, including in the KCL editor and the preview window. Alternatively, the preview script could just source `zfsbootmenu-lib.sh` and use the built-in functions to get the properly transformed argument list.

This does raise one important issue. Assuming the preview is fixed to show the transormed KCL, if the user selects `NORESUME` when mounting a pool R/W, then edits the KCL to remove the `noresume` argument and re-add a `resume=` argument, should the transformation be reapplied in the spirit of safety implied by the resume guard text, or should we assume the user is fully aware of the ramifications of making such changes and allow the modifications to supersede the `noresume`?

I guess I opt for the former: always re-transform to protect the user from accidental, massively destructive foot shooting.

PR #80, opened 2020-09-10
[WIP] Enable the testing of changes in QEMU for rapid development/debugging

zdykstra:
* `sudo ./setup.sh`
1.  Creates a 1GB RAW image file
2.  Attaches it to the `loop0` loopback device
3. Creates a GPT label, and a ZFS pool named `ztest`
4. Installs Void base-minimal onto the pool
5. Configures the OS, sets the bootfs, and sets the root password to `zfsbootmenu`

* `./run.sh`
The following defaults can be set to a local default in `.config`:

```
DRIVE="format=raw,file=zfsbootmenu-pool.img"
INITRD="initramfs-bootmenu.img"
MEMORY="2048M"
SMP="2"
DISPLAY="gtk"
APPEND="loglevel=7 timeout=5 zfsbotmenu:POOL=ztest"
```

`run.sh` optionally takes a `-a` argument to override the kernel command line for ZFSBootMenu (`APPEND`).

### DIFF
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -1,0 +1,7 @@
+*.img
+*.qcow2
+*.raw
+vmlinuz*
+vmlinux*
+local.yaml
+.config
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,30 @@
+# Creating a ZFSBootMenu Test Pool for QEMU
+
+First, run (as root) `./setup.sh` to:
+1. Create a 1GB RAW image file,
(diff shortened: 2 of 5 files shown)


ahesford:
I think this looks great just from reviewing the scripts, but I haven't run through the procedure yet. If you really want to wait for me to run through it, I'll try to take a look soon. Otherwise, merge what you have and we can shake out any remaining issues when we actually start testing with it.

A README derived from your PR comment was pushed so I don't have to dig up the instructions later. Rework it or drop it if you like.

PR #82, opened 2020-09-22
Further compartmentalize testing

zdykstra:
The current testing tooling utilizes system-wide Dracut and generate-zbm configuration files. This is an attempt to entirely decouple testing/ from the host OS.

* Create a modules.d directory with symlinks to system-wide Dracut modules, and `90zfsbootmenu` from the git checkout
* Add a new configuration option to `config.yaml` / `generate-zbm` allowing any additional arguments to be passed to Dracut. 
* Copy `config.yaml` and configure it for local use
* Create a new `dracut.conf.d` directory in place of `/etc/zfsbootmenu/dracut.conf.d` and populate it with a default `zfsbootmenu.conf` configuration file
* Add a help section to `setup.sh`, along with handlers for individual pieces (config.yaml, dracut.conf.d, modules.d, zpool image)

The current flaw with this is that `./setup.sh -i` should be run as root, and `./setup.sh -ymc` should be run as the local user.

### DIFF
--- a/bin/generate-zbm
+++ b/bin/generate-zbm
@@ -30,6 +30,8 @@ use boolean;
 
 use Storable qw( dclone );
 
+use constant REFARRAY => ref [];
+
 sub versionedKernel;
 sub latestKernel;
 sub createInitramfs;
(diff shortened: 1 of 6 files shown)


zdykstra:
If the `DracutFlags` configuration option is reasonable, it also needs to be documented globally. I'm holding off on writing that for now.

ahesford (approved):
This looks good after a read through. An alternative to adding a `DracutFlags` configuration option would be a `--local|-l` option to `generate-zbm`, but that would be less flexible. The question is whether additional flags might be useful beyond enabling local mode.

PR #84, opened 2020-10-03
Simplify handling of BE command lines, use in previews

ahesford:
`find_kernel_args` has been split into `preload_be_cmdline` and `load_be_cmdline`. The `preload_be_cmdline` function reads command-line arguments from a ZFS filesystem or the `/etc/default/{zfsbootmenu,grub}` files therein and stores the command line to `${BASE}/${BE}/cmdline` for a currently selected boot environment `${BE}`.

The `load_be_cmdline` only prints the contents of `${BASE}/cmdline` if it exists; otherwise, it will attempt to load the contents of `${BASE}/${BE}/cmdline` if that file exists (falling back to some generic default when the file does not exist) and attempt to remove any `resume` parameters and add a `noresume` parameter if a pool was mounted read-write in noresume mode.

Tested lightly in the new qemu testbed, with manual addition of a `noresume` trigger from an emergency shell.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -185,45 +185,8 @@ kexec_kernel() {
     return 1
   fi
 
-  cli_args="$( find_kernel_args "${fs}" "${mnt}" )"
+  cli_args="$( load_be_cmdline "${fs}" )"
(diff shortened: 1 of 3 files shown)


zdykstra (approved):
This looks great! It'll be nice seeing the KCL preview reflect reality, now.

PR #85, opened 2020-10-04
Allow ctrl-c to be used during user input

zdykstra:
Globally disable ctrl-c in /bin/zfsbootmenu by pointing it to an emptyhandler. Inside a subshell, unset the empty handler and then invoke zfsbootmenu-input with a single quoted argument; the pre-filled line that the user can edit. If ctrl-c is detected here, the script kills itself. Otherwise, it echos the line the user entered.

It is up to the caller to determine how to handle an empty response.


ahesford (approved):
This should definitely be used in the noresume prompt. Looks good.

ahesford:
I've changed the noresume prompt to use the new zfsbootmenu-input script, works as expected. I've also pushed some clean-up changes to master, but apparently this branch will rebase cleanly.

PR #87, opened 2020-10-23
Rewrite pool import process

zdykstra:
The historical pool import process relied on scraping the pool status and attempting to import any pool that is marked as ONLINE. This process fails for pools that feature flags enabled from a newer version of ZFS than what is in the initramfs. The process has been reworked as follows:

* Rely on `zpool` to import every pool possible in `read-only` mode
* If any pool has unavailable feature flags, set a global flag disabling read-write pool operations, and then provide a warning indicating that ZFS in the initramfs should be upgraded.
* If the preferred zpool could not be imported, warn and then drop to a recovery shell. 

This PR removes the `read_write` kernel command line option, as this behavior directly conflicts with requirement that all pools be imported read-only initially.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-exec.sh
+++ b/90zfsbootmenu/zfsbootmenu-exec.sh
@@ -2,7 +2,6 @@
 
 export spl_hostid
 export force_import
-export read_write
 export menu_timeout
 export root
 
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -624,7 +624,7 @@ import_pool() {
   fi
 
   # shellcheck disable=SC2086
-  status="$( zpool import "${import_args[@]}" ${pool} )"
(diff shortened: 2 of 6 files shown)


ahesford (approved):
All of my testing suggests this is ready to go, but there is one more test that needs to be done. The `set_rw_pool` function looks to `${BASE}/degraded` to decide if any particular pool should be prohibited from an read/write import.  It should be possible to reimport pools that are not degraded. I haven't yet reset my test setup to a state where at least one pool is not degraded to confirm that we can perform operations that require write privileges (e.g., setting a default BE) on fully online pools while continuing to honor the degraded state of others.

zdykstra:
I was able to boot a test VM with a 2.0.0-rc4 pool attached, and a 0.8.5 pool attached. Attempting to clone/promote a snapshot on the 2.0.0-rc4 pool failed with the correct warning. Cloning/promoting a snapshot on the 0.8.5 pool worked.

I'm not sure if there are any other test cases that should be explored, beyond what was discussed on IRC.

PR #91, opened 2020-10-28
Read-only helpers

zdykstra:
Add an alt-w helper to the main screen to set a pool r/w
Show r/o or r/w in the preview header for a selected pool (extracted from a BE)

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-preview.sh
+++ b/90zfsbootmenu/zfsbootmenu-preview.sh
@@ -25,10 +25,14 @@ do
   selected_arguments="${line}"
 done <<< "${BE_ARGS}"
 
+pool="${ENV%%/*}"
+readonly_prop="$( zpool get -H -o value readonly "${pool}" )"
+[[ ${readonly_prop} = "on" ]] && _readonly="r/o" || _readonly="r/w"
+
 if [[ "${BOOTFS}" =~ ${ENV} ]]; then
-  selected_env_str="${ENV} (default) - ${selected_kernel}"
+  selected_env_str="${ENV} (default, ${_readonly}) - ${selected_kernel}"
 else
-  selected_env_str="${ENV} - ${selected_kernel}"
+  selected_env_str="${ENV} (${_readonly}) - ${selected_kernel}"
 fi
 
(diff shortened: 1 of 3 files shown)


90zfsbootmenu/zfsbootmenu-preview.sh
```diff
@@ -25,10 +25,14 @@ do
   selected_arguments="${line}"
 done <<< "${BE_ARGS}"
 
+pool="${ENV%%/*}"
+readonly_prop="$( zpool get -H -o value readonly "${pool}" )"
+[[ ${readonly_prop} = "on" ]] && _readonly="r/o" || _readonly="r/w"
+
 if [[ "${BOOTFS}" =~ ${ENV} ]]; then
-  selected_env_str="${ENV} (default) - ${selected_kernel}"
+  selected_env_str="${ENV} (${_readonly}) (default) - ${selected_kernel}"
```
ahesford, line 33: I recommend `(default, ${_readonly})` instead of separate parentheses for each.

90zfsbootmenu/zfsbootmenu.sh
```diff
@@ -344,6 +344,11 @@ while true; do
       "alt-r")
         emergency_shell "alt-r invoked"
         ;;
+      "alt-w")
+        pool="${selected_be%%/*}"
+        set_rw_pool "${pool}"
+        CLEAR_SCREEN=1 key_wrapper "${pool}"
```
ahesford, line 350: Should this be
```
if set_rw_pool "${pool}"; then
  CLEAR_SCREEN=1 key_wrapper "${pool}"
fi
```
to mirror other uses in `zfsbootmenu-lib.sh`?

ahesford (changes requested):

90zfsbootmenu/zfsbootmenu-preview.sh
zdykstra: agreed, done!

90zfsbootmenu/zfsbootmenu.sh
zdykstra: blah, yeah. Fixed.

ahesford (approved):

PR #92, opened 2020-10-29
Embedded help system

zdykstra:
To enable a global help system, bind alt-h by default to all instances
of fzf and sk. The bind executes zfsbootmenu-help -L $_SECTION.

Before invoking each instance of fzf/sk, set _SECTION to the appropriate
value (MAIN, SNAPSHOT, ...). zfsbootmenu-help then uses this argument to
pre-sort the list of topics, pushing that topic to the head of the list
/ pre-selecting it.

The documentation is internal to zfsbootmenu-help, and is created using
a very simple templating system with basic colors and text centering.
fold is used to word wrap based on the size of the preview window. The
preview window size is calculated based on the number of columns minus
26 characters (the longest topic title + fzf marker ).

When adding new help sections, care should be taken with colorizing
text. Because ANSI escape sequences are used, fold is not able to
correctly calculate the length of given word, so it word wraps
incorrectly.

The text / pager can be reviewed by manually executing `90zfsbootmenu/zfsbootmenu-help.sh`.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-help.sh
+++ b/90zfsbootmenu/zfsbootmenu-help.sh
@@ -2,6 +2,7 @@
 # vim: softtabstop=2 shiftwidth=2 expandtab
 WIDTH="$( tput cols )"
 PREVIEW_SIZE=$(( WIDTH - 26 ))
+[ ${PREVIEW_SIZE} -lt 10 ] && PREVIEW_SIZE=10
 
 [ -z "${FUZZYSEL}" ] && FUZZYSEL="fzf"
 
@@ -70,13 +71,13 @@ Access a list of kernels available in the boot environment.
 $( colorize lightblue "[ALT+D] set bootfs" )
 Set the selected boot environment as the default for the pool.
 
-The operation will gracefully fail if the pool can not be set $( colorize red "read/write" ).
+The operation will fail gracefully if the pool can not be set $( colorize red "read/write" ).
 
 $( colorize lightblue "[ALT+S] snapshots" )
 Access a list of snapshots of the selected boot environment. New boot environments can be created here.
 
 $( colorize lightblue "[ALT+C] cmdline" )
-Temporarily edit the kernel command line that will be used to boot the next kernel and boot environment. This change is not persisted between boots.
+Temporarily edit the kernel command line that will be used to boot the chosen kernel in the selected boot environment. This change does not persist across reboots.
 
 $( colorize lightblue "[ALT+P] Pool status" )
 View the health and status of each imported pool.
@@ -93,7 +94,7 @@ This creates a boot environment that does not depend on any other snapshots, all
 
 A duplicated boot environment is commonly used if you need a new boot environment without any associated snapshots.
 
-The operation will gracefully fail if the pool can not be set $( colorize red "read/write" ).
+The operation will fail gracefully if the pool can not be set $( colorize red "read/write" ).
 
 If $( colorize red "mbuffer" ) is available, it is used to provide feedback.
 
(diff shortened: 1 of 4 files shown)


90zfsbootmenu/zfsbootmenu-help.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash
+# vim: softtabstop=2 shiftwidth=2 expandtab
+WIDTH="$( tput cols )"
+PREVIEW_SIZE=$(( WIDTH - 26 ))
```
ericonr, line 4: what happens if this becomes negative? Real question.

90zfsbootmenu/zfsbootmenu-help.sh
```diff
@@ -0,0 +1,195 @@
...
+$( colorize lightblue "[ENTER] boot" )
+Boot the selected boot environment, with the listed kernel and kernel command line visible at the top of the screen.
+
+$( colorize lightblue "[ALT+K] kernel" )
+Access a list of kernels available in the boot environment.
+
+$( colorize lightblue "[ALT+D] set bootfs" )
+Set the selected boot environment as the default for the pool.
+
+The operation will gracefully fail if the pool can not be set $( colorize red "read/write" ).
```
ericonr, line 73: "fail gracefully"

90zfsbootmenu/zfsbootmenu-help.sh
```diff
@@ -0,0 +1,195 @@
...
+
+$( colorize lightblue "[ALT+S] snapshots" )
+Access a list of snapshots of the selected boot environment. New boot environments can be created here.
+
+$( colorize lightblue "[ALT+C] cmdline" )
+Temporarily edit the kernel command line that will be used to boot the next kernel and boot environment. This change is not persisted between boots.
```
ericonr, line 79: "next kernel" is kinda weird... Perhaps "to boot the boot environment" or "to boot the kernel from the boot environment" ?

ericonr: "this change isn't persistent across boots" or shutdowns? Not sure.

90zfsbootmenu/zfsbootmenu-help.sh
```diff
@@ -0,0 +1,195 @@
...
+
+$( colorize lightblue "[ALT+P] Pool status" )
```
ericonr, line 81: What is the capitalization? Here it's `Pool status`, but above it's `set bootfs`.

90zfsbootmenu/zfsbootmenu-help.sh
```diff
@@ -0,0 +1,195 @@
...
+View the health and status of each imported pool.
```
ericonr, line 82: I think it can be "view health and status...", right?

90zfsbootmenu/zfsbootmenu-help.sh
```diff
@@ -0,0 +1,195 @@
...
+
+The operation will gracefully fail if the pool can not be set $( colorize red "read/write" ).
+EOF
+SECTIONS+=("SNAPSHOT Snapshot Management")
+
+# shellcheck disable=SC2034
+read -r -d '' KERNEL <<EOF
+$( colorize magenta "$( center "Kernel Management")" )
+$( colorize lightblue "[ENTER] boot" )
+Immediately boot the selected kernel in the boot environment, with the kernel command line shown at the top of the screen.
```
ericonr, line 125: "Boot into the boot environment with the kernel selected kernel immediately" ? Or close to this, I'm not 100%

90zfsbootmenu/zfsbootmenu-help.sh
```diff
@@ -0,0 +1,195 @@
...
+ $( colorize orange "F") Regular file
+
+EOF
+SECTIONS+=("DIFF Diff Viewer")
+
+# shellcheck disable=SC2034
+read -r -d '' POOL <<EOF
+$( colorize magenta "$( center "zpool Health")" )
+$( colorize lightblue "[ALT+R] Rewind checkpoint" )
+If a pool checkpoint is available, the selected pool is exported and then imported with $( colorize red "--rewind-to-checkpoint" ) set.
```
ericonr, line 164: "and then imported with the <> flag" ?

ericonr (commented):
Perhaps use "This operation" instead of "The operation"? I'm not sure on this one.

And switch all of them to "fail gracefully".

90zfsbootmenu/zfsbootmenu-help.sh
ahesford: This change does not persist across reboots.

90zfsbootmenu/zfsbootmenu-help.sh
ahesford: No, the definite article should be here

90zfsbootmenu/zfsbootmenu-help.sh
ahesford: Maybe "Immediately boot the chosen kernel in the selected boot environment"

90zfsbootmenu/zfsbootmenu-help.sh
ahesford: Temporarily edit the kernel command line that will be used to boot the chosen kernel in the selected boot environment.

ahesford (commented):
I pushed a commit to use `$HELP_SECTION` instead of `$_SECTION` to make the meaning more obvious, and believe I've made the global variable unnecessary. Please look over the commit. If you think these are positive changes, please squash; otherwise, just drop the commit.

Beyond this, the grammatical changes should be made but are not reflected in my commit.

90zfsbootmenu/zfsbootmenu-help.sh
zdykstra: Good catch!

90zfsbootmenu/zfsbootmenu-help.sh
zdykstra: I was mirroring the descriptions used in the footer. Those should probably get a normalization pass next.

PR #94, opened 2020-11-02
Console fixes

zdykstra:
zfsbootmenu-exec.sh was forcing stdin/stdout on `/dev/tty1`. This works correctly for systems with one GPU, but it causes problems for systems with multiple GPUs or when attempting to use a serial port for the console. To work around this problem, ZFSBootMenu now respects the `console=` parameter on the KCL, and plumbs the input/output of `zfsbootmenu.sh` to that device. If multiple `console=` entries are found, the last one found is used (consistent with kernel behavior). If no entries are found, `/dev/tty0` is used instead.

The testing infrastructure was reworked to default the output to a serial port (`ttyS0` or `hvc0`, depending on platform) which is then plumbed to the terminal used to invoke `run.sh`. `-v gtk` or `-v sdl` can instead be used to start a graphical display. When the serial output is active, it is automatically resized via stty, by reading `zbm.lines` and `zbm.columns` from the KCL.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-exec.sh
+++ b/90zfsbootmenu/zfsbootmenu-exec.sh
@@ -5,7 +5,8 @@ export force_import
 export menu_timeout
 export root
 export control_term
+export zbm_lines
+export zbm_columns
 
 # https://busybox.net/FAQ.html#job_control
 exec setsid bash -c "exec /bin/zfsbootmenu <${control_term} >${control_term} 2>&1"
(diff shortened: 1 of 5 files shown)


zdykstra:
Tested on my T450s test box, works like a charm on bare metal.

ahesford (approved):
Runs fine on my x86_64 laptop and honors the `zbm.lines` and `zbm.columns` KCL arguments on real hardware.

PR #95, opened 2020-11-03
Sort environments, fix preview

zdykstra:
Fix a string comparison in zfsbootmenu-preview.sh to make it an exact
match instead of a regexp match. This will prevent multiple environments
showing up as (default).

In populate_be_list, reverse sort the list of candidate boot
environments and then append the BOOTFS value if defined. Combined with
fzf/sk --tac, this pushes the boot environment to the top of the list so
that it's the first selected item on the main screen.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -957,7 +957,7 @@ key_wrapper() {
 # returns: 0 on success, 1 on failure
 
 populate_be_list() {
-  local be_list fs mnt active candidates
+  local be_list fs mnt active candidates sorted
 
   be_list="${1}"
   [ -n "${be_list}" ] || return 1
@@ -977,10 +977,19 @@ populate_be_list() {
       # All other datasets are ignored
       continue
     fi
+    if [ -n "${BOOTFS}" ] && [ "${BOOTFS}" = "${fs}" ] ; then
+      # If BOOTFS is defined, we'll manually append it to the array
+      continue
+    fi
+
     candidates+=( "${fs}" )
   done <<< "$(zfs list -H -o name,mountpoint,org.zfsbootmenu:active)"
 
-  for fs in "${candidates[@]}"; do
+  # Reverse sort the list, to line up with our global usage of --tac
+  readarray -t sorted < <( for env in "${candidates[@]}" ; do echo "$env" ; done | sort -r )
+  [ -n "${BOOTFS}" ] && sorted+=( "${BOOTFS}" )
+
+  for fs in "${sorted[@]}"; do
     # Unlock if necessary
     key_wrapper "${fs}" || continue
 
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -957,7 +957,7 @@ key_wrapper() {
 # returns: 0 on success, 1 on failure
 
 populate_be_list() {
-  local be_list fs mnt active candidates sorted
+  local be_list fs mnt active candidates
 
   be_list="${1}"
   [ -n "${be_list}" ] || return 1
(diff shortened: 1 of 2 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -977,10 +977,19 @@ populate_be_list() {
...
+      # If BOOTFS is defined, we'll manually append it to the array
+      continue
+    fi
+
     candidates+=( "${fs}" )
   done <<< "$(zfs list -H -o name,mountpoint,org.zfsbootmenu:active)"
 
-  for fs in "${candidates[@]}"; do
+  # Reverse sort the list, to line up with our global usage of --tac
+  readarray -t sorted < <( for env in "${candidates[@]}" ; do echo "$env" ; done | sort -r )
```
ahesford, line 989: I believe
```
printf "%s\n" "${candidates[@]}" | sort -r
```
works in place of the shell loop, even if individual entries of `candidates` have spaces.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -977,10 +977,19 @@ populate_be_list() {
       # All other datasets are ignored
       continue
     fi
+    if [ -n "${BOOTFS}" ] && [ "${BOOTFS}" = "${fs}" ] ; then
+      # If BOOTFS is defined, we'll manually append it to the array
```
ahesford, line 981: Might need to set `have_bootfs=1` here for some local variable `have_bootfs`, see comment on Line 990.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -977,10 +977,19 @@ populate_be_list() {
...
+      continue
+    fi
+
     candidates+=( "${fs}" )
   done <<< "$(zfs list -H -o name,mountpoint,org.zfsbootmenu:active)"
```
ahesford, line 986: What about
```
$(zfs list -H -o name,mountoint,org.zfsbootmenu:active | sort -r)
```
to ensure `candidates` is presorted? This would make the `readarray` in Line 989 and the entire `sorted` array unnecessary.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -977,10 +977,19 @@ populate_be_list() {
       # All other datasets are ignored
       continue
     fi
+    if [ -n "${BOOTFS}" ] && [ "${BOOTFS}" = "${fs}" ] ; then
```
ahesford, line 980: Do we need to test if `${BOOTFS}` is non-empty? The value of `${fs}` should never be empty so the test will still work as expected even when `${BOOTFS}` is empty.

Also, is there value in verifying that `${fs}` is nonempty at the start of the loop in Line 970?

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -977,10 +977,19 @@ populate_be_list() {
...
+      continue
+    fi
+
     candidates+=( "${fs}" )
   done <<< "$(zfs list -H -o name,mountpoint,org.zfsbootmenu:active)"
 
-  for fs in "${candidates[@]}"; do
+  # Reverse sort the list, to line up with our global usage of --tac
+  readarray -t sorted < <( for env in "${candidates[@]}" ; do echo "$env" ; done | sort -r )
+  [ -n "${BOOTFS}" ] && sorted+=( "${BOOTFS}" )
```
ahesford, line 990: If the `bootfs` property is set on a pool but the target filesystem doesn't exist, I believe we will still have a nonempty `${BOOTFS}`. This means we might potentially add a BE to the list that doesn't exist. If we add `have_bootfs` as suggested in Line 981, the test becomes `[ -n "${have_bootfs}" ]` to ensure we only add a BE that exists.

ahesford (changes requested):

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Agreed -  testing if BOOTFS is non-empty is largely pointless.  `${fs}` will always be defined when listing filesystems.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: I reverse sorted the list coming out of `zfs list ...`, works like a charm without a pointless duplicated array.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: In my testing, `bootfs` always has to exist, if it is defined. `zpool` won't let me assign a value to `bootfs` if the filesystem doesn't exist. If the `bootfs` filesystem is deleted, the pool property is automatically cleared.

If `bootfs` points to a filesystem that is not a valid BE, the tests that process the candidates array will catch it and discard it.

zdykstra:
Validating `bootfs` before the countdown timer could be useful; except it'll produce weird behavior on systems with native encryption turned on. There will be a password prompt and then the countdown timer will start - the reverse of the current process.

90zfsbootmenu/zfsbootmenu-lib.sh
ahesford: You're right, there is no value in omitting a non-empty `$BOOTFS` since it will be validated by `find_be_kernels` later in this function.

ahesford:
I didn't think about the effects of encryption on pre-validation of `$BOOTFS` to short-circuit the menu. Asking for a passphrase before presenting the countdown timer seems backwards. The current behavior seems best for now.

ahesford (approved):

PR #98, opened 2020-11-16
Encapsulate test environments within subdirectories

ahesford:
It is sometimes helpful to have multiple active test setups, and it is also convenient to be able to destroy a test setup with one command rather than having to delete multiple different files and directories.

This command adds a `-D` argument to setup.sh, allowing all activity in the setup script to occur within a specified directory. When a subdirectory is not specified, a default of `test.$(uname -r)` is used.

The run.sh script is also now aware of the test directory, searching for a default of `test.$(uname -r)` and allowing manual specification by setting the TESTDIR environment variable. If the TESTDIR does not exist, run.sh reverts to running from the CWD.

Also, the image-creation script has been pulled out of a heredoc in setup.sh into a standalone script, making maintenance a little easier.

### DIFF
--- a/testing/.gitignore
+++ b/testing/.gitignore
@@ -7,3 +7,4 @@ local.yaml
 .config
 dracut
 dracut.conf.d
+generate-zbm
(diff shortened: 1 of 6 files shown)


zdykstra (approved):
Everything here looks great! I can only test so far, because I've flipped most of my machines to ZFS 2.0.0-rc6. I was able to build a few different VMs, some glibc some musl and at least boot them into ZBM / part of the way into the boot environment.

PR #99, opened 2020-11-17
Fall back to legacy hostid creation for ZFS < 2.0

ahesford:
The dracut module in ZFS < 2.0 produces initramfs images with all-zero hostid files on musl systems, regardless of the contents of /etc/hostid on the host. In ZFS >= 2.0, the dracut module respects the contents of /etc/hostid when possible, and will otherwise create a new /etc/hostid only if the hostid(1) program returns a non-zero value. (NB: to the ZFS kmods, a non-existent /etc/hostid is equivalent to a file containing all zeros.)

To minimize the possibility of import failures, ZFSBootMenu should try hard to replicate the behavior of whichever version of ZFS is creating initramfs images.

The version comparison here should be considered deprecated and removed once ZFS >= 2.0 has had sufficient time to overtake prior versions in popularity.

### DIFF
--- a/90zfsbootmenu/module-setup.sh
+++ b/90zfsbootmenu/module-setup.sh
@@ -152,17 +152,36 @@ install() {
     type mark_hostonly >/dev/null 2>&1 && mark_hostonly /etc/zfs/vdev_id.conf
   fi
 
-  # Synchronize initramfs and system hostid
-  if [ -e /etc/hostid ]; then
-    # Prefer the hostid file if it exists
+  # Try to synchronize hostid between host and ZFSBootMenu
+  #
+  # DEPRECATION NOTICE: on musl systems, zfs < 2.0 produced a bad hostid in
+  # dracut images. Unfortunately, this should be replicated for now to ensure
+  # those images are bootable. After some time, remove this version check.
+  ZVER="$( zfs version | head -n1 | sed 's/zfs-\(kmod-\)\?//' )"
+  if [ -n "${ZVER}" ] && printf '%s\n' "${ZVER}" "2.0" | sort -VCr; then
+    NEWZFS=yes
+  else
+    NEWZFS=""
+  fi
+
+  if [ -n "${NEWZFS}" ] && [ -e /etc/hostid ]; then
+    # With zfs >= 2.0, prefer the hostid file if it exists
     inst /etc/hostid
-    type mark_hostonly >/dev/null 2>&1 && mark_hostonly /etc/hostid
-  elif HOSTID="$( hostid 2>/dev/null )" && [ "${HOSTID}" != "00000000" ]; then
-    # Fall back to `hostid` output when it is nonzero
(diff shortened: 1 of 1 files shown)


zdykstra:
`elif HOSTID="$( hostid 2>/dev/null )" && [ "${HOSTID}" != "00000000" ]; then` is giving me pause. The historical behavior was to unconditionally capture the output of `hostid` and echo it into `${initdir}/etc/hostid`. Won't this result in MUSL system without /etc/hostid having nothing at all in `${initdir}/etc/hostid` ?

ahesford:
Yes, but that doesn't matter. With the old behavior, the module setup (both ours and `90zfs`) would always dump all zeros to `/etc/hostid`. The `spl` module attempts to first look at `sysfs` to recover the `spl_hostid` command-line parameter and, if that fails, attempts to read the hostid from `/etc/hostid`. If that file cannot be read for any reason, the function returns `0` (see https://github.com/openzfs/zfs/blob/zfs-0.8-release/lib/libspl/gethostid.c for example), which is exactly the same value the function would read from an `/etc/hostid` filled with zeros as produced on a musl system.

If this still concerns you, we can always do `{ [ -z "${NEWZFS}" ] || [ "${HOSTID}" != "00000000" ] }`, which we'll just have to remember to remove when we eventually drop version comparisons.

zdykstra:
Thanks for the explanation - that behavior makes sense. As long as we're restoring the `hostid` behavior for ZFS 0.8.5/musl with this, I'm happy with it. I'll test tonight by replicating what Project Trident has setup ( /etc/hostid, spl_hostid not explicitly set on any KCL).

ahesford:
I've revised my thinking on this. For glibc systems, it is possible under some circumstances to remove `/etc/hostid` and get an all-zero output `hostid(1)` output. (This happens, for example, in the glibc test environment.) The old behavior would force an all-zero `/etc/hostid` in the output, whereas my proposed change would have omitted the file in that case. However, it is also possible under some conditions that, when booting to the glibc initramfs with no `/etc/hostid`, the `hostid(1)` executable will return some other, randomly generated ID.

This should *not* be an issue for ZFS for the reasons I described above. However, it means that, under these conditions, the "official" `90zfs` module would produce an initramfs with an all-zero `/etc/hostid` while `90zfsbootmenu` will produce an initramfs with no file. This seems to violate the principle of least astonishment.

Better to reproduce the behavior of `90zfs` with the best possibile fidelity.

In that spirit, since we now care about the ZFS version, we can rely on `zgenhostid` in OpenZFS 2, doing exactly what `90zfs` does in that case. For prior versions, we fall back to `echo`, which is what `90zfs` did there anyway.

zdykstra:
:+1:  shipit!

ahesford:
I did some tests in a musl VM and on a glibc host and believe the revised version does what's expected in all cases, so might as well merge it. We can take another look tomorrow, convince ourselves that this is correct and cut a release to fix behavior for musl users.

PR #102, opened 2020-12-02
Improvements to createInitramfs and handling out-of-space issues

ahesford:
There are a few changes in here that need review. The first is not directly connected to most of the changes: merging the `unifiedEFI` and `createInitramfs` subroutines to ensure consistent behavior in both modes (in particular, honoring dracut flags specified in the ZBM config for unified EFI images, where this feature was not present before).

Other changes are related to properly handling situations where the ESP free space is exhausted while copying a new initramfs, kernel or unified EFI image. The unified image will probably be copied in place, but may not be bootable in this situation; if individual components are copied, a partial initramfs may be left behind and appear bootable, but it may be missing key contents.

For unversioned backups, this is avoided by copying files (either the unified EFI image or the kernel and initramfs) into the target directory of the ESP *under a temporary name*, so we have a nondestructive way to learn that the copy was successful before rolling the backups and replacing the target file. The replacement is done with `move`, which I don't expect to fail because it should be doing `rename` under the hood.

For versioned backups, the temporary dance is not required, because the destination file(s) have unique names by design. The change here is confirming that the files can be copied successfully before rolling backups, and aborting if they can't.

In both modes, any partial files (either temporary or versioned) are cleaned up on failure. The `verboseUnlink` subroutine now supports custom status messages so we can print some informative text while deleting these partial files.

**Question**: I would like to use `rename` rather than `move` to guarantee we are using the atomic operation instead of falling back to a copy in some edge case (which could still fail due to space exhaustion, rendering all of this work pointless). I don't think `move` should *ever* resort to copying as used here, since the source and destination are on the same filesystem (same directory, no less). Is there *any* reason not to explicitly use `rename` here?

Tests: I've done some tests on a space-constrained target directory and believe everything works, but the following tests should be independently verified:
- [x] Component mode, versioned backups
- [x] Component mode, unversioned backups
- [x] Unified EFI mode, versioned backups
- [x] Unified EFI mode, unversioned backups

**NB**: It is conceivable that there are still some unlikely circumstances that could result in destruction of viable backups with this implementation. In unversioned backup mode, if the copy to a temporary file succeeds and we roll the backups, but the move to final somehow fails, we've inappropriately discarded the backup. I'm not sure under what circumstances a move might fail, but I think the likelihood is small enough that we shouldn't worry about trying to avoid this scenario.

### DIFF
--- a/bin/generate-zbm
+++ b/bin/generate-zbm
@@ -206,7 +206,6 @@ my $tempdir = $dir->dirname;
 if ( nonempty $config{Kernel}{Path} and !nonempty $runConf{kernel} ) {
   $runConf{kernel} = $config{Kernel}{Path};
 }
-
 if ( nonempty $config{Kernel}{Prefix} and !nonempty $runConf{kernel_prefix} ) {
   $runConf{kernel_prefix} = $config{Kernel}{Prefix};
 }
@@ -537,6 +536,7 @@ sub latestKernel {
     my @kernels = glob($glob);
     next if !@kernels;
     for ( sort { versioncmp( $b, $a ) } @kernels ) {
+      Log("Latest kernel: $_");
       return $_;
     }
   }
@@ -580,6 +580,9 @@ sub createInitramfs {
   }
 
   push( @cmd, ( $output_file, $kver ) );
+
+  my $command = join(' ', @cmd);
+  Log("Executing: $command");
 
   my @output = execute(@cmd);
   my $status = pop(@output);
@@ -656,6 +659,7 @@ sub maxRevision {
     }
   }
 
+  Log("maxRevision discovered: $revision");
   return $revision;
 }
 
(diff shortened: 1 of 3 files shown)


zdykstra (approved):
I've run this branch through a gamut of tests and it's passed each test with flying colors. The core logic here for versioned/unversioned component and EFI image generation is sound.

The Perl documentation for `move` (via File::Copy) says:

`If possible, move() will simply rename the file. Otherwise, it copies the file to the new location and deletes the original. If an error occurs during this copy-and-delete process, you may be left with a (possibly partial) copy of the file under the destination name.`

It's not clear what conditions have to be met for it to rename the file.

`rename` says that it either works or doesn't, and won't work across filesystem boundaries - but that's not an issue here. I support switching the move to a rename.

ahesford:
Did some rearrangment of the commits and merged.

PR #103, opened 2020-12-07
Enable TMUX to support debugging efforts

zdykstra:
It'd be nice to have the ability to see what is on the filesystem, what processes are running, etc in a running initramfs no matter where we are in ZFSBootMenu. Enter `tmux`. Set `zbm.tmux=1` on the ZBM KCL and build the initramfs with [dracut-tmux](https://github.com/zbm-dev/dracut-tmux) and /bin/zfsbootmenu will be launched in `tmux`. This has the added benefit of enabling job control for all processes launched under tmux.

I've moved anything considered 'setup' from `zfsbootmenu-countdown.sh` to `zfsbootmenu-exec.sh` so that whatever is launched (tmux, or zfsbootmenu-countdown) is started with the correct serial terminal size, environment variables, etc.

This requires someone to doubly opt in - both by explicitly enabling the `dracut-tmux` module and by setting the right KCL argument. This is otherwise a very very small code path change in `zfsbootmenu-exec.sh`.

### DIFF
--- a/90zfsbootmenu/module-setup.sh
+++ b/90zfsbootmenu/module-setup.sh
@@ -81,6 +81,7 @@ install() {
     "blkid"
     "awk"
     "fold"
+    "ps"
   )
 
   for _exec in "${essential_execs[@]}"; do
--- a/90zfsbootmenu/zfsbootmenu-countdown.sh
+++ b/90zfsbootmenu/zfsbootmenu-countdown.sh
@@ -1,14 +1,6 @@
 #!/bin/bash
 # vim: softtabstop=2 shiftwidth=2 expandtab
 
-# store current kernel log level
-read -r PRINTK < /proc/sys/kernel/printk
-PRINTK=${PRINTK:0:1}
-export PRINTK
-
-# Set it to 0
-echo 0 > /proc/sys/kernel/printk
-
 # disable ctrl-c (SIGINT)
 trap '' SIGINT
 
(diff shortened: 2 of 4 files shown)


ahesford (approved):
The rearrangement looks solid to me. It might be nice to move all of the import and investigation into zfsbootmenu-exec and leave zfsbootmenu-countdown just for the actual countdown and menu/shell loop. However, I don't think this is feasible bcause the imports and beyond require interaction, possibly including emergency shells, that should have proper job control and ought to be included in tmux when conifigured.

zdykstra:
On bare metal, this isn't doing the right thing when `zfsbootmenu-countdown` draws a count down timer. It thinks the window is 80x25 - it's only when `fzf` is started that the true display size is known. `tput cols` and `tput rows` also get the wrong size. I'm hacking at this for a bit.

zdykstra:
`stty size`, `TIOCGWINSZ ioctl [ioctl_tty(2)]`

ahesford:
At least from a qemu serial console, `stty size` reports `24 80` in a 100x90 terminal window, so there is no automatic conveyance of console size to whatever mechanism `stty` is using involved in the serial link handshake.

zdykstra:
The short-term solution to the problem was to execute `echo "Loading ZFSBootMenu ..."` in `zfsbootmenu-exec.sh`. This triggers *something* that allows exec'd processes (tmux, bash) to see the console at the correct size.

PR #104, opened 2020-12-07
Support optional "teardown" script to run before kexec

ahesford:
A new `zfsbootmenu_teardown` dracut option has been added to `module-init.sh` that, when set to a path that refers to an executable, will install the named executable as `/libexec/zfsbootmenu-teardown` in the initramfs. In `kexec_kernel`, if `/libexec/zfsbootmenu-teardown` is found to exist and be executable, the script will be run immediately before the final `kexec` call to jump into the new kernel.

On my 2015 "VoidBook Air", the system keyboard, trackpad and bluetooth controller are connected to the XHCI USB hub. These devices are fully functional in ZBM, but when I boot the final kernel,  I can see the hub but not the peripherals. With the following teardown hook:
```bash
#!/bin/sh

SYS_XHCI=/sys/bus/pci/drivers/xhci_hcd

for DEVPATH in ${SYS_XHCI}/????:??:??.?; do
	[ -L "${DEVPATH}" ] || continue
	DEVICE="${DEVPATH#${SYS_XHCI}/}"
	echo "Tearing down USB controller ${DEVICE}..."
	echo "${DEVICE}" > ${SYS_XHCI}/unbind
done
```
I can unbind the XHCI device from the driver, which allows the final kernel to properly initialize the device and my peripherals to function.

I suspect there are a few use cases like this where a teardown script will be useful to ease the kexec transition. The script is not intended to interact with any ZFS pools or filesystems, so is executed at the last possible moment (specifically, after the pool is exported.)

### DIFF
--- a/90zfsbootmenu/module-setup.sh
+++ b/90zfsbootmenu/module-setup.sh
@@ -138,6 +138,16 @@ install() {
   inst_hook cmdline 95 "${moddir}/zfsbootmenu-parse-commandline.sh" || _ret=$?
   inst_hook pre-mount 90 "${moddir}/zfsbootmenu-exec.sh" || _ret=$?
 
+  # Install a "teardown" hook if specified and it exists
+  # shellcheck disable=SC2154
+  if [ -n "${zfsbootmenu_teardown}" ]; then
+    if [ -x "${zfsbootmenu_teardown}" ]; then
+      inst_simple "${zfsbootmenu_teardown}" "/libexec/zfsbootmenu-teardown" || _ret=$?
+    else
+      dwarning "no executable teardown script (${zfsbootmenu_teardown}); cannot install"
+    fi
+  fi
+
   if [ ${_ret} -ne 0 ]; then
     dfatal "Unable to install core ZFSBootMenu functions"
     exit 1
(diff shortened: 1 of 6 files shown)


ahesford:
The XHCI teardown script has been added to `contrib/`.

zdykstra (changes requested):
Very clean/user-expandable solution here. It might be nice to document this mechanism in `zfsbootmenu.7` under a new section, and to update the Makefile to copy these to /usr/share/examples/zfsbootmenu/contrib.

ahesford:
On 2020-12-07 at 20:15 (UTC -0500), Zach Dykstra wrote:
> @zdykstra requested changes on this pull request.
> 
> Very clean/user-expandable solution here. It might be nice to document
> this mechanism in `zfsbootmenu.7` under a new section, and to update
> the Makefile to copy these to /usr/share/examples/zfsbootmenu/contrib.

I incorporated your requests. Not seeing a need for another level in the
hierarchy, I left out the `contrib` subdirectory in the `examples` tree.
If you object, I will put it in the subdirectory.

zdykstra (approved):

PR #112, opened 2020-12-10
Add option to auto-add all -pool.img disks in a test dir

zdykstra:
I'm tired of manually adding multiple disks in test directories. As long as disks are suffixed with `-pool.img`, this will glob 'em all.

### DIFF
--- a/testing/run.sh
+++ b/testing/run.sh
@@ -61,7 +61,7 @@ case "$(uname -m)" in
   ;;
 esac
 
-DRIVE=("-drive" "format=raw,file=${TESTDIR}/zfsbootmenu-pool.img")
+DRIVE=()
 INITRD="${TESTDIR}/initramfs-bootmenu.img"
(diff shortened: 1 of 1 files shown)


ahesford (approved):
Either do it this way, or make it the default behavior, which sounds good to me:
1. Drop the `-g` argument
2. Don't pre-seed `DRIVE` with `zfsbootmenu-pool.img` before argument processing
3. After argument processing:
    a. If `DRIVE` is empty, do the glob loop
    b. If `DRIVE` is not empty, do nothing use as is

The big change here is that using even one `-d` argument means you'd have to manually specify `zfsbootmenu-pool.img`. but that sems OK to me because we'll generally rely on globbing and it also allows switching between two alternative pool images, if there is a use case for that.

Either approach is fine with me.

ahesford (approved):
Seems to work as intended

PR #119, opened 2020-12-17
Support chroot'ing into a snapshot

zdykstra:


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -62,25 +62,35 @@ mount_zfs() {
   mnt="${BASE}/${fs}/mnt"
   test -d "${mnt}" || mkdir -p "${mnt}"
 
+  # @ always denotes a snapshot
+  if [[ "${fs}" =~ @ ]]; then
+    is_snapshot=1
```
ahesford, line 67: Add this to the `local` list

ahesford (approved):
Nice work, this could be a useful tool for diagnostics.

zdykstra:
Moved the `/tmp` bind mount to `/tmp` inthe chroot, instead of `/var/tmp`. There's no functional reason to have `/var/tmp` read-write. The text displayed when entering the chroot now mentions that `/tmp` is shared and read/write.

90zfsbootmenu/zfs-chroot.sh
```diff
@@ -17,19 +17,17 @@ if mountpoint="$( allow_rw=yes mount_zfs "${selected}" )"; then
   mount -t proc proc "${mountpoint}/proc"
   mount -t sysfs sys "${mountpoint}/sys"
   mount -B /dev "${mountpoint}/dev"
-  mount -B /tmp "${mountpoint}/var/tmp"
+  mount -B /tmp "${mountpoint}/tmp"
```
ahesford, line 20: I never noticed that this was already bound on `/var/tmp`, but I think `/tmp` is more intuitive.

PR #120, opened 2020-12-18
Namespace our KCL args, organize parsing

zdykstra:
Move all ZFSBootMenu-specific arguments to the `zbm.*` argument space. Legacy values are still accepted in lieu of the new `zbm.` value, but will produce a warning on the kernel ring buffer. Timeout values have been expanded to work as follows:

zbm.timeout (or timeout) to set an explicit timeout option
OR zbm.show sets `menu_timeout=-1` to force the main menu
OR zbm.skip sets `menu_timeout=0` to skip the menu and directly boot `BOOTFS`
finally if none of these options are set, `menu_timeout=10` is set. 

The order of these tests might want to be flipped around to give precedence to `zbm.show` / `zbm.skip` - thoughts?

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-parse-commandline.sh
+++ b/90zfsbootmenu/zfsbootmenu-parse-commandline.sh
@@ -16,21 +16,7 @@ else
   warn "ZFSBootMenu: Pools may not import correctly."
 fi
 
-# Force import pools only when explicitly told to do so
-if getargbool 0 force_import ; then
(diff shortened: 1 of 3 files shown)


ahesford (approved):
I don't think the "show" or "skip" options should take priority over "timeout", because the latter is more specific. Looks good as is.

zdykstra:
I'll have to have you re-review this after I write documentation for these new options / new option names.

PR #121, opened 2020-12-18
Test if :2222 is in use, increment until a free port is found

zdykstra:
Aside from the hard dependency on `lsof`, this does what it should so that multiple testing instances can be run at once (though not with the same backing disks).

### DIFF
--- a/testing/run.sh
+++ b/testing/run.sh
@@ -158,6 +158,19 @@ elif [ ! -f "${INITRD}" ] ; then
   exit 1
 fi
 
+SSH_PORT=2222
+while true; do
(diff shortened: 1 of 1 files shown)


ahesford:
Nice improvement; this has bitten me before. If you're going to print the port in use, maybe sleep (for five seconds?) if it won't use 2222, since that message will tend to get pushed off the screen in serial-console mode.

zdykstra:
> Nice improvement; this has bitten me before. If you're going to print the port in use, maybe sleep (for five seconds?) if it won't use 2222, since that message will tend to get pushed off the screen in serial-console mode.

Good call. I moved the message to only be shown if a non-standard port is used, and then set it to sleep for a few seconds.

PR #132, opened 2021-01-15
Improve kernel version detection.

ahesford:
The dracut lsinitrd utility determines kernel versions by looking for strings of the form `[0-9]+\.[0-9]+\.[0-9]+`, so we might as well adopt similar logic to try to recover versions when the file name is not sufficient. This improvement attempts to parse both the file name and kernel strings for version-like sequences. If it finds both, they must agree or the kernel is ignored; if it finds only one source of versioning, that is preferred automatically.

With more reliable version detection, we can again relax the kernel detection to look for `/boot/<prefix>` rather than requiring the more strict `/boot/<prefix>-<version>`, and an output kernel prefix becomes just the name of the kernel with the determined version removed (if it exists).

This *should* allow Arch users to rely on automatic version detection rather than requiring explicit specification of the kernel path and version. This also allows Arch users to run `generate-zbm` after updating a kernel package but before rebooting, when using a `%{current}` version string will fail because `/lib/modules/$(uname -r)` has been removed and the kernel file is of a different version anyway.

To test, I populated a false boot directory in a test tree:
```
vmlinuz -> /boot/vmlinuz-5.10.7_1
vmlinuz-5.10.3_1 -> /boot/vmlinuz-5.10.3_1
vmlinuz-5.10.7-generic-53 -> /boot/vmlinuz-5.10.6_1
vmlinuz-kernel-5.10.7_1 -> /boot/vmlinuz-5.10.7_1
vmlinuz-linux -> /boot/vmlinuz-5.10.6_1
```
Note some Arch-style names as well as an Ubuntu-style name (with a file-name version that disagrees with the actual version). Also, the `vmlinuz-kernel-5.10.7_1` name would have thwarted our original scheme, which would pick `vmlinuz` as the prefix and `kernel-5.10.7_1` as the version.

Running
```
PATH=./dracut:${PATH} ./generate-zbm -c ./local.yaml -b ./boot -d
```
in the test tree returns the expected output about kernel versioning:
```
## No version found in path ./boot/vmlinuz, using 5.10.7_1 from kernel strings
## Identified version 5.10.7_1 for kernel ./boot/vmlinuz
## Found version 5.10.3_1 in both path ./boot/vmlinuz-5.10.3_1 and kernel strings
## Identified version 5.10.3_1 for kernel ./boot/vmlinuz-5.10.3_1
## Ignoring inconsistent versions 5.10.7-generic-53 from path ./boot/vmlinuz-5.10.7-generic-53 and 5.10.6_1 in kernel strings
## Found version 5.10.7_1 in both path ./boot/vmlinuz-kernel-5.10.7_1 and kernel strings
## Identified version 5.10.7_1 for kernel ./boot/vmlinuz-kernel-5.10.7_1
## No version found in path ./boot/vmlinuz-linux, using 5.10.6_1 from kernel strings
## Identified version 5.10.6_1 for kernel ./boot/vmlinuz-linux
## Latest kernel: 5.10.7_1
## Found version 5.10.7_1 in both path ./boot/vmlinuz-kernel-5.10.7_1 and kernel strings
Creating ZFSBootMenu 1.8.1 from kernel ./boot/vmlinuz-kernel-5.10.7_1
```

### DIFF
--- a/bin/generate-zbm
+++ b/bin/generate-zbm
@@ -176,13 +176,11 @@ $runConf{umount_on_exit} = 0;
 if ( nonempty $config{Global}{BootMountPoint} ) {
   my $mounted = 0;
 
-  my $cmd    = "mountpoint $config{Global}{BootMountPoint}";
-  my @output = execute($cmd);
+  my @output = execute(qq(mountpoint $config{Global}{BootMountPoint}));
   my $status = pop(@output);
   unless ( $status eq 0 ) {
     print "Mounting $config{Global}{BootMountPoint}\n";
-    $cmd = "mount $config{Global}{BootMountPoint}";
-    my @output = execute($cmd);
+    my @output = execute(qq(mount $config{Global}{BootMountPoint}));
     my $status = pop(@output);
     if ( $status eq 0 ) {
       $runConf{umount_on_exit} = 1;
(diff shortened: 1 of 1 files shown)


bin/generate-zbm
```diff
@@ -523,22 +526,91 @@ sub versionedKernel {
...
+    Log("Unable to read path $kernel, assuming no version");
+    return;
+  }
+
+  # Read strings in the kernel to recover a version, if possible
+  my @output = execute(qq(strings $kernel));
+  my $status = pop(@output);
+  if ( $status eq 0 ) {
+    for (@output) {
+      if (/[0-9]+\.[0-9]+\.[0-9]+/) {
```
zdykstra, line 578: Should this be anchored to the start of a line, to avoid possibly matching against a random string in the middle of gibberish?

zdykstra (approved):
I don't have access to an Ubuntu or Debian host with vendor kernels - do you have access to a semi-modern Debian/Ubuntu machine to check that `strings $kernel` will find a value that aligns with what was detected in the filename? Aside from that, this looks like it produces the correct behavior on Void.

Edit: A friend with an Ubuntu 20.04 / Kernel 5.4.x machine was able to verify that the vmlinuz-$version pattern as defined in this PR will work there.

zdykstra:
Ubuntu:
```

# strings vmlinuz-5.4.0-61-generic | grep 5.4.0
5.4.0-61-generic (buildd@lcy01-amd64-018) #69-Ubuntu SMP Thu Jan 7 17:31:02 UTC 2021
5.4.0-61-genericq
5.4.0/arch/x86/purgatory
        5.4.0-61-generic SMP mod"
A5E45074588
P51490Y
A59450
5.4.0-61-
Ubuntu 5.4.0-61.69
s/5.4.0-61-9
5.4.0-61-genericA
5.4.0-61H
5.4.0-61-generic (buildd@lcy01-amd64-018) (gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04)) #69-Ubuntu SMP Thu Jan 7 17:31:02 UTC 2021
```

Issue #135, opened 2021-01-16
Detect EFI stub binary from systemd-boot


vozhyk-:
`generate-zbm` tries to find the EFI stub binary at `/usr/lib/gummiboot/linuxx64.efi.stub` by default and fails on my system, as I have `systemd-boot`, not `gummiboot`, installed.

It works if I specify the `systemd-boot` location of the binary in the config:
```
sed -i -e '/EFI:/a\  Stub: /usr/lib/systemd/boot/efi/linuxx64.efi.stub' /etc/zfsbootmenu/config.yaml
```

It would be nice to not have to configure this.

I see that `dracut` is supposed to guess the location by itself if it is not specified, but doesn't (https://bugs.gentoo.org/765208; same on my system with `dracut` version `050`).

ahesford:
It looks like dracut does as advertised; it looks first to the systemd location and falls back to the gummiboot location when `--uefi-stub` is not supplied on the command line. It seems reasonable for `generate-zbm` to omit the `--uefi-stub` command line when a specific value is not specified in the configuration and let dracut do its thing.

vozhyk-:
I see, `dracut` works as advertised on your system. Then that bug might be specific to Gentoo.

ahesford:
Sorry---I haven't verified it actually runs properly without `--uefi-stub`, I was just looking through the `dracut` script and everything seems reasonable. More testing is definitely warranted to determine whether the script works as it seems.

ahesford:
Not sure if it's relevant in your case, but if dracut can't use `pkg-config` to find the systemd path, it uses the parent of `[/usr]/lib/systemd/systemd-udevd` [as the $systemdutil path](https://github.com/dracutdevs/dracut/blob/master/dracut.sh#L1100-L1106). If this path doesn't exist, `$systemdutildir` will not be defined, and [dracut will look for the EFI stub](https://github.com/dracutdevs/dracut/blob/master/dracut.sh#L1181-L1188) at either `/boot/efi/linuxx64.efi.stub` or `/usr/lib/gummiboot/linuxx64.efi.stub`.

vozhyk-:
Thanks for looking into this in depth. I see now that on my system, neither of those exists:
```
# pkg-config systemd --variable=systemdutildir
Package systemd was not found in the pkg-config search path.
Perhaps you should add the directory containing `systemd.pc'
to the PKG_CONFIG_PATH environment variable
No package 'systemd' found
[...]

PR #138, opened 2021-01-17
Only specify `--uefi-stub` when EFI.Stub is configured

ahesford:
When `--uefi` is passed to dracut without a `--uefi-stub` argument, it looks first for a stub provided by systemd-boot and, if one isn't found, for a stub provided by gummiboot. The gummiboot location agrees with our default location at /usr/lib/gummiboot/linuxx64.efi.stub.

By dropping the `--uefi-stub` unless the `generate-zbm` configuration explicitly specifies a value, `generate-zbm` should work out of the box on systems that use gummiboot as well as its successor, systemd-boot. If a user specifies a value in EFI.Stub, the `--uefi-stub` argument will be passed with the configured value.

Fixes: #135.

Tested and produces the expected results on Void:
- With the default stub at `/usr/lib/gummiboot/linuxx64.efi.stub`
- With no gummiboot stub but the following two systemd components:
```
/usr/lib/systemd/systemd-udevd
/usr/lib/systemd/boot/efi/linuxx64.efi.stub
```
(Note: `dracut` requires that `systemd-udevd` exist before it looks in `boot/efi` for the stub.)
- With no default locations (fails, as expected)
- With no default location but `EFI.Stub` set to `/etc/zfsbootmenu/linuxx64.efi.stub` (which exists)

### DIFF
--- a/bin/generate-zbm
+++ b/bin/generate-zbm
@@ -648,13 +648,11 @@ sub createInitramfs {
 
   # If $kernfile is provided, make a unified EFI image with the named kernel
   if ( defined $kernfile ) {
-    my $efi_stub = $config{EFI}{Stub} || "/usr/lib/gummiboot/linuxx64.efi.stub";
-
-    unless ( -e $efi_stub ) {
-      die "Missing EFI stub: $efi_stub";
+    if ( nonempty $config{EFI}{Stub} ) {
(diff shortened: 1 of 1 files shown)


zdykstra (approved):
Your test cases answered all of the questions I had. Perfect-o.

Issue #136, opened 2021-01-17
Pass root=ZFS= to the BE kernel on Gentoo


vozhyk-:
I'm using `genkernel` to create the initramfs in the BE (which gets kexec'd by `zfsbootmenu`). That initramfs only accepts a ZFS root option in the form of `root=ZFS=<dataset>` or `root=ZFS`. If an option of the form `root=zfs:<dataset>` is supplied, it fails to boot.

I have set `org.zfsbootmenu:rootprefix="root=ZFS="` on `<pool>/ROOT`, and this fixed the problem.

Not having to configure this would be nice.

vozhyk-:
One way would be to make the `root=ZFS=<dataset>` format the default, at least in Gentoo.

I initially thought this wouldn't make sense on Gentoo, since people use different initramfs generators, but it seems only one I know of uses a different format, which is neither of the two.

- GRUB2 passes `root=ZFS=<dataset>`: https://git.savannah.gnu.org/gitweb/?p=grub.git;a=blob;f=util/grub.d/10_linux.in;h=e8b01c0d0c71cf1f85d95d73d204e09dc8ad7eba;hb=refs/heads/master#l76
- `dracut` (also popular in Gentoo) appears to accept `root=ZFS=<dataset>`: https://github.com/openzfs/zfs/blob/60a2434b295f395bbedcec2c4cd6a54530edf29a/contrib/dracut/90zfs/parse-zfs.sh.in#L31

The one using a different format is `bliss-initramfs` (which I used to use a long time ago for ZFS on LUKS) - and it seems to take `root=<dataset>` now: https://github.com/fearedbliss/bliss-initramfs/blob/master/USAGE.md#general

By the way, `initramfs-tools` (which is used in Ubuntu, IIRC), also accepts `root=ZFS=<dataset>` along with `root=zfs:<dataset>`: https://github.com/openzfs/zfs/blob/60a2434b295f395bbedcec2c4cd6a54530edf29a/contrib/initramfs/scripts/zfs#L755

vozhyk-:
Another way would be to read the initramfs image and try to guess what generated it - `lsinitrd` from `dracut` shows the contents of files in `/lib/dracut` in the image:
```
# lsinitrd /boot/initramfs-5.4.51-gentoo-x86_64.img
initrd in UEFI: : 14M
========================================================================
Version: Genkernel 3.5.2.5-848-g78db2b9-git (2021-01-16 16:22:18 UTC)
<...>
[...]

PR #139, opened 2021-01-19
Use genkernel root prefix by default on Gentoo

vozhyk-:
`genkernel` is often used to generate the initramfs on Gentoo.
An initramfs generated by genkernel accepts a ZFS root option
in the form of `root=ZFS=<dataset>`,
but does not accept `root=zfs:<dataset>`,
which is the default format for passing the root dataset
to the BE initramfs in `zfsbootmenu`.

For Gentoo BEs, make the default root prefix `root=ZFS=`,
which is accepted by both `genkernel`- and `dracut`-generated initramfs.

Users can still set the `org.zfsbootmenu:rootprefix` property
on `<pool>/ROOT` or the BE datasets to override the default.

Fixes: #136.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -740,6 +740,10 @@ find_root_prefix() {
             echo "zfs="
             break
             ;;
+          gentoo)
+            echo "root=ZFS="
+            break
+            ;;
           *)
             ;;
         esac


vozhyk-:
I have tested this manually on my Gentoo system by booting BEs with both `genkernel` and `dracut` initramfs. The root prefix property is not set, but the kernel/initramfs gets the root dataset in the expected format:
```
# zfs list -r -d 1 -o org.zfsbootmenu:rootprefix,org.zfsbootmenu:commandline xa/ROOT | uniq
ORG.ZFSBOOTMENU:ROOTPREFIX  ORG.ZFSBOOTMENU:COMMANDLINE
-                           dozfs=cache mem_sleep_default=deep
# cat /proc/cmdline
root=ZFS=xa/ROOT/gentoo-zfsbootmenu-test2 dozfs=cache mem_sleep_default=deep
# mount | grep ' / '
xa/ROOT/gentoo-zfsbootmenu-test2 on / type zfs (rw,relatime,xattr,noacl)
```

zdykstra (changes requested):
Please fix indenting to be consistent with the rest of `zfsbootmenu-lib.sh` and then this is good to go. Thank you!

zdykstra:
Your help getting this up-to-snuff for Gentoo is greatly appreciated!

vozhyk-:
Thank you too for the prompt replies and suggestions on solving problems!

Issue #133, opened 2021-01-16
/usr/lib/udev is hardcoded


vozhyk-:
When I run `generate-zbm` on Gentoo (`openrc`, `USE=split-usr` in case that makes a difference), I get:
```
# generate-zbm
Mounting /boot/efi
Creating ZFSBootMenu 1.8.1 from kernel /boot/vmlinuz-5.4.51-gentoo-x86_64
dracut-install: ERROR: installing '/usr/lib/udev/vdev_id'
dracut: FAILED: /usr/lib/dracut/dracut-install -D /var/tmp/dracut.djkjGC/initramfs -a /usr/lib/udev/vdev_id
dracut: failed to install essential executable '/usr/lib/udev/vdev_id'
Failed to create /tmp/mo7y7dzWV3/zfsbootmenu.efi
Unmounting /boot/efi
```

I see that `/usr/lib/udev` is hardcoded in https://github.com/zbm-dev/zfsbootmenu/blob/ac8a8edafbbcd8e22013de676982edcd590445d9/90zfsbootmenu/module-setup.sh . If I change every `/usr/lib/udev` to `/lib/udev`, the problem goes away.

zdykstra:
Thank you for opening this issue. I'll get a more permanent fix for you to test later today.

ahesford:
Upstream ZFS uses some placeholder strings and replaces them on install, that's probably a good approach here too.

vozhyk-:
Looks like `genkernel` (which I use for the initramfs) finds the directory using `pkg-config`: https://github.com/gentoo/genkernel/blob/369cfa98f72675a285ec98439e9054ab57b234c9/gen_funcs.sh#L1142

PR #140, opened 2021-01-21
Use inst_rules to install rule programs

vozhyk-:
`inst_rules` installs any programs rules use by itself.

Also, don't specify udev directory for rules -
`inst_rules` finds them by itself.

Fixes: #133.

### DIFF
--- a/90zfsbootmenu/module-setup.sh
+++ b/90zfsbootmenu/module-setup.sh
@@ -46,9 +46,9 @@ install() {
   local _rule _exec _ret
 
   local udev_rules=(
-    "/usr/lib/udev/rules.d/90-zfs.rules"
-    "/usr/lib/udev/rules.d/69-vdev.rules"
-    "/usr/lib/udev/rules.d/60-zvol.rules"
+    "90-zfs.rules"
+    "69-vdev.rules"
+    "60-zvol.rules"
   )
 
   for _rule in "${udev_rules[@]}"; do
@@ -59,8 +59,6 @@ install() {
   done
 
   local essential_execs=(
-    "/usr/lib/udev/vdev_id"
-    "/usr/lib/udev/zvol_id"
     "zfs"
     "zpool"
     "hostid"


vozhyk-:
I have tested this on my system: generated a ZFSBootMenu and checked that the following exist in the running ZFSBootMenu:
- `/lib/udev/rules.d/90-zfs.rules`
- `/lib/udev/rules.d/69-vdev.rules`
- `/lib/udev/rules.d/60-zvol.rules`
- `/lib/udev/vdev_id`
- `/lib/udev/zvol_id`

There is one concern, though: I see that dracut's `inst_rules` function didn't always have the functionality I'm using here, so this might not work on some older versions. I'll try to figure out exactly how old.

ahesford:
Thanks, this looks like a good solution. I'll do some testing on Void, Arch and (maybe) Ubuntu to see if there are any issues.

vozhyk-:
Parsing `IMPORT{program}` has been added to `dracut` with this commit: https://github.com/dracutdevs/dracut/commit/65d1a8a4aee600f7d5016382dd351bf9b835327f, which appears in tag `023`.

There is also a different way to do this, if we want to support an older version - using the `udevdir` variable, added in https://github.com/dracutdevs/dracut/commit/604d1eb21a6a5747173cc6cdf9292de97c239acb (in tag `019`).

I don't have any BEs on any of my computers with `dracut` older than `044`. I see that an older `zfs-0.7.0-r2` ebuild blocked `<dracut-044-r1` for some reason: https://github.com/gentoo/gentoo/commit/618255ce30bd22ccd172eb3acd641f446af7b69e. That block doesn't exist in any newer ebuilds, also for a reason I don't know.

ahesford:
Basic support is actually older than that, because that commit only alters PROGRAM statement recognition. The ZFS module relies on RUN statements, and a rule for those was already present before that commit.

I have a CentOS 7 installation with dracut 033 and have verified that it includes the necessary support.

In Void Linux, Dracut was updated to 023 in 2012. I don't think we should be too worried about supporting versions that old.

@zdykstra what do you think?

zdykstra:
This looks like the right fix to me. The Dracut mechanism used has been around for ages - I'm comfortable assuming it is there. I'm more concerned about it doing the right thing and correctly installing the udev rules without an explicit path set. I haven't had a spare second to touch this, but if it works on Void and Ubuntu I'm okay with merging it.

ahesford:
I tested this on Void and also on a dummy Ubuntu chroot that I built. It looks like everything works as expected.

PR #142, opened 2021-01-24
update README.md to solve issue 141

RoundDuckKira:
Nvidia is poorly documented and this could be a big roadblock for many to use zfsbootmenu, unless if they beg for help on the issues page or look deep. So I'm gonna add a section for Nvidia based on the advice I got from @ahesford.


ahesford:
Thanks for your contribution! I worked your suggestions into a broader edit that describes driver omission in more generic terms, and also talks about teardown hooks that help when drivers can't be disabled in ZBM but still cause problems at boot.

RoundDuckKira:
ah ok

Issue #137, opened 2021-01-17
Sort the BEs by creation time in the boot menu


vozhyk-:
Currently, the BEs seem to be sorted by name, except the default one, which is put at the beginning of the list.

Having the BEs sorted by `creation` and the default one highlighted in its place seems more natural to me - it helps make sense of a long list of BEs and see whether the default one is the newest.

```
# beadm list
BE                       Active Mountpoint Space Policy Created
gentoo-92                -      -          7.64M static 2020-03-14 21:32
gentoo-waybar            -      -          42.1M static 2020-06-06 12:55
gentoo-95                -      -          6.53M static 2020-06-13 22:27
gentoo-gnome-3.36        -      -          19.1M static 2020-06-15 12:51
gentoo-gnome-3.36-2      -      -          8.15M static 2020-06-18 09:58
gentoo-sway-1.5_rc1      -      -          24.3M static 2020-07-12 11:43
gentoo-iris              -      -          6.14M static 2020-07-12 20:15
gentoo-96                -      -          11.2M static 2020-07-12 21:04
gentoo-hplip             -      -          21.3M static 2020-08-11 18:59
gentoo-guitar            -      -          27.0M static 2020-08-24 17:12
gentoo-rt                -      -          621M  static 2020-10-02 17:51
gentoo-low-latency-2     -      -          2.06M static 2020-10-10 19:01
gentoo-rt-2              -      -          982M  static 2020-10-24 17:09
gentoo-meson-test        -      -          57.9M static 2020-10-24 23:48
gentoo-meson-test-2      -      -          27.6M static 2020-10-25 19:54
gentoo-meson-test-3      -      -          644M  static 2020-10-25 20:02
gentoo-zfs-0.8.6-test    -      -          20.9M static 2021-01-06 15:09
gentoo-zfs-0.8.9999-test -      -          44.4M static 2021-01-06 15:43
gentoo-genkernel-test    -      -          388M  static 2021-01-06 20:01
gentoo-zfsbootmenu       NR     /          63.5G static 2021-01-13 18:11
gentoo-boot-dataset      -      -          44.4M static 2021-01-16 20:42
gentoo-zfsbootmenu-test2 -      -          88.4M static 2021-01-16 23:03
```

Can an option to sort the list of BEs like that be added?

zdykstra:
That's a good idea! I'll see how feasible this is to integrate in the -lib function with a KCL toggle for it.

ahesford:
[...]

PR #143, opened 2021-02-06
Allow BE and snapshot sorting by different criteria 

zdykstra:
On systems with a large number of boot environments or snapshots, sorting by name can be less than optimal. This PR adds the following:

* Allow `zbm.sort_key` to be set on the KCL, with `name`, `creation`, `used` considered valid sort keys
* Add a hook for `mod-o` on the BE and snapshot list screens, allowing the sorting method to be changed. Changing the sorting method on one screen changes it on the other, as well. This change persists until `zfsbootmenu.sh` is exited, at which point the selected method reverts to the default (`name`) or the method defined on the KCL.

Packing the sort methods into the global `zbm_sort` variable allows easy expansion of sort keys in the future. Simply add the list of valid keys to the array in parse-commandline.sh and it will be accepted and rotated through the list.

The shortcut key combinations have been documented in the online help and the man page, but not in the footer text. I'm open to persuasion on this.

Closes https://github.com/zbm-dev/zfsbootmenu/issues/137

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -100,6 +100,13 @@ mount_zfs() {
   return ${ret}
 }
 
+# prints: first sort key
+# returns: nothing
+
+get_sort_key() {
+  echo -n "${zbm_sort%%;*}"
+}
+
 # arg1: value to substitute for empty first line (default: "enter")
 # prints: concatenated lines of stdin, joined by commas
 
@@ -196,7 +203,7 @@ draw_be() {
     "[CTRL+E] edit kcl" "[CTRL+K] kernels" "[CTRL+D] set bootfs" "[CTRL+S] snapshots" "" \
     "[CTRL+I] interactive chroot" "[CTRL+R] recovery shell" "[CTRL+P] pool status" )"
 
-  expects="--expect=alt-e,alt-k,alt-d,alt-s,alt-c,alt-r,alt-p,alt-w,alt-i,alt-l"
+  expects="--expect=alt-e,alt-k,alt-d,alt-s,alt-c,alt-r,alt-p,alt-w,alt-i,alt-o"
 
   if ! selected="$( ${FUZZYSEL} -0 --prompt "BE > " \
       ${expects} ${expects//alt-/ctrl-} ${expects//alt-/ctrl-alt-} \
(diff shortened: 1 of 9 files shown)


vozhyk-:
I have tried this on my system, and it appears to work correctly:
- Adding `zbm.sort_key=creation` to the zfsbootmenu kernel command line does make it sort the BEs by creation time (with the active BE on top).
- Pressing `Control-o` in the boot menu does cycle between sorting by `creation`, `name`, and `used`.

zdykstra:
> I have tried this on my system, and it appears to work correctly:
> 
>     * Adding `zbm.sort_key=creation` to the zfsbootmenu kernel command line does make it sort the BEs by creation time (with the active BE on top).
> 
>     * Pressing `Control-o` in the boot menu does cycle between sorting by `creation`, `name`, and `used`.

Thank you for testing this! Pending a merge and official release, does this otherwise meet your needs?

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -100,6 +100,16 @@ mount_zfs() {
...
 }
 
+# prints: first sort key
+# returns: nothing
+
+get_sort_key() {
+  local sort_key
+  sort_key="${zbm_sort%%;*}"
+  zdebug "Using sorting key ${sort_key}"
+  echo -n "${sort_key}"
```
ahesford, line 110: Maybe `echo -n "${sort_key:-name}"` to prevent degenerate output.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -253,20 +263,22 @@ draw_kernel() {
...
+
   header="$( header_wrap \
     "[ENTER] duplicate" "[ESC] back" "[CTRL+H] help" "" \
     "[CTRL+X] clone and promote" "[CTRL+C] clone only" "" \
     "[CTRL+I] interactive chroot" "[CTRL+D] show diff" )"
 
-  expects="--expect=alt-x,alt-c,alt-d,alt-i"
+  expects="--expect=alt-x,alt-c,alt-d,alt-i,alt-o"
 
-  if ! selected="$( zfs list -t snapshot -H -o name "${benv}" |
```
ahesford, line 269: If the input is always sanitized, this is fine. Alternatively, what do you think about a `zfs_list_sorted` function that runs `zfs list -S "$( get_sort_key )" "@"` and prints the output on success, otherwise falls back to `zfs list "@"` in failure?

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Good call, done.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: There are two places currently where that'd be helpful, `populate_be_list()` and `draw_snapshots()`.  I'm not sure the added complexity is worth it for that, but I'm not opposed to it either.

ahesford: I don't feel strongly about it. The main value would be in allowing arbitrary user fields without verifying them against an approved list. As long as there is a pre-approved list, the behavior should be correct without a fallback.

zdykstra: I think I'm going to punt on this for now.

PR #144, opened 2021-02-07
Move more helper functions to -lib, cleanup scripts

### LINKED ISSUE #115: Move colorize() to zfsbootmenu-lib.sh
Instead of embedding this in `zfsbootmenu-help.sh`, this function should be available for everything in the ZBM ecosystem to use - notably `zfs-chroot`.

zdykstra:
Move the helper functions from zfsbootmenu-help.sh:

* center
* colorize

`zfsbootmenu-preview.sh` had a lot of legacy code / procedures in it, before `zfsbootmenu-lib.sh` was as fleshed out as it is now. Clean up the usage / flow dramatically to use helper functions where possible.

`zfs-chroot.sh` now colorizes text when entering the chroot to add 37 pieces of flair to the output.

`timed_prompt`, `resume_prompt`, and the snapshot manipulation user input sections in `zfsbootmenu.sh` could possibly use an update to use `colorize` - but with the possible future use of `dialog`, that might be wasted effort. Thoughts?

Closes #115

### DIFF
--- a/90zfsbootmenu/zfs-chroot.sh
+++ b/90zfsbootmenu/zfs-chroot.sh
@@ -22,12 +22,12 @@ if mountpoint="$( allow_rw=yes mount_zfs "${selected}" )"; then
 
   # Snapshots and read-only pools always produce read-only mounts
   if [[ "${selected}" =~ @ ]] || ! is_writable "${pool}"; then
-    writemode="read-only"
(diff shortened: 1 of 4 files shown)


ahesford:
If you want 37 pieces of flair, why don't you just make that the minimum?

PR #146, opened 2021-02-16
Logging lifecycle

zdykstra:
This PR is an attempt to make logging play a more focused role in the ZBM ecosystem.  Various components would correctly catch and log an error through `zerror()`, but unless you know how to look for them, they were entirely hidden from view. As such, I attempted to meet the following goals:

* Where an error is critical, (kexec, duplicating a BE), show a prompt and log an error
* For all other error conditions, log an error with enough contextual information
* When an error is generated, show an indicator on the main screen
* Allow the error log to be viewed on any selector screen via `MOD-L`

In addition to the error log integrations, I made a pass through virtually every function in `zfsbootmenu-lib.sh`.  Where positional parameters are non-optional in a function call, a test for empty is performed for each parameter. If the parameter is empty, an error is logged and and the function returns. Some functions had a check / silent return, others did not. They've all been normalized.

I've tested the behavior with an x86_64 BE in a ppc64le VM to confirm kexec and chroot error handling.  

Are there any other error conditions that should warrant a `timed_prompt`, or should the current usage be pared back?

One side effect of this work is that the heavy usage of `zdebug()` does have a performance impact, so booting ZBM with `loglevel=7`should now only be done if you actually need debug level logging. Nothing is free.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -850,7 +850,7 @@ select_kernel() {
     done <<<"$( tac "${BASE}/${zfsbe}/kernels" )"
   fi
 
-  zdebug "using kexec args: ${spec_kexec_args}"
+  zdebug "using kexec args: ${kexec_args}"
   echo "${kexec_args}"
 }
 
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -22,7 +22,6 @@ zlog() {
   _func="${FUNCNAME[2]}"
 
   WIDTH="$( tput cols )"
-
   echo -e "<${1}>ZBM:\033[0;33m${_script}[$$]\033[0;31m:${_func}()\033[0m: ${2}" | fold -s -w "${WIDTH}" > /dev/kmsg
 }
 
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -257,7 +257,7 @@ draw_be() {
 
   [ -r "${env}" ] || return 130
 
-  header="$( header_wrap "[ENTER] boot" "[ESC] refresh view" "[CTRL+H] help" "" \
+  header="$( header_wrap "[ENTER] boot" "[ESC] refresh view" "[CTRL+H] help" "[CTRL-L] error log" "" \
     "[CTRL+E] edit kcl" "[CTRL+K] kernels" "[CTRL+D] set bootfs" "[CTRL+S] snapshots" "" \
     "[CTRL+I] interactive chroot" "[CTRL+R] recovery shell" "[CTRL+P] pool status" )"
 
@@ -293,7 +293,7 @@ draw_kernel() {
   [ -r "${_kernels}" ] || return 130
 
   header="$( header_wrap \
-    "[ENTER] boot" "[ESC] back" "" "[CTRL+D] set default" "[CTRL+H] help" )"
+    "[ENTER] boot" "[ESC] back" "" "[CTRL+D] set default" "[CTRL+H] help" "[CTRL-L] error log" )"
 
   expects="--expect=alt-d"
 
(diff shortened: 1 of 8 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -250,12 +258,19 @@ draw_be() {
...
+  if [ ! -r "${env}" ] ; then
+    zerror "environment file ${env} is missing"
+    return 130
+  fi
 
-  [ -r "${env}" ] || return 130
+  zdebug "using environment file: ${env}"
 
-  header="$( header_wrap "[ENTER] boot" "[ESC] refresh view" "[CTRL+H] help" "" \
+  header="$( header_wrap "[ENTER] boot" "[ESC] refresh view" "[CTRL+H] help" "[CTRL-L] error log" "" \
```
ahesford, line 273: `CTRL+L`

zdykstra: Whoops.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -486,8 +516,18 @@ kexec_kernel() {
     unset tdhook
   fi
 
+  # restore kernel log level just before we kexec
+  # shellcheck disable=SC2154
+  if [ -n "${PRINTK}" ] ; then
+    echo "${PRINTK}" > /proc/sys/kernel/printk
+    zdebug "restored kernel log level to ${PRINTK}"
+  fi
```
ahesford, line 524: What is the value of this, anyway?

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -447,20 +480,17 @@ kexec_kernel() {
...
   if ! output="$( kexec -l "${mnt}${kernel}" \
     --initrd="${mnt}${initramfs}" \
     --command-line="${root_prefix}${fs} ${cli_args}" 2>&1 )"
   then
     zerror "unable to load ${mnt}${kernel} and ${mnt}${initramfs} into memory"
     zerror "${output}"
     umount "${mnt}"
+    color=red delay=10 timed_prompt "Unable to load kernel or initramfs into memory" \
+      "${mnt}${kernel}" \
+      "${mnt}${initramfs}"
```
ahesford, line 492: Would it be reasonable to drop this BE from the menu after this message? There's no reason to expect a later attempt to load will suddenly succeed.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -486,8 +516,18 @@ kexec_kernel() {
...
+    echo "${PRINTK}" > /proc/sys/kernel/printk
+    zdebug "restored kernel log level to ${PRINTK}"
+  fi
+
   if ! output="$( kexec -e -i 2>&1 )"; then
-    zdebug "${output}"
+    zerror "kexec -e -i failed!"
+    zerror "${output}"
+    echo 0 > /proc/sys/kernel/printk
+    color=red delay=10 timed_prompt "kexec run of ${kernel} failed!"
```
ahesford, line 530: Again, it might be worth dropping the BE from the menu if it can't be kexec'ed.

ahesford (approved):
This looks good in general. We can run it through some more rigorous testing as part of the next release.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: It's debatable, honestly. I'm not sure there's anything really logged when the kexec happens, but before the new kernel actually takes over. It can probably just be thrown away, along with storing `PRINTK` originally.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: The only thing I can think of is that maybe you'd chroot into the BE and fix whatever is wrong, but that's a bit of a stretch. I'm okay with discarding the BE after a failed kexec load, or kexec execute.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Stripped out storing the original value and restoring it just before kexec.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: There's no readily available mechanism to remove what is (by current tests) a valid BE. We'll need to establish how to blacklist a BE at runtime in a new PR/branch.

PR #147, opened 2021-02-23
Add support to discover and assume a hostid, as well as fix arguments passed to a BE

zdykstra:
A persistent problem with ZFS and now with ZFSBootMenu is getting `spl.spl_hostid` lined up on all pool imports. The contents of an initramfs and the kernel command line can influence this value, making it difficult to determine exactly where and how to change it to get a system to boot. 

To help combat this recurring issue, ZBM now accepts the following options to control importing behavior. 

* `zbm.import_policy=strict`: If no pools can be imported with the hostid embedded in the initramfs, or set via `spl_hostid` on the ZBM command line, drop to a recovery shell.
* `zbm.import_policy=hostid`: If no pools can be imported with the hostid embedded in the initramfs, or set via `spl_hostid` on the ZBM command line, attempt to scrape the hostid used to import the pool defined in `root=zfsbootmenu:POOL=<value>` and import the pool. If that pool is missing or can't be imported, any other available pool is scraped for a hostid and attempted to be imported.
* `zbm.import_policy=force`: YOLO mode, just force import a pool. Replaces `zbm.force_import'.

Right now, when undefined, `zbm.import_policy` is set to `strict`.

With this option available, ZBM can now in most cases import any pool - regardless of the `hostid` of the system that built the initramfs or the EFI executable. 

Next is to ensure that the boot environment imports the pool correctly. Some ZFS initramfs modules can use `spl_hostid` to write out `/etc/hostid` in the initramfs before loading SPL/ZFS modules. Others ... like Arch ... not only do not by default include `/etc/hostid` in the initramfs, they do not have any methods to line up `spl_hostid` on the kernel command line with `/etc/hostid`. However, `spl.spl_hostid` takes precedence over `/etc/hostid`, so we can simply set this on the kernel command line. To this end, we now:

* Suppress `spl_hostid` if it's part of the discovered command line for a boot environment
* Suppress `spl.spl_hostid` if it's part of the discovered command line for a boot environment
* Insert `spl.spl_hostid=0x<hostid>` on the command line for each boot environment.

Since we know the `hostid` used to import a pool - either discovered in `zfsbootmenu-parse-commandline.sh`, or discovered when `zbm.import_policy=hostid` is set.  Suppression / rewriting can be disabled via `zbm.set_hostid=0` when booting ZBM. 

I'm unsure if we should default this behavior to on, or make it opt-in initially for people installing ZBM from `master` or Void packages.  When building binary releases, `zbm.import_policy=hostid` and `zbm.set_hostid=1` should be set, to be maximally usable for people randomly downloading it to test.


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -1012,39 +1123,22 @@ load_be_cmdline() {
...
+  fi
 
-          # If output is not suppressed, print the field
-          if ( supp == 0 && length($(i)) > 0 ) {
-            print $(i);
-          }
+  # shellcheck disable=SC2154
+  if [ "${zbm_set_hostid}" -eq 1 ]; then
+    # Try to assume a matched hostid if none is defined
+    if [ -z "${spl_hostid}" ] && [ -r "${BASE}/spl_hostid" ]; then
```
ahesford, line 1132: It would be nice to grab this from /etc/hostid instead of a byproduct of host matching, but reading the binary /etc/hostid requires an external tool. If we install `od`, we can grab it with `od -t x -An /etc/hostid`. However, reading it from this byproduct paves the way for per-BE hostids if we want them.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -1012,39 +1123,22 @@ load_be_cmdline() {
...
-          }
-        }
-        printf "noresume";
-      }
-    ' )"
+    # Only override hostid arguments if one can be provided
+    if [ -n "${spl_hostid}" ]; then
+      zdebug "overriding spl_hostid and spl.spl_hostid in: ${zfsbe_args}"
+      zfsbe_args="$( suppress_kcl_arg spl_hostid "${zfsbe_args}" )"
+      zfsbe_args="$( suppress_kcl_arg spl.spl_hostid "${zfsbe_args}" ) spl.spl_hostid=0x${spl_hostid}"
```
ahesford, line 1140: I put this at the end just in case we miss something somehow; I think the kernel prefers the last value.

ahesford (approved):
This is a good solution to a long-standing problem. When we distribute a UEFI bundle, it should set `zbm.import_policy=hostid` and not disable `zbm.set_hostid`. We can advise people who use prebuilt initramfs/kernel combinations to set these by whatever means they use.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: `od` is part of coreutils, I don't think it's unreasonable to require it - especially if we want to do the right thing on MUSL systems.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Good call.

PR #148, opened 2021-03-03
Support for testing of multiple distributions

ahesford:
I think this ready to be squashed and moved to master.

The ZFSBootMenu testing environment now supports any distribution by providing scripts

    helpers/install-<distro>.sh
    helpers/chroot-<distro>.sh

The `helpers/image.sh` script will set up a pool in a qemu disk image, call the per-distro install script to initially populate it, then set up a chroot environment that will run the corresponding chroot script for final setup. Optionally, per-distro image scripts can be installed at

    helpers/image-<distro>.sh

to override the standard generic behavior.

For starters, this framework is used to provide Void glibc, Void musl, Arch and Debian test environments.


zdykstra (approved):
I think realistically the only thing missing is adding an encryption key to the Debian initramfs. This isn't a huge concern though, since that's largely just exercising the Debian initramfs generator at that point and not ZBM.

ahesford:
Closed in 36a29ccc0d4931118aee2eed98a1fe8ca139ef93

PR #150, opened 2021-03-05
Support old versions of craaaaap

### LINKED ISSUE #149: Kernel-log viewer issues
The new kernel log viewer has two issues that should be addressed before the next release:
1. The `fzf` binding for this (`Ctrl-L`) uses the `refresh-preview` action, which is relatively new and not provided in `fzf-0.20.0` included with Ubuntu Focal Fossa. This causes a hard failure when trying to show the boot menu, with the screen repeatedly printing "unknown action: refresh-preview" because `fzf` is dying.
2. Warnings don't seem to be shown properly with Ubuntu's dmesg. With a `readonly=on` pool, select a BE and use `Ctrl+I` to get an interactive (read-only) chroot. When returning to the menu screen, a yellow `[!]` appears in the preview pane to indicate a warning, but pressing `Ctrl+L` shows an empty log. Dropping to a recovery shell and looking at `dmesg` reveals the warning is "ZFSBootMenu: read-write mount of ztest/ROOT/ubuntu forbidden, pool ztest is not writable".

Regarding issue (1), I don't really have a good solution. Without `refresh-preview` in the binding, the warning/error indicator remains after viewing logs, until the menu selection changes or `Esc` is used to refresh the view by killing `fzf`. It is obviously desirable to update the preview after leaving the log viewer. We could do version comparisons (horrible) or use a command like
```bash
echo "abc" | fzf -f "abc" --bind "alt-l:refresh-preview" --exit-0 >/dev/null 2>&1
```
which seems to return 0 (when the "abc" strings match), 1 when the string isn't found, and 2 when `refresh-preview` isn't a recognized action. This is non-ideal but is less horrible than version comparisons. We can do it at runtime when picking the fuzzy finder or we could do it at module-setup time, but we'd have to provide some kind of trigger file to indicate the refresh-preview is available and that starts to look hacky.

Regarding issue (2), I believe this is caused because Ubuntu's `dmesg` does not support `--noescape`, so `dmesg` errors out when invoked. I believe all other options we use for log review are supported there. Looks like we're back to more feature discovery.

@zdykstra

zdykstra:
Check if dmesg supports the '--noescape' flag
Check if fzf supports the 'refresh-preview' action

Create /etc/zfsbootmenu.conf in the initramfs with exports, so the file can be sourced in the initramfs. Variables in here should be considered feature flags - if the variable is defined with any value, the flag is set to true. If the variable is undefined, consider the feature disabled.

BYTE_ORDER: influence how /etc/hostid is written
HAS_NOESCAPE: enable color escape codes in debug-level messages / parse them in zlogtail
HAS_REFRESH: fzf is new enough to support the refresh-preview action. This has the added benefit of un-breaking skim support, again

Fixes #149

### DIFF
--- a/90zfsbootmenu/module-setup.sh
+++ b/90zfsbootmenu/module-setup.sh
@@ -215,9 +215,6 @@ install() {
     endian="le"
   fi
 
-  # shellcheck disable=SC2154
-  echo -n "${endian}" > "${initdir}/etc/byte-order"
-
   # Try to synchronize hostid between host and ZFSBootMenu
   #
   # DEPRECATION NOTICE: on musl systems, zfs < 2.0 produced a bad hostid in
@@ -254,4 +251,29 @@ install() {
   if [ -e "${initdir}/etc/hostid" ] && type mark_hostonly >/dev/null 2>&1; then
     mark_hostonly /etc/hostid
   fi
+
+  # Check if dmesg supports --noescape
+  if dmesg --noescape -V >/dev/null 2>&1 ; then
+    has_escape=1
+  else
+    has_escape=
+  fi
+
+  # Check if fzf supports the refresh-preview flag
+  # Added in fzf 0.22.0
+  if command -v fzf >/dev/null 2>&1 && \
+    echo "abc" | fzf -f "abc" --bind "alt-l:refresh-preview" --exit-0 >/dev/null 2>&1 
+  then
+    has_refresh=1
+  else
+    has_refresh=
+  fi
+
+  # Collect all of our build-time feature flags
+  # shellcheck disable=SC2154
+  cat << EOF > "${initdir}/etc/zfsbootmenu.conf"
+export BYTE_ORDER=${endian}
+export HAS_NOESCAPE=${has_escape}
+export HAS_REFRESH=${has_refresh}
+EOF
 }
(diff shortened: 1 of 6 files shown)


zdykstra:
I should note that I haven't actually tested building an initramfs on Ubuntu with these fixes!

ahesford (approved):
I've tested images produced in Debian, Ubuntu and Arch. All produce working images. In Debian, fzf is v0.17.5, which causes two problems I've seen:
1. `zlogtail.sh` uses the `--no-info` option, which means `zlogtail` fails immediately.
2. `zfsbootmenu-help.sh` uses `--preview-window="right:${PREVIEW_SIZE}:sharp:wrap"`, which causes old fzf to repeatedly print "invalid preview window layout: right:64:sharp:wrap" rather than showing the expected help screen.

Either additional feature tests are the order of the day, or we let these slide because they aren't core features and we recommned a sufficiently new fzf to get things working properly.

Neither of these issues has any bearing on the fixes here.

zdykstra:
I've flipped zlogtail to use '--inline-info' which is supported in `sk` and `fzf` going back for quite some time. It's not bad to see how many log lines your current filter actually matches.

I removed both references to `:sharp` in for preview window format strings - it's a minor cosmetic change that isn't worth it to set a feature flag to preserve.

PR #151, opened 2021-03-07
Fix PS1 in arch/debian/ubuntu, set custom prompt in emergency shell

zdykstra:
Seems to do the right thing in Arch/Ubuntu/Debian/Void chroots, and in the emergency shell.


ahesford (approved):
This is probably the right thing to do anyway. Setting up the full bash environment in a chroot into a BE that hasn't been properly initialized seems inappropriate.

PR #153, opened 2021-03-07
Support reuse of existing pool images in setup.sh

ahesford:
It is now easy to combine multiple distributions into a single test image:
```bash
./setup.sh -a
./setup.sh -i -x -o arch
./setup.sh -i -x -o debian
```

I haven't tested the encryption path yet, but the basic setup seems to work. I'm open to alternative implementations that are a bit cleaner.

### DIFF
--- a/testing/.gitignore
+++ b/testing/.gitignore
@@ -9,3 +9,4 @@ dracut
 dracut.conf.d
 generate-zbm
 test.*
+keys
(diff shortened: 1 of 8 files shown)


zdykstra (approved):

zdykstra:
This seems like a pretty reasonable way to solve the problem - and the code flow is reasonable/readable. I haven't actually tested this, but the underlying idea here is good.

ahesford:
Closed by 58f5dd06

PR #155, opened 2021-03-10
Use rootfs tarball and preinstall ZBM in Void testbeds

ahesford:
In keeping with earlier changes to Arch, Debian and Ubuntu, the Void images should be installable without any specific packaging tools available on the host. Using a rootfs tarball allows us to populate Void testbeds from any host. I've also disabled architecture limitations since the only XBPS magic happens within a chroot now, and if the host can chroot into the target, everything should work fine.

### DIFF
--- a/testing/helpers/chroot-void.sh
+++ b/testing/helpers/chroot-void.sh
@@ -1,23 +1,21 @@
 #!/bin/sh
 # vim: softtabstop=2 shiftwidth=2 expandtab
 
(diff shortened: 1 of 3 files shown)


zdykstra (approved):
I like that we can specify an exact kernel now, instead of chasing what the basesystem ships. Is it possible to cache the rootfs tarball and use it if the copy on the repo is the same as what we have in `.cache/` ? That'd be nice to avoid downloading that repeatedly.

PR #156, opened 2021-03-11
Support caching of downloads when building test environments

ahesford:
Tarball downloads can be cached using the extract_remote.sh helper, and image.sh will bind-mount mount a cache directory (when one exists) into each chroot during setup for arbitrary use and retention across setups.

Currently the Arch and Void installation scripts use the bind-mounted host directory as a package cache. Debian and Ubuntu make no use of the cache.

### DIFF
--- a/testing/.gitignore
+++ b/testing/.gitignore
@@ -10,3 +10,4 @@ dracut.conf.d
 generate-zbm
 test.*
 keys
+cache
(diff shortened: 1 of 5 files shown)


zdykstra (approved):
Once I created `testing/cache`, this worked like a charm. It helped with VM creation times once the cache had been populated for that specific OS.

PR #158, opened 2021-03-11
Create SSH includes for running VMs

zdykstra:
SSH'ing into a running VM was greatly improved with key additions to each BE. This should help even more by doing the following:

* Creates an SSH config file for the running VM that records the test name and configured SSH port forward
* Sets the remote username to `root`
* Host key checks are disabled, and the host key is no longer stored in `known_hosts`

With this in place, you can `ssh test.void` and have an immediate shell in the running test VM with out needing to accept any host keys, etc.

### DIFF
--- a/testing/run.sh
+++ b/testing/run.sh
@@ -197,6 +197,9 @@ Host ${TESTDIR}
   HostName localhost
   Port ${SSH_PORT}
   User root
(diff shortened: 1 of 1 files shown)


ahesford (approved):
Nice idea. I've tested with environments in `/tmp` and it works as expected.

PR #160, opened 2021-03-14
Binary release helpers / run-time changes

zdykstra:
This PR comprises two major sections - the releng helper changes and ZFSBootMenu internal changes. 

releng:

* Add `releng/make-binary.sh` to create a clean room EFI and components build
* Attach the assets along with an sha256sum of each to the release via `gh` (can we retire hub now?)

ZBM internals:

To better support a kernel command line for these binary releases, a new mechanism has been added to `module-setup.sh` and `zfsbootmenu-parse-commandline.sh`. Because we're setting `rd.hostonly=0` on the command line, Dracut is removing anything tagged as a `host_only` file, including a customized file in `/etc/cmdline.d`. Since `getarg` provided by Dracut builds the kernel command line from multiple sources every single call, we can simply put a file in `/etc/cmdline.d` after `99base` has executed but before anything in ZFSBootMenu tries to parse the command line.

 If `embedded_kcl` is defined as a variable when Dracut is run, the variable and it's contents are written to `/etc/zfsbootmenu.conf` in the initramfs.  This is then written back out to `/etc/cmdline.d/zfsbootmenu.conf` to trick `getarg`-based calls to use our customized commandline. This is a hack.


zdykstra:
It might be ideal to put a prompt forcing user input in `tag-release.sh` just before the release is actually committed. This will allow for a final test of the binary build artifacts, if we choose to do that.

releng/tag-release.sh
```diff
@@ -90,11 +90,37 @@ if echo "${release}" | grep -q "[A-Za-z]"; then
...
+  echo "sha256sum asset not found: ${shasum_asset}"
+  exit 1
+fi
+
 # Use github-cli or hub to push the release
 if command -v gh >/dev/null 2>&1; then
   # github-cli does not automatically strip header that hub uses for a title
   sed -i '1,/^$/d' "${relnotes}"
-  gh release create "${tag}" ${prerelease} -F "${relnotes}" -t "ZFSBootMenu ${tag}"
+  gh release create "${tag}" ${prerelease} -F "${relnotes}" -t "ZFSBootMenu ${tag}" "${efi_asset}" "${components_asset}" "${shasum_asset}"
```
ahesford, line 123: If we want to support `hub` as well, it looks like it uses `-a "${file}"` to attach an asset.

Otherwise, we can just drop the `hub` fallback and go all-in on `gh`.

releng/make-binary.sh
```diff
@@ -0,0 +1,80 @@
...
+
+components="${build}/zfsbootmenu-${release}"
+mkdir -p "${components}"
+mv "${build}/initramfs-bootmenu.img" "${components}"
+mv "${build}/vmlinuz-bootmenu" "${components}"
+
+cd "${build}" || exit 1
+tar -czvf "${assets}/zfsbootmenu-${release}.tar.gz" "$( basename "${components}" )"
+
+cd "${assets}" || exit 1
```
ahesford, line 77: You can drop the `realpath` in line 64 if you
```shell
( cd "${build}" && tar czvf "${assets}/zfsbootmenu-${release}.tar.gz" "$( basename "${components}" )" ) || exit 1
( cd "${assets}" && rm -f sha256sum.txt && sha256sum -- * > sha256sum.txt ) || exit 1
```

releng/make-binary.sh
```diff
@@ -0,0 +1,80 @@
...
+yq-go eval "del(.Kernel.CommandLine)" -i "${yamlconf}"
+
+"${temp}/generate-zbm" \
+  --version "${release}" \
+  --config "${yamlconf}" \
+  --cmdline "loglevel=4 nomodeset"
+
+assets="$( realpath -e releng )/assets/${release}/"
+
+mkdir -p "${assets}"
```
ahesford, line 66: Maybe this should go right after the [Dracut test in Lines 18--20](https://github.com/zbm-dev/zfsbootmenu/pull/160/files#diff-e6dc91c779edbf33adba67870152f47e59f610e2ff69a3c1d881476fe924f4a8R18-R22) and then do
```shell
rm -f "${assets}/vmlinuz-bootmenu" \
      "${assets}/initramfs-bootmenu.img \
      "${assets}/zfsbootmenu-${release}.EFI" \
      "${assets}/zfsbootmenu-${release}.tar.gz"
```
so there won't be any lingering components around.

releng/make-binary.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+# vim: softtabstop=2 shiftwidth=2 expandtab
+
+cleanup() {
+  test -d "${temp}" && rm -rf "${temp}"
+  exit
+}
+
```
ahesford, line 8: Should there be a `set -e` here to trigger a hard failure for any uncaught nonzero command return?

ahesford (approved):

releng/make-binary.sh
zdykstra: I'll try that tonight when I test my changes

PR #163, opened 2021-04-02
Set a hostname in ZBM, check if spl.spl_hostid matches /etc/hostid

zdykstra:
* Set a hostname in ZBM, so that when a pool is imported read/write, `(ZFSBootMenu)` is recorded instead of `(none)`

* In generate-zbm, warn if  `/sys/module/spl/parameters/spl_hostid`and the output of `hostid` do not match. This needs additional logic, since very few people set their hostid via `spl.spl_hostid`, ZFS's brain-dead logic of setting the sysfs entry to 0 by default makes it very challenging to detect what the source of truth actually is. Thoughts?

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-exec.sh
+++ b/90zfsbootmenu/zfsbootmenu-exec.sh
@@ -35,6 +35,9 @@ EOF
 
 getcmdline > "${BASE}/zbm.cmdline"
 
+# Set a non-empty hostname so we show up in zpool history correctly
+echo "ZFSBootMenu" > /proc/sys/kernel/hostname
+
 modprobe zfs 2>/dev/null
 udevadm settle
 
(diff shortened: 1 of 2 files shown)


ahesford:
Regarding Point 2:
- Using `hostid` is not reliable since that program always returns `0` on musl. `generate-zbm` will have to read four bytes directly from `/etc/hostid` to check. If Perl reads this as an `int32`, do we need to worry about endianness?
- I think the logic should be simple here; if the `sysfs` parameter value is zero, just stop, because `/etc/hostid` wins in the SPL. Otherwise, compare and complain when different.

zdykstra:
This should be good to go now on both little-endian and big-endian systems.

ahesford (approved):
Works as expected in the test apparatus. Importing a pool read/write in ZBM and then restarting the emulator and booting a BE with another hostid causes the import to fail and list "ZFSBootMenu" as the last host to access the pool.

Forcing `spl.spl_hostid` to something other than the value in `/etc/hostid` causes `generate-zbm` to complain about mismatch.

PR #166, opened 2021-04-09
Make sure wrapped log lines have a ZBM/tracing prefix

zdykstra:
Long messages, particularly debug messages, were naively wrapped on the
screen width boundary. While this worked, all tracing information was
lost in the wrapped lines.

A log line prefix (based on the message level and dmesg capabilities) is
created and then prepended to all log lines after they've been wrapped.

Wrapping uses the screen width minus the prefix length minus the kernel
timestamp and fzf gutter.

Wrapped lines have '+   ' after the prefix to indicate that they're wrapped.
Non-wrapped lines have ': ' after the prefix.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -9,7 +9,7 @@
 # returns: nothing
 
 zlog() {
-  local _script _func
+  local _script _func prefix offset join
   [ -z "${1}" ] && return
   [ -z "${2}" ] && return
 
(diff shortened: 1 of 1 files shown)


zdykstra:
I'm open to formatting tweaks - I just want wrapped lines to not be lost when I'm tracing function calls via a filter.

ahesford (approved):
It would be nice to make `offset - 15` dynamic in the `_WIDTH` definition, but this is a nice change that solves a problem I've seen in the log browser.

zdykstra:
> It would be nice to make `offset - 15` dynamic in the `_WIDTH` definition, but this is a nice change that solves a problem I've seen in the log browser.

I'm happy to attempt something for this - I'm just not quite sure what should be made dynamic!

ahesford:
Please disregard; I was thinking the "gutter" you reference in the comment was the width of a left column like in the help browser. I forgot that the log viewer consumes the entire width.

PR #167, opened 2021-04-10
Preserve selected item

zdykstra:
When entering a chroot or the diff browser on the snapshot selection screen, preserve the snapshot selection when exiting back out of that action. 

This is accomplished by creating a simple shell script that accepts two or more arguments - the first is the name of the zfsbotmenu-lib.sh function, the rest of the arguments are passed on to that function. This shell script is then executed via `--bind` actions. When the script exits, the original fzf snapshot browser is still running, with the snapshot still selected.

I opted to write out the explicit list of binds for the chroot and diff actions; I wasn't able to get the quoting to work correctly when trying to represent the entire string as a variable for key regexps.


ahesford (approved):
Looks good to me, nice to move the processing out of the main ZBM event loop.

PR #169, opened 2021-04-14
Create and use /etc/profile everywhere

zdykstra:
`bash` login shells automagically source `/etc/profile` . 

* `dropbear` invokes bash as a login shell 
* `tmux` invokes bash as a login shell. 
* Switch the recovery shell to launch bash as a login shell.

Set a universal path, a generic terminal type, PS1, and source `/lib/zfsbootmenu-lib.sh`. Write all variables defined/controlled by the command line to this file. In both main ZBM entry point scripts (`zfsbootmenu.sh`, `zfsbootmenu-countdown.sh`),  source this file or source `/lib/zfsbootmenu-lib.sh` and warn that the environment file couldn't be sourced.

Any shell - tmux, recovery, or SSH should now have an identical environment.

### DIFF
--- a/90zfsbootmenu/module-setup.sh
+++ b/90zfsbootmenu/module-setup.sh
@@ -307,9 +307,9 @@ EOF
 
   # Setup a default environment for all login shells
   cat << EOF >> "${initdir}/etc/profile"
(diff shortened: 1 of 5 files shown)


ahesford (approved):
I think this is a clean solution to a few environment issues we've seen. Looks good.

PR #170, opened 2021-04-23
Move documentation to discrete files

zdykstra:
Instead of embedding documentation in `zfsbootmenu-help.sh`, it will now find and display files on disk. This lets us embed zfsbootmenu.7.pod in the initramfs so that the full system integration documentation is available at runtime. 

The downside to this is that there's no reasonable way to fold/wrap these lines. `releng/pod2help.sh` now pre-computes three different sizes of documentation - suitable for 80, 120 and 160 column displays. Extremely narrow displays will fall back to the documentation formatted for 80 character displays. It won't look great, but it'll be there.

The file display name in the help menu is extracted from the `zfsbootmenu - <description>` line in each file. The first few lines are then stripped to normalize the display of man pages and help pages.

[Preview](https://pastebin.stratumzero.date/u/p/1a6576b3dd/stream)


ahesford (approved):
It would be nice to have a truly dynamic solution that is aware of color sequences, but I don't see one.

zdykstra:
I gave it the old college try in https://github.com/zdykstra/pod2ansi . I'm done with that avenue.

PR #172, opened 2021-04-28
Force spl.spl_hostid=0 when matching hostid

ahesford:
If the SPL kmod has a non-zero `spl_hostid` parameter, nothing we write to `/etc/hostid` will have any effect. Forcing the module parameter to zero will cause ZFS to look to `/etc/hostid` for a value regardless of the initial state.

Needs testing to confirm that the parameter is writable on all systems, especially on the geriatric Ubuntu ZFS.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -122,7 +122,8 @@ center_string() {
 # returns: 0 on successful write, 1 on error
 
 write_hostid() {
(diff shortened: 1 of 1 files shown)


zdykstra:
Pending testing on ZFS <2.0, this looks good. It still grates on me that a value of 0 causes it to look at `/etc/hostid` - that was a bad upstream design choice.

zdykstra:
Works like a charm under ZFS 0.8.3!

PR #173, opened 2021-05-04
Allow references to parent properties in org.zfsbootmenu:commandline

ahesford:
Any reference to `%{parent}` in `org.zfsbootmenu:commandline` will be replaced with the value of the same property on the parent filesystem (with parent references above recursively expanded), allowing easy specification of common options at a mutual parent of two BEs and overrides or additions of individual options per-BE. The value of `%{parent}` is always an empty string on a root filesystem.

This is not intended to be sophisticated, and `%{parent}` appearing within other words will be replaced regardless. The assumption is that `%{parent}` is unique enough and will not conflict with real KCL options, so dumb global replacement is sufficient.

Tested with the following setup:
```
# zfs list -o name,org.zfsbootmenu:commandline
NAME             ORG.ZFSBOOTMENU:COMMANDLINE
ztest            loglevel=4 spl_hostid=43b06f17 %{parent}
ztest/ROOT       %{parent} rw console=tty1 console=ttyS0
ztest/ROOT/void  %{parent} loglevel=7
```
The command-line as parsed by ZBM appears in the preview pane as
```
loglevel=4 spl_hostid=43b06f17  rw console=tty1 console=ttyS0 loglevel=7
```
which is the expected value. (Note the double space after the `spl_hostid` argument, from the space before the empty `%{parent}` on `ztest` and the space following the expanded `%{parent}` on `ztest/ROOT`.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -1149,6 +1149,53 @@ find_root_prefix() {
 }
 
 # arg1: ZFS filesystem
+# prints: value of org.zfsbootmenu:commandline, with %{parent} recursively expanded
+# returns: 0 on success
+
+read_kcl_prop() {
+  local zfsbe args parfs par_args
+
+  zfsbe="${1}"
+  if [ -z "${zfsbe}" ]; then
+    zerror "zfsbe is undefined"
+    return 1
+  fi
+
+  if ! args="$( zfs get -H -o value org.zfsbootmenu:commandline "${zfsbe}" )"; then
+    zerror "unable to read org.zfsbootmenu:commandline on ${zfsbe}"
+    return 1
+  fi
+
+  # KCL is empty, nothing to see
+  if [ "${args}" = "-" ]; then
+    echo ""
+    return 0
+  fi
+
+  # KCL does not specify parent inheritance, just return the args
+  if ! [[ "${args}" =~ "%{parent}" ]]; then
+    echo "${args}"
+    return 0
+  fi
+
+  # Need to recursively expand "%{parent}"
+
+  parfs="${zfsbe%/*}"
+  if [ -z "${parfs}" ] || [ "${parfs}" = "${zfsbe}" ]; then
+    # There is no parent, par_args is empty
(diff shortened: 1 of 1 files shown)


zdykstra (commented):
The double-space in the KCL is visually obnoxious, but harmless. Everything else here looks good.

ahesford:
Hrm, we have a problem here.
```
zfs set org.zfsbootmenu:commandline='ztest_arg' ztest
zfs set org.zfsbootmenu:commandline='%{parent} ztest_root_arg ztest_root_arg2' ztest/ROOT
```
will cause the property to be inherited as `%{parent} ztest_root_arg ztest_root_arg2` on `ztest/ROOT/bootenv`. Then when we do the substitution, we get `%{parent} ztest_root_arg ztest_root_arg2` again for the `%{parent}` value associated with `ztest/ROOT`, and then we add the value for `ztest` to get
```
ztest ztest_root_arg ztest_root_arg2 ztest_root_arg ztest_root_arg2
```
Repeated values are not guaranteed to be safe. (Add two `ip=` commands for `dracut-crypt-ssh` and watch your image fail to boot.)

I think the right fix here is to read the source of the `org.zfsbootmenu:commandline` property in addition to the value. If the value does not match `%{parent}`, nothing special must be done. Otherwise, if the property is inherited (from any parent), we can do one of two things:
1. Ignore the source of the inheritance, and just return the (interpolated) string for `read_kcl_prop` from the immediate parent. Recursively, this will work itself out, but it will touch every parent until it hits a filesystem where the property isn't inherited.
2. Scrape the inherited `source` value for the property in the desired filesystem, and jump right to that property. For example, when I do
    ```
    zfs set org.zfsbootmenu:dummyprop=myprop system
    zfs get -H -o source,value org.zfsbootmenu:dummyprop system/home/ajh/music
    ```
    the printed result is
    ```
    inherited from system	myprop
    ```
    We know we can jump right up to `system` and skip the query for all the intermediate ancestors, but we increase the parsing complexity.

It's not clear what option is best, but I guess I favor Option 1.

ahesford:
I implemented Option 1 and believe it now behaves as expected with inheritance.

PR #176, opened 2021-06-03
Countdown to extinction

ahesford:
I got tired of `zfsbootmenu-exec.sh` and `zfsbootmenu-countdown.sh` having non-descriptive and inaccurate names, so I renamed them. I also added support for import retries.

This is totally untested, but seems sound.


ahesford:
To test, I forced imports to fail for a set number of times:
```diff
diff --git a/90zfsbootmenu/zfsbootmenu-init.sh b/90zfsbootmenu/zfsbootmenu-init.sh
index 7ede990..3fd2e46 100755
--- a/90zfsbootmenu/zfsbootmenu-init.sh
+++ b/90zfsbootmenu/zfsbootmenu-init.sh
@@ -53,6 +53,8 @@ else
   first_pass=1
 fi
 
+: > "${BASE}/IMPORT_FAIL"
+
 zbm_import_attempt=0
 while true; do
   if [ "${first_pass}" -eq 0 ]; then
@@ -95,6 +97,7 @@ while true; do
   fi
 
   if [ "${zbm_import_attempt}" -lt "${zbm_import_retries:-0}" ]; then
+    [ "${zbm_import_attempt}" -gt 3 ] && rm -f "${BASE}/IMPORT_FAIL"
     # The max number of import attempts has not been reached;
     # retry unless the user decides to abort immediately
     zbm_import_attempt="$((zbm_import_attempt + 1))"
diff --git a/90zfsbootmenu/zfsbootmenu-lib.sh b/90zfsbootmenu/zfsbootmenu-lib.sh
index 759e476..4ba9f79 100755
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -1387,6 +1387,11 @@ import_pool() {
     pool="-a"
   fi
 
+  if [ -e "${BASE}/IMPORT_FAIL" ]; then
+    zerror "deliberately failed to import pool"
+    return 1
+  fi
+
   # Import /never/ mounts filesystems
   import_args=( "-N" )
 
```

Next, I started a test environment:
```bash
./run.sh -D /tmp/void -A zbm.import_retries=5 -A zbm.import_delay=2 -f
```

Behavior is as expected. The import failure countdown ticks away two seconds, repeates four times, and then succeeds on the import. Pressing `[RETURN]` short-circuits each countdown and re-fires the import retry. Pressing `[ESC]` aborts the countdown and drops right into an emergency shell. Changing `zbm.import_delay=10` changes the countdown timer to 10 seconds, omitting it or using invalid values like `mechanix`, `-1` or `0` changes the countdown the default 5-second value.

Setting `zbm.import_retries=3` will cause three countdowns and then a hard failure because the `IMPORT_FAIL` file is never removed. Using invalid values like `trust`, `-1` or `0` will disable retries altogether.

90zfsbootmenu/zfsbootmenu-init.sh
```diff
@@ -93,6 +94,16 @@ while true; do
     continue
   fi
 
+  if [ "${zbm_import_attempt}" -lt "${zbm_import_retries:-0}" ]; then
+    # The max number of import attempts has not been reached;
+    # retry unless the user decides to abort immediately
+    zbm_import_attempt="$((zbm_import_attempt + 1))"
+    if delay="${zbm_import_delay:-5}" prompt="Unable to import pool, retrying in %0.2d seconds" \
+        timed_prompt "[RETURN] to retry immediately" "[ESCAPE] for a recovery shell"; then
+      continue
```
zdykstra, line 103: Can you throw in a zdebug or zinfo statement here indicating that it's on import attempt X of Y?

zdykstra (changes requested):
The rename to preinit and init is good.

The integer value detection of the two new parameters works as expected. I don't have a reasonable way to test the retry loop, but the logic seems correct.

ahesford:
I did a test with hostid matching:
```diff
diff --git a/90zfsbootmenu/zfsbootmenu-init.sh b/90zfsbootmenu/zfsbootmenu-init.sh
index ca4334c..3c7edb4 100755
--- a/90zfsbootmenu/zfsbootmenu-init.sh
+++ b/90zfsbootmenu/zfsbootmenu-init.sh
@@ -53,6 +53,8 @@ else
   first_pass=1
 fi
 
+: > "${BASE}/MATCH_FAIL"
+
 zbm_import_attempt=0
 while true; do
   if [ "${first_pass}" -eq 0 ]; then
@@ -99,6 +101,8 @@ while true; do
     # retry unless the user decides to abort immediately
     zbm_import_attempt="$((zbm_import_attempt + 1))"
 
+    [ "${zbm_import_attempt}" -gt 1 ] && rm -f "${BASE}/MATCH_FAIL"
+
     zinfo "pool import failed on attempt ${zbm_import_attempt} of ${zbm_import_retries}"
 
     if delay="${zbm_import_delay:-5}" prompt="Unable to import pool, retrying in %0.2d seconds" \
diff --git a/90zfsbootmenu/zfsbootmenu-lib.sh b/90zfsbootmenu/zfsbootmenu-lib.sh
index 759e476..9aeff51 100755
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -228,6 +228,11 @@ match_hostid() {
     zdebug "trying to import: ${pool}"
     hostid="$( zpool import -o readonly=on -N "${pool}" 2>&1 | grep -E -o "hostid=[A-Za-z0-9]{1,8}")"
 
+    if [ -e "${BASE}/MATCH_FAIL" ]; then
+      hostid="hostid=1badbabe"
+      zerror "Deliberate hostid override: ${hostid}"
+    fi
+
     if [ -n "${hostid}" ]; then
       hostid="$( printf "%08x" "0x${hostid##*=}" )"
       zdebug "discovered pool owner hostid: ${hostid}"

```
and booted the test image with
```bash
run.sh -D /tmp/void -A zbm.import_retries=5 -A spl_hostid=deadbeef -A zbm.prefer=bpool -A zbm.import_policy=hostid
```

The test VM had two pools, "apool" and "bpool", but this was only half a test because both had the same encoded hostid. Nevertheless, imports failed the first three times, and then the loop tried (and succeeded) to import `apool` rather than the preferred `bpool`. This is expected, since the preference is disregarded after the first failed import attempt.

I believe things are working properly and also added your zinfo line.

ahesford:
As an aside, notice the last commit to ignore `linux` and `linux-headers` in Void, since `dracut` was pulling in `linux-headers` and our later installation of `linux5.10{,-headers}` was causing the test setup to build ZFS for two kernels (one of which wasn't even installed). I think it better if we explicitly request a versioned kernel package for Void test environments.

zdykstra (approved):

PR #177, opened 2021-06-04
Small quality of life improvements

zdykstra:
When console=ttyS0,1152800n8 is passed in to the kernel it does not directly translate to a /dev entry. Simply strip the ,value off of the console=x,y parameter and set that as our /dev entry.

When a pool import fails for some reason, log the text from the zpool import process at the debug log level.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -1372,7 +1372,7 @@ load_be_cmdline() {
 # rewind_to_checkpoint=1: enable --rewind-to-checkpoint
 
 import_pool() {
-  local pool import_args
+  local pool import_args import_output
 
   pool="${1}"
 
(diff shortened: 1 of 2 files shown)


zdykstra:
This'll need to be rebased once #176 is merged.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -1413,14 +1413,18 @@ import_pool() {
 
   zdebug "zpool import arguments: ${import_args[*]} ${pool}"
 
-  zpool import "${import_args[@]}" "${pool}" >/dev/null 2>&1
+  import_output="$(
+    zpool import "${import_args[@]}" "${pool}" 2>&1
+    exit $?
```
ahesford, line 1418: Doesn't the return value of `zpool import` fall through to be the return value of the `import_output=` assignment without needing `exit $?`?

zdykstra: You're right, the last return code in a subshell is the exit code of the subshell.

ahesford (approved):

PR #180, opened 2021-06-20
Add keyboard shortcut to remove pinned kernel

zdykstra:
Add the `MOD+U` keyboard shortcut to allow 'unsetting' a pinned kernel, by way of inheriting the `org.zfsbootmenu:kernel` property. Under normal conditions (kernels pinned by ZBM), it'll simply remove the pin. The code to handle this was already present in the `set_default_kernel` function, it was just never glued to the user interface.


ahesford (approved):
This is a good feature. I'd rather nail the user interface to it, but we can use glue if we must.

PR #181, opened 2021-06-20
Change ZBM defaults

zdykstra:
After a few months of testing, the `zbm.import_policy=hostid` and `zbm.set_hostid=on` have been tested enough to be the new default behaviors. Enable them and then update documentation to note that they're on by default.


ahesford (approved):
I'm glad to see hostid taking a diminished role.

PR #183, opened 2021-06-21
Improved pool imports

ahesford:
1. Allow `zbm.prefer=<pool>!` syntax for a *mandatory* preference; a mandatory poll must be imported before other pools will be attempted and ZBM continues.

2. Eliminate limits on import retries; the loop will continue indefinitely, displaying a countdown that the user can interrupt to drop to an emergency shell.

I need some help thinking through the logic to make sure the overhauled import loop will terminate properly, and this needs some real testing. My assumptions are:
- [ ] If we have a boot pool preference, we try that first; otherwise, the preference is empty to try any import.
- [ ] If an import attempt succeeds, then either:
    - A preference was indicated, so the preferred pool *must* have succeeded; in this case, try again with no preference to pick up any other importables. (This should happen exactly once.)
    - A preference was not indicated (because it never existed or we are retrying after dropping it); we are done because we have found at least one pool.
- [ ] If the import attempt fails but host matching is desired, attempt to match the hostid.
    - When hostid matching succeeds, it means *exactly one* pool has been imported with an ID match. If this succeeds when we requested a specific preference, that means we have imported our preference and can try again (exactly once) with no preference to pick up other pools if possible. If this succeeds when we requested no specific preference, we still wish to try again (with no preference) to make sure to grab other pools with a (possibly new) hostid.
    - When hostid matching fails, we should have never have imported pools.
- [ ] If we have asked for a specific pool but failed to import it, and the pool is not hard required, just drop the preference and make another import attempt.
- [ ] Otherwise, we exist in an error state---either we have asked for any pool but imported nothing, or we ask for a specific pool and it is not available. Show the countdown before trying again to throttle retries and allow the user to interrupt.

Please confirm that my assumptions are reasonable and that the reworked loop conforms to my assumption. I believe there are no cases where the loop can continue indefinitely *without* a countdown timer between iterations---any continue without a timer should jump to the final iteration in all cases---but would like additional validation.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-init.sh
+++ b/90zfsbootmenu/zfsbootmenu-init.sh
@@ -91,7 +91,7 @@ while true; do
   zinfo "unable to import a pool on attempt ${zbm_import_attempt}"
 
   # Just keep retrying after a delay until the user presses ESC
-  if delay="${zbm_import_delay:-5}" prompt="Unable to import pool, retrying in %0.2d seconds" \
+  if delay="${zbm_import_delay:-5}" prompt="Unable to import ${try_pool:-pool}, retrying in %0.2d seconds" \
     timed_prompt "[RETURN] to retry immediately" "[ESCAPE] for a recovery shell"; then
       continue
   fi
--- a/90zfsbootmenu/zfsbootmenu-init.sh
+++ b/90zfsbootmenu/zfsbootmenu-init.sh
@@ -46,31 +46,21 @@ else
   boot_pool="${root}"
 fi
 
-# Do a dedicated pass for the preferred pool if one was provided
-if [ -n "${boot_pool}" ]; then
-  first_pass=0
-else
-  first_pass=1
-fi
-
+# If a boot pool is specified, that will be tried first
+try_pool="${boot_pool}"
 zbm_import_attempt=0
 while true; do
-  if [ "${first_pass}" -eq 0 ]; then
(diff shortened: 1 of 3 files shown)


ahesford:
I did some basic tests, installing a Void BE on an encrypted `zpool` pool and an Arch BE on a cleartext `anthillsSiberia` pool. Booting with `zbm.prefer=badpool!` causes an (apparently) unbounded loop where I see the countdown for "Unable to import badpool, retrying in %02d seconds". I can press `RETURN` to immediately retry the import and reset the counter; I can press `ESC` to drop to a recovery shell. `zlogtail` shows repeated attempts to import `badpool`. If I manually import one pool (say `anthillsSiberia`) and then exit the recovery shell, it tries and fails once again to import `badpool`. However, because it then finds one imported pool, it moves to the final "all-pools" import stage and attempts to import the encrypted `zpool`. This seems to be consistent with expected behavior.

If I instead boot with `zbm.prefer=badpool`, ZBM tries and fails to import `badpool`. However, because the pool is not required, it then proceeds to the "all-pool" import and proceeds as it would have before.

zdykstra:
I've read and re-read the modified import loop, and your description of it lines up with how I'm reading the code. The _only_ way out of the loop is if at least one pool has been imported. The exit condition is open-ended enough that it doesn't look like you can get 'stuck' in the loop if at some stage a pool is present.

zdykstra (approved):

PR #184, opened 2021-06-23
Validate module parameters, verify that kernel modules are loaded

zdykstra:
It was relatively easy to break ZFSBootMenu to the point that it was almost unrecoverable by passing an invalid `spl.spl_hostid` parameter. When an invalid parameter is set on the kernel command line it can't be overridden except by directly calling `insmod` with the full path to the kernel module and the fixed argument to the parameter.

Additionally, if `spl.spl_hostid` was passed into ZBM, incorrect default-hostid logic was applied because `spl_hostid` wasn't detected on the kernel command line, nor is `/etc/hostid` part of the release builds. Both forms of passing the hostid are now accepted and the argument is normalized when possible to an 8 character hex value. If it can't be normalized, `spl_hostid` is unset and the global default of `00bab10c` is used.

In `zfsbootmenu-init.sh`,  find the on-disk path for `spl.ko` and manually load it with `insmod` and `spl_hostid=0`. This accomplishes two things - it bypasses bad `spl.spl_hostid` values set on the kernel command line, and it immediately switches the module to use `/etc/hostid` as the `hostid` source, which we already do in `write_hostid`.

Loading the `zfs` kernel module has also been moved to `zfsbootmenu-init.sh` so that all of the helper functions are available. If `modprobe zfs` fails, the error output is logged. A recovery shell is invoked so that users can look at logs to determine what happened.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-parse-commandline.sh
+++ b/90zfsbootmenu/zfsbootmenu-parse-commandline.sh
@@ -21,10 +21,33 @@ fi
 
 # Let the command line override our host id.
 # shellcheck disable=SC2034
-spl_hostid=$(getarg spl_hostid=)
+cli_spl_hostid=$( getarg spl.spl_hostid ) || cli_spl_hostid=$( getarg spl_hostid )
+if [ -n "${cli_spl_hostid}" ] ; then
+  # Start empty so only valid hostids are set for future use
+  spl_hostid=
+
+  # Test for decimal
+  if (( 10#${cli_spl_hostid} )) >/dev/null 2>&1 ; then
+    spl_hostid="$( printf "%08x" "${cli_spl_hostid}" )"
+  # Test for hex. Requires 0x, if present, to be stripped
+  # The change to cli_spl_hostid isn't saved outside of the test
+  elif (( 16#${cli_spl_hostid#0x} )) >/dev/null 2>&1 ; then
+    spl_hostid="${cli_spl_hostid#0x}"
+    # printf will strip leading 0s if there are more than 8 hex digits
+    # normalize to a maximum of 8, then run through printf to fill in
+    # if there are fewer than 8 digits.
+    spl_hostid="$( printf "%08x" "0x${spl_hostid:0:8}" )"
+  # base 10 / base 16 tests fail on 0
+  elif [ "${cli_spl_hostid#0x}" -eq "0" ] >/dev/null 2>&1 ; then
+    spl_hostid=0
+  # Not valid hex or dec, log
+  else
+    warn "ZFSBootMenu: invalid hostid value ${cli_spl_hostid}, ignoring"
+  fi
+fi
 
 # Use the last defined console= to control menu output
-control_term=$( getarg console=)
+control_term=$( getarg console )
 if [ -n "${control_term}" ]; then
   control_term="/dev/${control_term%,*}"
   info "ZFSBootMenu: setting controlling terminal to: ${control_term}"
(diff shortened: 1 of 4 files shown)


90zfsbootmenu/zfsbootmenu-parse-commandline.sh
```diff
@@ -21,10 +21,31 @@ fi
...
+
+  # Start empty so only valid hostids are set for future use
+  spl_hostid=
+
+  # Test for decimal
+  if (( 10#${cli_spl_hostid} )) >/dev/null 2>&1 ; then
+    spl_hostid="$( printf "%08x" "${cli_spl_hostid}" )"
+  # Test for hex. Requires 0x, if present, to be stripped
+  # The change to cli_spl_hostid isn't saved outside of the test
+  elif (( 16#${cli_spl_hostid//0x/} )) >/dev/null 2>&1 ; then
```
ahesford, line 37: 1. Probably better just to do `${cli_spl_hostid#0x}` since accepting something like `0x0xdead0xbeef` as `deadbeef` seems dubious.
2. `(( 10#0 ))` and `(( 16#0 ))` both silently return `1` and cause the tests to fail. What's the right response here? Setting `spl_hostid=0` might indicate a desire to force all zeros in `/etc/hostid`.

90zfsbootmenu/zfsbootmenu-parse-commandline.sh
```diff
@@ -21,10 +21,31 @@ fi
...
+    spl_hostid="${cli_spl_hostid//0x/}"
```
ahesford, line 38: Should this go through `printf` too to normalize things like `1234` to `00001234`?

90zfsbootmenu/zfsbootmenu-parse-commandline.sh
zdykstra: Good catch, I'll special-case 0 / 0x0. I can change the `//0x/` to just remove a single prefix. It's fairly unlikely that we'd ever need to protect against a random smattering of `0x` spread throughout the hostid string.

90zfsbootmenu/zfsbootmenu-parse-commandline.sh
zdykstra: I'd get normalized in `write_hostid`, but it should be done early as well so that it's visually consistent throughout logging.

zdykstra:
Instead of trying to modprobe `spl` and then extract a path to the file on failure, `modinfo` is used to find the on-disk filename for `spl` and then load it with a forced `spl_hostid=0` module parameter.

90zfsbootmenu/zfsbootmenu-parse-commandline.sh
```diff
@@ -21,10 +21,34 @@ fi
...
+  # The change to cli_spl_hostid isn't saved outside of the test
+  elif (( 16#${cli_spl_hostid#0x} )) >/dev/null 2>&1 ; then
+    spl_hostid="${cli_spl_hostid#0x}"
+    # printf will strip leading 0s if there are more than 8 hex digits
+    # normalize to a maximum of 8, then run through printf to fill in
+    # if there are fewer than 8 digits.
+    spl_hostid="${spl_hostid:0:8}"
+    spl_hostid="$( printf "%08x" "${cli_spl_hostid}" )"
+  # base 10 / base 16 tests fail on 0
+  elif [ "${cli_spl_hostid#0x}" = "0" ] ; then
```
ahesford, line 42: If you use `-eq`, this should handle cases with any number of zeros gracefully, but you probably need to suppress stderr in case cli_spl_hostid is some non-numeric string.

PR #185, opened 2021-06-24
Mention that Arch+Dractut needs org.zfsbootmenu:rootprefix property set

### LINKED ISSUE #179: Arch fails to boot (when using Dracut?)
I recently upgraded from ~0.7 of zbm to 1.9, and my Arch  OS no longer booted, instead complaining about a timout on /dev/gpt-auto-root.mount.

The fix for me was to set `org.zfsbootmenu:rootprefix` to `root=zfs:`, which is the behaviour I had previously.

So maybe the setting depends not just on the OS, but on what kind of initramfs is in use?

ashb:
Closes #179


ahesford:
Arch was specifically mentioned in the manual page because, at the time the test was written, it was the only special case. Since then, we've added a special case for Gentoo as well. I have some pending edits to the manual page to note the deviation in each special case and some clarifying text that the defaults are chosen based on the typical initramfs generator for each distribution. Beyond that, I don't think our documentation should call special attention to nonstandard setups for Arch.

PR #186, opened 2021-06-30
Release notes and fixes for 1.10

zdykstra:
* `insmod` doesn't gracefully fail if a module is already loaded. An initramfs generated on Arch loads `spl.ko` before our insmod attempts it, so check if it's not loaded and then load it.
* Split the filesystem/kernel/initramfs before the subkey handler under the kernel selection handler, so that it's available for pinning and unpinning kernels.
* Remove documentation for `zbm.import_retries` and regenerate the online help pages for that section.
* First pass at release notes for 1.10.


ahesford (approved):
I've cleaned up the documentation and changelog a bit; if you agree with my changes, squash everything down and we're ready to go.

PR #192, opened 2021-07-15
Add easy 'takeover' helper

zdykstra:
To facilitate easily taking over a running ZFSBootMenu instance, regardless of where it is (ZFS password prompt, at the main menu, etc), the 'takeover' function has been created. 

In broad strokes, this works as follows:
##### `/bin/zfsbootmenu`
* `/bin/zfsbootmenu` writes it's PID to `/zfsbootmenu/active`
* `/bin/zfsbootmenu` Installs a `USR1` handler that runs `exit 0`

##### `takeover()`
* reads the pid out of `/zfsbootmenu/active` if it exists, sends it a `USR1` signal (which won't be processed until child processes are killed)
* Walks down the list of child PIDs for the active `/bin/zfsbootmenu` process; stops at the last child
* Kills the last child (typically either `fzf` or `zfs` password prompt)
* The `SIGUSR1` trap now executes, causing the running `zfsbootmenu` instance to exit
* On exit, `/zfsbootmenu/active` is removed
* A new `/bin/zfsbootmenu` is exec'd out of the `takeover()` function

A couple small changes were made to help out with this:

`BASE` is now added to `/etc/profile` and exported from it so all subsequent shells inherit it. The duplicate defines in a few different script entrypoints were removed.

`emergency_shell()` now runs `stty echo` to re-enable input echoing if something like a ZFS password prompt has disabled it.

zlog has a fixed `basename` usage for when it has a `$0` of `-sh`. 

I've tested this by booting a VM and leaving it at the password prompt for the zpool, then SSH'ing in and running `takeover`. This results in the instance attached to ttyS0 dropping to an emergency shell and the password prompt migrating to the SSH connection. Additionally, I've tested entering a password on ttyS0 such that it gets to the main menu and then taking it over via SSH. This also causes ttyS0 to drop to a shell.

In both cases, running `takeover` on the serial console results in the SSH session dropping to a recovery shell again.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -2121,6 +2121,21 @@ zbmcmdline() {
 }
 
 # prints: nothing
+# returns: nothing
+
+takeover() {
+  local pid
+
+  # Kill the other running instance
+  if [ -f "${BASE}/active" ] ; then
+    read -r pid < "${BASE}/active"
+    [ -n "${pid}" ] && kill "${pid}"
+  fi
+
+  exec /bin/zfsbootmenu
+}
+
+# prints: nothing
 # returns: 0
 
 is_lib_sourced() {
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -18,7 +18,7 @@ zlog() {
   [ "${1}" -le "${loglevel:=4}" ] || return
 
   # shellcheck disable=SC2086
-  _script="$( basename $0 )"
+  _script="$( basename -- $0 )"
   _func="${FUNCNAME[2]}"
 
   # Only add script/function tracing to debug messages
@@ -2120,18 +2120,61 @@ zbmcmdline() {
   [ -f "${BASE}/zbm.cmdline" ] && echo | cat "${BASE}/zbm.cmdline" -
 }
 
+# arg1: pid
+# prints: child pid
+# returns: 0 if a child pid was found, 1 if there are no children
+
+find_child_pid() {
+  local pid child
+
+  pid="${1}"
+  if [ -z "${pid}" ]; then
+    zdebug "empty pid"
+    return 1
+  fi
+
+  if [ -e "/proc/${pid}/task/${pid}/children" ] ; then
+    read -r child < "/proc/${pid}/task/${pid}/children"
+
+    if [ -n "${child}" ] ; then
+      echo "${child}"
+      return 0
+    fi
+  fi
+
+  return 1
+}
+
 # prints: nothing
 # returns: nothing
 
 takeover() {
-  local pid
-  echo "Starting takeover"
+  local pid child
 
   # Kill the other running instance
   if [ -e "${BASE}/active" ] ; then
     read -r pid < "${BASE}/active"
-    [ -n "${pid}" ] && kill "${pid}"
-    zinfo "Killing active zfsbootmenu with a PID of ${pid}"
+    parent=${pid}
+
+    echo "Stopping active zfsbootmenu with a PID of ${pid}"
+
+    # Trip the USR1 handler in /bin/zfsbootmenu - 'exit 0'
+    kill -USR1 "${parent}"
+
+    # find the last child process of the active /bin/zfsbootmenu
+    while child="$( find_child_pid "${parent}" )" ; do
+      zdebug "Child pid of ${parent} is ${child}"
+      if [ -n "${child}" ] ; then
+        parent=${child}
+        continue
+      fi
+      break
+    done
+
+    zdebug "Final child pid is ${parent}"
+
+    # Kill the blocking child so that the USR1 handler actually executes
+    [ -n "${parent}" ] && kill "${parent}"
   fi
 
   /bin/zfsbootmenu
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -2160,12 +2160,11 @@ takeover() {
     echo "Stopping active zfsbootmenu with a PID of ${pid}"
 
     # Trip the USR1 handler in /bin/zfsbootmenu - 'exit 0'
-    zdebug "Sending USR1 to ${parent}"
+    zdebug "sending USR1 to ${parent}"
     kill -USR1 "${parent}"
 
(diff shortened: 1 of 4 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -18,7 +18,7 @@ zlog() {
   [ "${1}" -le "${loglevel:=4}" ] || return
 
   # shellcheck disable=SC2086
-  _script="$( basename $0 )"
+  _script="$( basename -- $0 )"
```
ahesford, line 21: Is this going to break when we start using FreeBSD userland?

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -2120,6 +2121,65 @@ zbmcmdline() {
...
+      if [ -n "${child}" ] ; then
+        parent=${child}
+        continue
+      fi
+      break
+    done
+
+    # Kill the blocking child so that the USR1 handler actually executes
+    zdebug "killing child process ${parent}"
+    [ -n "${parent}" ] && kill "${parent}"
```
ahesford, line 2177: I'm out of my depth here, but it would be great if there were built-in job control primitives that would do this without requiring a tree walk.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -2120,6 +2121,65 @@ zbmcmdline() {
...
+find_child_pid() {
+  local pid child
+
+  pid="${1}"
+  if [ -z "${pid}" ]; then
+    zdebug "empty pid"
+    return 1
+  fi
+
+  if [ -e "/proc/${pid}/task/${pid}/children" ] ; then
```
ahesford, line 2137: Any idea how widely supported this is?

90zfsbootmenu/zfsbootmenu-preinit.sh
```diff
@@ -24,6 +24,7 @@ mkdir -p "${BASE}"
 # shellcheck disable=SC2154
 cat >> "/etc/profile" <<EOF
 # Added by zfsbootmenu-preinit.sh
+export BASE="/zfsbootmenu"
```
ahesford, line 27: What do you think about changing this to `ZBMROOT` or something less generic than `BASE`?

ahesford (commented):
This is a great idea. Frankly, if it's rock-solid, I'd like to effectively run `takeover` *within* `/bin/zfsbootmenu` before entering the spinlock when another instance is running. This way, logging in and just running `zfsbootmenu` would do the right thing unless the kill-and-cleanup logic fails.

90zfsbootmenu/zfsbootmenu-preinit.sh
zdykstra: `ZBMROOT` is much more meaningful than `BASE` - I'd be glad to see that globally changed if/after this PR is merged.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Yes, horribly. Patches welcome.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: I looked and nothing jumped out at me. I'll dig around a bit more though, this loop and the logic in `find_child_pid` feels sketchy.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: I'll have to see when this was added to the `/proc` vfs - I image it's been there for quite a few major kernel versions though.

zdykstra: Looks like it's been there since at least 2012 - https://lwn.net/Articles/475688/. I think we can safely assume it's always available.

ahesford: Are you really prepared to abandon support for anything before the latest 2.6 kernels? That probably makes us incompatible with Slackware.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: The only other way that I can see to do this is to use `ps --sid <session id> -o pid -h` to find all of the processes in a fixed session ID, but there's a lot of faffing about parsing the output of `ps` to discover the session ID of the active zfsbootmenu process, then find all the pids in that session, etc. A /proc tree walk is probably the best interface here for the problem.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Don't forget CentOS 5.

90zfsbootmenu/zfsbootmenu-lib.sh
ahesford: Yeah; if we have to signal children manually, walking `/proc` is as clean as it gets.

We need bash to implement `set --interruptible-children` or something to just do the right thing without our involvement.

zdykstra:
> This is a great idea. Frankly, if it's rock-solid, I'd like to effectively run `takeover` _within_ `/bin/zfsbootmenu` before entering the spinlock when another instance is running. This way, logging in and just running `zfsbootmenu` would do the right thing unless the kill-and-cleanup logic fails.

I've implemented this change by no longer having `takeover()` exec `/bin/zfsbootmenu`, but instead have `zfsbootmenu` invoke `takeover()` if `/zfsbootmenu/active` exists. Once takeover returns, the existing spinlock is still there to wait for `/zfsbootmenu/active` to disappear. In general it shows the screen for one to two seconds and then goes straight to the main menu. 

The only remaining issue here that should maybe be addressed is slightly convoluted, but it could happen. If you set `zbm.timeout=120` (or some other high value), then SSH in and run `zfsbootmenu`, you'll immediately take control of the system because `/zfsbootmenu/active` hasn't been created. However, the countdown timer is still running. If you unlock the system in the interim and then start doing something, the timer will still be counting down and can/will auto-boot the system out from under you.

If this is even considered a problem, some of the countdown / boot logic in -init should be moved to a standalone script that can be killed by the takeover function. Then, of course, it'll immediately fall-through to launching /bin/zfsbootmenu, where it'll kill the copy running under SSH ...

PR #194, opened 2021-07-18
releng/make-binary.sh: use containerized image builds

ahesford:
Using podman to containerize the production of release assets avoids potential leakage of personal information. To support this, the containerized build script has been modified to separate the repo path at /zbm from a "build" directory (previously /zbm/contrib/docker was hard coded, now it's just the default choice). The releng script creates a temporary directory to serve as the "build" path, populates configs, and runs the build container with the current repo at /zbm and the temporary directory at /build. The outputs are copied as before, with the EFI executable standing alone and the kernel/initramfs components stored in a gzipped tarball.

If the releng script does not find the expected builder image (zbm-builder by default, but this can be passed as a second argument to the script), it will invoke `podman build` to create the image.

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -8,13 +8,21 @@
 
 # Use the official Void Linux container
 FROM voidlinux/voidlinux:latest
+ARG ZBM_COMMIT_HASH
 
 # Ensure everything is up-to-date
 RUN xbps-install -Suy xbps && xbps-install -uy
 
+# Prefer an LTS version over whatever Void thinks is current
+RUN echo "ignorepkg=linux" > /etc/xbps.d/10-nolinux.conf \
+	&& echo "ignorepkg=linux-headers" >> /etc/xbps.d/10-nolinux.conf
+
 # Install components necessary to build the image
 RUN xbps-query -Rp run_depends zfsbootmenu | xargs xbps-install -y
-RUN xbps-install -y linux linux-headers zfs gummiboot-efistub
+RUN xbps-install -y linux5.10 linux5.10-headers zfs gummiboot-efistub
+
+# Record a commit hash if one was provided
+RUN if [ -n "${ZBM_COMMIT_HASH}" ]; then echo "${ZBM_COMMIT_HASH}" > /etc/zbm-commit-hash; fi
 
 # Copy the build script
 COPY zbm-build.sh /zbm-build.sh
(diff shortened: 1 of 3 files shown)


zdykstra (changes requested):
I generally like everything here, and a local test worked as expected!

One of the concerns I have with a containerized build environment is that the current setup will allow the support software we use (ZFS, Linux, fzf, and so on) to potentially be out of date without any real feedback.  

I'd like to see a default-enabled option to `releng/make-binary.sh` to destroy `zbm-builder:latest` and recreate it. That will add a bit of time to each binary build, but it'll also ensure that everything is current.

It's potentially out of the scope of this PR, but it'd be nice to have the binary build script also create a versions.txt file or some such that records the specific versions of key pieces of software used in each image. At first blush, that could be:

* Linux kernel
* ZFS
* fzf (if for no other reason than to give a shout out to the fzf dev for his great tool)
* Dracut (?)


The rest of the `essential_execs` are pretty generic and stable at this point.

ahesford:
Okay, some new changes:
1. `releng/make-binary.sh` now uses the current date for a tag instead of the default `:latest`, so it will generally ensure that the build environment is up to date.
2. `contrib/docker/Dockerfile` now ignores `linux` and `linux-headers` in favor of `linux5.10` and `linux5.10-headers` so we build images from an LTS kernel.
3. Because we're now tagging builder images by date, we might want to publish them at some point, and we want the asset production to be (somewhat) reproducible. The `Dockerfile` now supports `--build-arg=ZBM_COMMIT_HASH=<some-value>` and, if this argument is set, the image will contain a file `/etc/zbm-commit-hash` that holds the text `<some-value>`. When a container is run from this image, `zbm-build.sh` will look to this file to determine a commit to checkout iff it clones the ZFSBootMenu repo (*i.e.*, iff a new container is instantiated and `/zbm` is not bind-mounted from the host). The script will also respect a `ZBM_COMMIT_HASH` environment variable set when the container is run, which supersedes anything in `/etc/zbm-commit-hash`. If the file does not exist and no environment variable has been set, the default `master` checkout will prevail.
4. `make-binary.sh` now sets `ZBM_COMMIT_HASH` to the hash of `HEAD` and also modifies the ZBM dracut configuration to install `/etc/zbm-commit-hash` in the the EFI and component outputs, if it exists. This isn't foolproof; for example, if you run `make-binary.sh` to create a builder image and some assets, then add some commits and re-run `make-binary.sh` on the same day, it won't recreate the builder and `/etc/zbm-commit-hash` will still point to the old `HEAD`.

zdykstra (approved):

Issue #195, opened 2021-07-21
Improving containerized builds


ahesford:
Now that we have a container definition for building ZBM images, it's time to think about how the image can be made adaptable to diverse workflows. It would be nice to support them generically so people can just run a pre-built image to spit out local images without installing a bunch of dependencies. The three major uses of the container appear to be:

1. Building a custom image from a local repository
2. Building a release image, possibly from a specific commit, without requiring a local repo
3. Building a custom image, possibly from a specific commit, without requiring a local repo.

The local-repo builds are handled now by bind-mounting whatever repo to `/zbm` in the container. Automatic clone and optional checkout of a specific commit are handled now whenever `/zbm` is found to be empty in the container. The release image is handled by extensive configuration in `releng/make-binary.sh`, but there is no good solution now for custom builds without requiring a local repo.

## Differentiation between ZFSBootMenu sources, configuration data and output locations is too coarse.
 
The build script differentiates `/zbm`, which must be like a ZFSBootMenu repository that provides `bin/generate-zbm` and `90zfsbootmenu` and a `$BUILDROOT` that holds the ZBM configuration file and optional `hostid` and `zpool.cache` files. By default, `$BUILDROOT=/zbm/contrib/docker`. If `config.yaml` is not found at `${BUILDROOT}`, it is copied from `${BUILDROOT}/config.yaml.default`. The default `config.yaml.default` uses `/zbm/etc/zfsbootmenu/dracut.conf.d` as the configuration directory for dracut (*i.e.*, whatever is in the `/zbm` repository) and dumps output to `/zbm/contrib/docker/build` (*i.e.*, in a `build` subdirectory of the default `$BUILDROOT`).

I think that assuming `/zbm` always contains a repo should be fine. Using the repo as-is if it seems non-empty seems generally OK, and I think doing a `git clone` if `/zbm` appears empty is sensible too. The container image provides an optional `/etc/zbm-commit-hash` or an overriding `$ZBM_COMMIT_HASH` environment variable that will checkout a specific named commit iff a `git clone` is performed; if the container already appears to contain a non-empty `/zbm`, these are ignored. Generally, this is sensible; we don't want to try changing the state of an existing repo when it's bind-mounted. What about if an old container is re-used? We probably don't care, because these containers are generally meant to be created for a single image and then thrown away. However, if the container should remain and be restarted, I'm open to debate on how we should handle `$ZBM_COMMIT_HASH` stuff.

[...]

PR #196, opened 2021-08-03
Overhaul, generalize containerized build scripts

ahesford:
Accept command-line arguments and environment variables to specify:
- BUILDROOT: the default source of config, hostid and zpool.cache files
- ZBMCONF: a specific configuration file to use inside the container
- ZBMOUTPUT: a directory where build artifacts will be copied
- HOSTID: a specific hostid file to be copied to /etc/hostid
- POOLCACHE: a specific cache to be copied to /etc/zfs/zpool.cache
- ZBMTAG: a tag to fetch if /zbm is not pre-populated in container

The zbm-build.sh script now overrides ImageDir values and removes Global.BootMountPoint from any configuration, writing artifacts to a temporary directory and copying them to the output directory after a successful run.

When /zbm is not pre-populated, it is now built in-container from a tarball fetched from github.com rather than a git clone. This reduces instantiation time and lightens the dependency burden.

Closes: #195.

I've tested lightly but it needs a bit more testing. Also, I want to migrate the `make-binary.sh` script to rely more heavily on container configuration.


zdykstra (approved):

zdykstra:
The only change that will (eventually) be needed for GitHub support is the removal of `VOLUME /zbm` in releng/docker/Dockerfile, but that can go in when that work is committed.

ahesford:
We can easily remove the volume since I `mkdir -p /zbm` anyway. Just swap `/zbm` for `ZBMREPO` (`/zbm` by default) and add a `-z` option to specify the root where we want to work and we should be all set. Merging as-is for now.

PR #197, opened 2021-08-19
[WIP] Recovery shell improvements

zdykstra:
Add a few more niceties to the recovery shell:

- `help` now shows the help pager with recovery shell documentation
- A few useful `-lib` functions now have tab completion so that you can handily interact with ZFS filesystems
- Added a `set_ro_pool` function so that it's available in the recovery shell alongside `set_rw_pool`
- On recovery shell exit, `/etc/mtab` is parsed and all ZFS filesystems are unmounted. This protects against a BE being mounted manually, which will break the discovery process.
- The recovery shell is now invoked as an interactive shell, which causes it to source `/.bashrc`. This allows /etc/profile to stay strictly focused on values that should be used for all bash processes, with prompts, aliases and completions moved to the interactive shells only.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -1710,6 +1710,33 @@ set_rw_pool() {
   return 1
 }
 
+# arg1: pool name
+# prints: nothing
+# returns: 0 on success, 1 on failure
+
+set_ro_pool() {
+  local pool ret
+
+  pool="${1}"
+  if [ -z "${pool}" ]; then
+    zerror "pool is undefined"
+    return 1
+  fi
+  zdebug "pool set to ${pool}"
+
+  if export_pool "${pool}" ; then
+    read_write='' import_pool "${pool}"
+    ret=$?
+
+    zdebug "import_pool: ${ret}"
+
+    return ${ret}
+  fi
+
+  return 1
+}
+
+
 # arg1: ZFS filesystem
 # prints: name of encryption root, if present
 # returns: 0 if system has an encryption root, 1 otherwise
@@ -2082,7 +2109,7 @@ zfs_chroot() {
 # returns: nothing
 
 emergency_shell() {
-  local message
+  local message skip mp fs
   message=${1:-unknown reason}
 
   tput clear
@@ -2091,7 +2118,17 @@ emergency_shell() {
 
   echo -n "Launching emergency shell: "
   echo -e "${message}\n"
-  /bin/bash -l
+
+  # -i (interactive) mode will source /.bashrc
+  /bin/bash -i
+
+  # shellcheck disable=SC2034
+  while read -r skip mp fs skip ; do
+    if [ "${fs}" == "zfs" ]; then
+      umount "${mp}"
+      zdebug "unmounting: ${mp}"
+    fi
+  done < /proc/self/mounts
 }
 
 # prints: nothing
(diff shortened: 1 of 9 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -33,7 +33,7 @@ zlog() {
   offset="${#prefix}"
 
   # account for the fzf gutter and kernel timestamp
-  _WIDTH=$(( $( tput cols ) - ( offset - 15 ) ))
+  _WIDTH=$(( COLUMNS - ( offset - 15 ) ))
```
ahesford, line 36: Is `$COLUMNS` guaranteed to be defined, or does `tput cols` buy anything extra?

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -2099,7 +2126,16 @@ emergency_shell() {
 
   echo -n "Launching emergency shell: "
   echo -e "${message}\n"
-  /bin/bash -l
+
+  # -i (interactive) mode will source /.bashrc
+  /bin/bash -i
```
ahesford, line 2131: I'm surprised bash doesn't source `/.bashrc` in `-l` mode.

Closer review of the man page suggests that `-l` does not necessarily imply `-i`, so maybe `bash -l -i` would be better because the login shell will source `/etc/profile` automatically. You could then drop the manual sourcing in `/.bashrc`. I don't know if this makes a difference anywhere.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -2099,7 +2126,16 @@ emergency_shell() {
...
+
+  while read -r _ mp fs _ ; do
+    if [ "${fs}" == "zfs" ]; then
+      umount "${mp}"
+      zdebug "unmounting: ${mp}"
+    fi
+  done < /etc/mtab
```
ahesford, line 2138: On my system, `/etc/mtab` is just a symlink to `/proc/self/mounts`. There is also a `/proc/mounts` that links to `/proc/self/mounts`.

I wonder whether there are cases on modern systems where `/etc/mtab` is a static file and whether we should prefer `/proc/mounts` or `/proc/self/mounts`, maybe falling through a chain of preferences if one can't be found.

ahesford (approved):

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: I found that `$COLUMNS` is only defined if `bash` is an interactive shell of some kind. If it's not, that variable is undefined. In either case, that's dead code now that I've rebased onto the new logging handler.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: It looks like if both `-l` and `-i` are passed in as arguments to bash, `-l` wins and `${HOME}/.bashrc` isn't sourced.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: This is probably a more universal way to ensure access to an `mtab` file. I've also removed usage of `_` as a dummy variable, `$_` is apparently a bash special variable.

PR #198, opened 2021-08-21
Create a global debug/trace log viewer

zdykstra:
The previous debug logging had a few limitations. It encoded color escape sequences in the log line, didn't show the full picture of what was calling what, it tried to wrap log lines to the terminal width, and it generally relied on fzf. This work attempts to address those issues.

Debug log lines are now semi-structured, and take the following form:
` ZBM:[pid]|<function,file,line;..>|<log message>`
`[    3.316907] ZBM:[556]|main,/bin/zfsbootmenu,124;draw_be,/lib/zfsbootmenu-lib.sh,493|selected: mod-r,ztest/ROOT/void`

Because the log lines are now semi-structured, the can be broken down to the component parts by a new tool, `ztrace`, and reconstructed in a more meaningful fashion. Log lines starting with `ZBM` are parsed out according to the above structure.

`ztrace` outputs look like this:

image

The output is piped into `less -R -S` automatically.


ahesford (approved):
This is awesome!

PR #199, opened 2021-08-21
Remove tmux integrations

zdykstra:
With global trace logging available via a hot key on almost any screen, or available over SSH / from the recovery shell, tmux serves a decreasingly limited purpose.

Removing it from the initramfs de-clutters/de-complicates quite a few areas of code.


ahesford (approved):
I never built an image with tmux, so I won't miss its absence.

PR #203, opened 2021-08-31
Start pushing users to drop /etc/default/grub

ahesford:
We're a real bootloader and we hate grub with a passion, so let's start the process of getting users away from its configuration file.

The color of the prompt is up for grabs, but let's stay away from red for now. We'll move red to the last release that supports the file.

Once we drop the deprecated behavior, I think we should hard fail if we detect an /etc/default/grub with non-empty KCL, then provide a simple helper people can run from the recovery shell to migrate the KCL to a property.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -1015,10 +1015,13 @@ find_be_kernels() {
         done
       done
     done
-
   done
 
+  # No further need for the mount
+  umount "${mnt}"
+
   defaults="$( select_kernel "${fs}" )"
+
   # shellcheck disable=SC2034
   IFS=' ' read -r def_fs def_kernel def_initramfs <<<"${defaults}"
 
@@ -1027,16 +1030,13 @@ find_be_kernels() {
   # If no default kernel is found, there are no kernels; leave the BE
   # directory in the same state it would be in had no /boot existed
   if [ -z "${def_kernel}" ]; then
-    umount "${mnt}"
     zdebug "no default kernel found for ${fs}"
     rm -f "${kernel_records}" "${def_kernel_file}"
     return 1
   fi
+
   zdebug "default kernel set to ${def_kernel}"
-
   echo "${def_kernel##*/}" > "${def_kernel_file}"
-
-  umount "${mnt}"
 
   # Pre-load cmdline arguments, possibly from files in the environment
   preload_be_cmdline "${fs}"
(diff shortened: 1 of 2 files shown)


zdykstra:
A little aggressive, but that's probably not a bad thing. I was going to suggest that `zwarn` be used, but most people don't boot to the main menu, so they'd never see that warning.

zdykstra (approved):

ahesford:
I've modified the auto-migration logic a bit:
1. The message is more detailed.
2. When auto-migrating, the pool must always be imported r/w, but the BE must be r/w only if we are removing `/etc/default/zfsbootmenu`.
3. Make sure the BE is unmounted before `preload_be_cmdline` exits.
4. When the user uses `[ESCAPE]` to ignore the auto-migration, the choice is stored globally in the exported variable `${zbm_ignore_kcl_deprecation}` and written to `/etc/zfsbootmenu.conf` so the warning will not appear again until the next reboot. The exported variable is important because, for example, jumping into a chroot after ignoring the message will trigger the warning again if the variable is not exported in the current shell. The `zfsbootmenu.conf` write is important to suppress the warning if the user quits the current shell (for example, by using `CTRL+R` to drop to an emergency shell).

This seems to do the right thing in general; @zdykstra, please take another look.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -1216,50 +1216,93 @@ read_kcl_prop() {
...
+  CLEAR_SCREEN=1 load_key "${fs}"
+
+  if ! mnt="$( allow_rw="${need_rw}" mount_zfs "${fs}" )"; then
+    zerror "unable to mount ${fs}"
+    return 1
+  fi
+
+  read -r args < "${args_file}"
+
+  if ! zfs set org.zfsbootmenu:commandline="${args}" "${fs}" ; then
```
zdykstra, line 1296: It might be good to capture the output of this `zfs set` so that it can be logged through either `zdebug` or `zerror` if it fails for some reason.

zdykstra:
Everything here is great - you've ironed out the rough edges I had from my commit.  VERY good call on using `/etc/zfsbootmenu.conf` to hold state, as well as adding start/end delimiters for what `-preinit` puts in.

90zfsbootmenu/zfsbootmenu-lib.sh
ahesford: Done

PR #204, opened 2021-09-01
Allow diffs between two snapshots

zdykstra:
You know what I'd do if I won a million dollars?

Two snapshots at the same time.

Add `--multi 2` to the snapshot browsing screen to pass in up to two arguments to `draw_diff` . If two arguments are passed in, the `creation` value for each snapshot is retrieved and the oldest (smallest) value dictates which snapshot is the first term passed to `zfs diff`. If only one snapshot is passed into `draw_diff`, the previous behavior of comparing that snapshot against the current state of the filesystem is used.

For non-diff browsing actions, the second snapshot (`mod-c,snapshot@1,snapshot@2`) is discarded and the first is exclusively acted upon.  The selected snapshot name is now shown at the top of the screen when creating a new duplicated/cloned boot environment.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -540,7 +540,7 @@ draw_kernel() {
   expects="--expect=alt-d,alt-u"
 
   if ! selected="$( HELP_SECTION=kernel-management ${FUZZYSEL} \
-     --prompt "${benv} > " --tac --with-nth=2 --header="${header}" \
+      --prompt "${benv} > " --tac --with-nth=2 --header="${header}" \
       ${expects} ${expects//alt-/ctrl-} ${expects//alt-/ctrl-alt-} \
       --preview="/libexec/zfsbootmenu-preview ${benv} ${BOOTFS}"  \
       --preview-window="up:${PREVIEW_HEIGHT}" < "${_kernels}" )"; then
@@ -556,7 +556,7 @@ draw_kernel() {
 }
 
 # arg1: ZFS filesystem name
-# prints: selected snapshot
+# prints: selected snapshot name, optionally a second snapshot
 # returns: 130 on error, 0 otherwise
 
 draw_snapshots() {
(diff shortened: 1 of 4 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -604,15 +604,33 @@ draw_snapshots() {
...
 
   snapshot="${1}"
   if [ -z "${snapshot}" ]; then
     zerror "snapshot is undefined"
     return 130
   fi
+
+  # if a second parameter was passed in and it's a snapshot, compare
+  # creation dates and make sure diff_target is newer than snapshot
+  if [ -n "${2}" ] && [[ "${2}" =~ @ ]]; then
```
zdykstra, line 618: The second test here is potentially useless; a second parameter from the snapshot screen should always be a snapshot. There's no way to select a non-snapshot value there.

90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -622,24 +641,30 @@ draw_diff() {
...
   # Bash won't use an FD referenced in a variable on the left side of a pipe
   exec 3>&"${zfs_diff[0]}"
 
   # shellcheck disable=SC2154
+  line_one="$( center_string "---${snapshot}" )"
+  left_pad="${line_one//---${snapshot}/}"
+  line_one="$( colorize red "${line_one}" )"
+  line_two="${left_pad}$( colorize green "+++${diff_target}" )"
+
   sed "s,${mnt},," <&3 | HELP_SECTION=diff-viewer ${FUZZYSEL} --prompt "${snapshot} > " \
```
ahesford, line 666: Since the preview now shows diff-like headers that provide necessary context, I think the prompt here should be something generic (maybe just a `>`). Using `${snapshot}` is a bit confusing since the diff contents really applies to both sides.

ahesford (approved):
Just tried this out in a test environment and it looks great. Aside from the noted prompt issue, it would be nice to find a way to tell the user that tab-selecting two snapshots is possible in the snapshot selection menu.

What about amending the header program to allow an optional "context" line? This might prove useful on some other menus, too.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Agreed. Chagned!

zdykstra:
> Just tried this out in a test environment and it looks great. Aside from the noted prompt issue, it would be nice to find a way to tell the user that tab-selecting two snapshots is possible in the snapshot selection menu.
> 
> What about amending the header program to allow an optional "context" line? This might prove useful on some other menus, too.

I like the idea of one or more lines of context help on some screens. Are you thinking those would be shown at the top of the screen, or at the bottom above/below the shortcut key text? If we put them at the top in place of the normal preview tool, the `Snapshot > ` text could instead be changed to the name of the ZFS filesystem currently selected.

PR #206, opened 2021-09-08
Early setup hooks, luks keystore contrib script

zdykstra:
Add in another hook - just after ZFS/SPL kmods are loaded but before anything else is attempted.

Add a proof-of-concept LUKS keystore script, intended to be used as an early setup hook.  This script will look for anything with a partition label of `KEYSTORE` and then verify that it is a block device with a LUKS header.  It then goes into a decryption prompt loop (broken by `ctrl-c`) to unlock the LUKS device. Upon successful unlock, the newly created `/dev/mapper/KEYSTORE` device is mounted to `/etc/zfs/keys` .  It's up to the end user to end user to to point any encryption root keys to be available on-disk at `/etc/zfs/keys` . 

This deserves some documentation in `README.md` and/or a wiki entry. 

image


ahesford (approved):
Looks good to me. I added some explanatory comments to the `contrib` hook; rework as you see fit.

PR #208, opened 2021-09-13
Make cache_key aware of mountpoints on key sources

ahesford:
When attempting to unlock a BE using keys from another filesystem named as a key source via org.zfsbootmenu:keysource, attempt to discover and strip the expected mountpoint from the key source when looking for keys. This allows, for example,

    zfs create -o mountpoint=/etc/zfs/keys zpool/keys

    zfs set org.zfsbootmenu:keysource=zpool/keys \
        keylocation=file:///etc/zfs/keys/zpool.key zpool/ROOT/bootenv

to work with ZFSBootMenu. Previously, the cache_key function would only properly identify keys when the key source had a mountpoint of `/`.

When mountpoint=none or mountpoint=legacy, or in the event a key does not exist at the new relative path, the prior behavior of looking for a path verbatim with respect to the key source is preserved.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -1925,7 +1925,8 @@ be_keysource() {
 # returns: 0 iff a key was cached
 
 cache_key() {
-  local fs keylocation keyfile keydir keysrc keycache mutex mnt ret
+  local fs mnt ret mutex keycache
+  local ksmount relkeyloc keypath keylocation keyfile keydir keysrc
 
   fs="${1}"
   if [ -z "${fs}" ]; then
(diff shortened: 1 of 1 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -1995,19 +1996,60 @@ cache_key() {
     return 1
   fi
 
+  relkeyloc=""
+  if ksmount="$(zfs get -o value -H mountpoint "${keysrc}" )"; then
+    case "${ksmount}" in
+      none|legacy)
+        zdebug "no discernable mountpoint for ${keysrc}, using only absolute key path"
```
zdykstra, line 2003: I rather enjoy that a debug message is using the word `discernable`.

zdykstra (approved):
After a read through or two, everything here makes sense. This doesn't look like it realistically increases the time `cache_key` will take to execute.

ahesford:
Merged in d208103

PR #209, opened 2021-09-14
Snapshot creation, reorganization

zdykstra:
Some times it's helpful to be able to create a snapshot from the bootloader, outside of the OS. On the snapshot screen, there's now a MOD+N option to create a (N)ew snapshot. It rides along the coat-tails of most of the snapshot cloning/duplication logic, as that has checks for everything we also care about.

The `MOD+S` handler in `zfsbootmenu.sh` was growing quite large. It created a substantial number of global variables. A large part of the logic in this handler has been moved to a dedicated subroutine that does all of the appropriate space checks, user input prompts/checking, and then dispatches to the right snapshot manipulation function.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -571,13 +571,12 @@ draw_snapshots() {
 
   sort_key="$( get_sort_key )"
 
-  header="$( header_wrap "[RETURN] duplicate" "[ESCAPE] back" "" \
-    "[CTRL+X] clone and promote" "[CTRL+C] clone only" "" \
-    "[CTRL+J] jump into chroot" "[CTRL+D] show diff" "" \
-    "[CTRL+L] view logs" "[CTRL+H] help" )"
+  header="$( header_wrap "[RETURN] duplicate" "[CTRL+C] clone only" "[CTRL+X] clone and promote" "" \
+    "[CTRL+N] create new snapshot" "[CTRL+J] jump into chroot" "[CTRL+D] show diff" "" \
+    "[CTRL+L] view logs" "[CTRL+H] help" "[ESCAPE] back" )"
   context="Note: for diff viewer, use tab to select/deselect up to two items"
 
-  expects="--expect=alt-x,alt-c,alt-j,alt-o"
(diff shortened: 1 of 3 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -901,6 +900,162 @@ clone_snapshot() {
...
+
+  subkey="${2}"
+  if [ -z "$subkey" ]; then
+    zerror "subkey is undefined"
+    return 1
+  fi
+  zdebug "subkey: ${subkey}"
+
+  local parent_ds
+  parent_ds="${selected%/*}"
```
ahesford, line 967: 1. I don't think we ever considered what happens if the BE is on the top-level filesystem. It seems like everything else would still work, but this probably requires
    ```sh
    parent_ds="${selected%/*}"
    parent_ds="${selected%@*}"
    ```
    to strip the snapshot name in that case.
2. Should this also test for empty `${parent_ds}` and fail in that case?

ahesford (approved):
Everything looks good here, but I make a couple of points about `${parent_ds}` for your consideration.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Do you mean the BE would be at `zroot` or `zroot/void` ?

ahesford: If the BE is at `zroot`, the selected snapshot will be `zroot@snap`. If `parent_ds` tries to strip out just `/*`, it will still be `zroot@snap`, and everything that follows will probably fail. If we also strip out `@*` from `parent_ds` for the case when there is no `/` in the selected name, I think all of the snapshot and clone logic still works. It just means that clones and send/recv duplicats will be *children* of the BE (at `zroot/clone`, for example), but that should still be OK.

This is probably not a common configuration, and most definitely isn't advisable, but we can probably support it for free.

zdykstra: Ah yeah. It's such an obscure/ill-advised combination that I wanted to make sure I was on the same page as you. I've added in the removal of `@snapshot`, and then a check to make sure that `parent_ds` is not empty.

PR #213, opened 2021-10-05
Default stty cols/rows values for serial consoles

zdykstra:
fzf doesn't handle a serial device that hasn't explicitly seen `stty rows` and `stty cols` called on it. Check if the controlling terminal is NOT the standard `/dev/tty[0-9]` device and then either set the user-provided values or default values for rows and columns.


90zfsbootmenu/zfsbootmenu-preinit.sh
```diff
@@ -4,15 +4,27 @@
...
 #shellcheck disable=SC2154
-if [ -n "$zbm_columns" ]; then
-  stty cols "$zbm_columns"
+if ! [[ ${control_term} =~ ${tty_re} ]]; then
+  #shellcheck disable=SC2154
+  if [ -n "$zbm_lines" ]; then
+    stty rows "$zbm_lines"
+  else
+    stty rows 25
+  fi
```
ahesford, line 20:
```suggestion
  stty rows "${zbm_lines:-25}"
```
for a one-shot replacement; also for `stty cols` below.

ahesford (changes requested):
This change will disable console resizing on the local VTs even if the variables are set. I'm OK with this; it doesn't make much sense to try resizing the physical console anyway (unless there are some cases where the kernel fails to determine the right size, but I haven't seen that). If this stays, we should update the documentation to reflect that these parameters are ignored with local consoles. Otherwise, something like
```sh
if ! [[ ${control_term} =~ ${tty_re} ]] || [[ -n "${zbm_lines}" ]] || [[ -n "${zbm_columns}" ]]
```
should do the trick.

90zfsbootmenu/zfsbootmenu-preinit.sh
zdykstra: good call, done.

zdykstra:
I think this is reasonable now. Since the check is in -preinit, the `stty` calls are called on ZBM startup. There's no way for `stty` to be run against an SSH pty, only the default `/dev/tty1` or a user-provided controlling terminal value. 

Since these KCL options aren't really documented or otherwise exposed, I don't think we need to add special-case support for them if people try to use them with `/dev/tty1` as the controlling terminal.

ahesford (approved):

PR #218, opened 2021-10-16
Add zreport function

zdykstra:
Move the zpool list / zfs list commands in -init to the function `zreport()`. This function prints the `zfs`/`spl` versions, the list of pools, and a report of filesystems + properties we care about.

`zreport` can be called from a recovery shell for an easy to record debugging picture/report.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-init.sh
+++ b/90zfsbootmenu/zfsbootmenu-init.sh
@@ -150,7 +150,7 @@ while IFS=$'\t' read -r _pool _health; do
   fi
 done <<<"$( zpool list -H -o name,health )"
 
-zdebug "$( zreport )"
+zdebug && zdebug "$( zreport )"
 
 unsupported=0
 while IFS=$'\t' read -r _pool _property; do
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -13,10 +13,6 @@ zlog() {
   [ -z "${1}" ] && return
   [ -z "${2}" ] && return
 
-  # Assume a default log severity of "warning" if not specified
-  # shellcheck disable=SC2154
-  [ "${1}" -le "${loglevel:=4}" ] || return
-
   # Remove everything but new lines from the string, count the length
   lines="${2//[!$'\n']/}"
   lines="${#lines}"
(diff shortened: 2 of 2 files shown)


zdykstra:
> Maybe bash needs lazy evaluation so we can defer calling `zreport` until we know the log level is sufficiently high.
> 
> You chould do
> 
> ```shell
> [ "${loglevel:-4}" -ge 7 ] && zdebug "$( zreport )"
> ```
> 
> to avoid the run. Yes, I know that double-tests `$loglevel`, but I wonder if it's worthwhile for people with large numbers of filesystems on spinning disks that might slow down the `zfs list`.
> 
> Your call.

What about something like this: 

```
diff --git a/90zfsbootmenu/zfsbootmenu-init.sh b/90zfsbootmenu/zfsbootmenu-init.sh
index 6b8bee8..570f916 100755
--- a/90zfsbootmenu/zfsbootmenu-init.sh
+++ b/90zfsbootmenu/zfsbootmenu-init.sh
@@ -150,7 +150,7 @@ while IFS=$'\t' read -r _pool _health; do
   fi
 done <<<"$( zpool list -H -o name,health )"

-zdebug "$( zreport )"
+zdebug && zdebug "$( zreport )"

 unsupported=0
 while IFS=$'\t' read -r _pool _property; do
diff --git a/90zfsbootmenu/zfsbootmenu-lib.sh b/90zfsbootmenu/zfsbootmenu-lib.sh
index beb1674..624e995 100755
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -46,19 +46,23 @@ zlog() {
 # returns: nothing

 zdebug() {
+  [ "${loglevel:-4}" -ge 7 ] || return 1
   zlog 7 "$@"
 }

 zinfo() {
+  [ "${loglevel:-4}" -ge 6 ] || return 1
   zlog 6 "$@"
 }

 zwarn() {
+  [ "${loglevel:-4}" -ge 4 ] || return 1
   : > "${BASE}/have_warnings"
   zlog 4 "$@"
 }

 zerror() {
+  [ "${loglevel:-4}" -ge 3 ] || return 1
   : > "${BASE}/have_errors"
   zlog 3 "$@"
 }
```

Call the log function with out any arguments and if it returns true, call it again with a (potentially) expensive operation.

PR #230, opened 2021-11-15
Provide, and use, a buildah script to construct zbm-builder images

ahesford:
Using buildah directly provides flexibility that can not be achieved with a Dockerfile. It also prevents the layer problem that bloats image sizes, avoiding the need to squash the image.

Eventually, we can do all sorts of neat things, like mount the container used to construct the image and manipulate the image from within a chroot.

Still testing, but I think it does what we expect.

### DIFF
--- a/releng/docker/README.md
+++ b/releng/docker/README.md
@@ -15,8 +15,28 @@ the commands should work just as well by substituting `docker` for `podman`.
 
 # Creating a ZFSBootMenu Builder Image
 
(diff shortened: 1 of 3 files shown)


zdykstra (approved):

zdykstra:
Needs a rebase onto master, otherwise this looks good! `buildah` sure beats a Dockerfile for ease of use.

PR #232, opened 2021-11-18
Define BE selection variables in teardown hook environment

### LINKED ISSUE #231: Xen support
Hello!

I use ZBM happily on quite a few systems & find its kexec-based boot chain to be the most reliable, no-compromises way to boot from an encrypted disk. On some systems I also don't use ZBM, & the reason for that is that they run [Xen](https://en.wikipedia.org/wiki/Xen).

Xen, being a type 1 hypervisor, runs the Linux kernel under its kernel. This is obviously not something ZBM supports as of right now. But it can! Xen [supports kexec perfectly well](https://xenbits.xenproject.org/docs/4.2-testing/misc/kexec_and_kdump.txt), & it requires no special tooling. Not even a kexec-tools built with `--with-xen`, unless you're booting zfsbootmenu's image under Xen too[1], I believe.

Supporting Xen means:

- for zfsbootmenu to be able to use the specific kexec syntax required to boot Linux under Xen[2]
-  being able to scan for Xen kernels (generally stored in the same directory as Linux kernels, e.g.: `/boot/xen-4.14.1.gz`[3])
     * selecting the latest one automatically, like what's already done with Linux kernels
     * as well as being able to select an arbitrary combination of Xen kernel & Linux kernel, including the possibility to run without the Xen kernel
- being able to set the Xen cmdline, hopefully independently of the Linux kernel cmdline for simplicity (they're not really related)
     * default set via, say, `org.zfsbootmenu:xen_commandline` or perhaps `org.zfsbootmenu:vmm_commandline` if you think a new type 1 hypervisor that boots the exact same way will arise at some point
     * tweakable in the UI, much like the Linux kernel cmdline

& I believe this should really be all that's needed on the ZBM side. I imagine a good way to detect whether to boot the system with Xen by default is just to check for the existence of Xen kernels in the same directory as the Linux kernels. That's usually a good sign, as even distros which enable Xen support for applications (e.g. libvirt) usually don't install its kernel in the same package. Void Linux (seemingly the "home distro" of the project) follows this pattern, with the Xen kernel being in the `xen-hypervisor` package, & Xen's userspace libraries being in `xen-libs`.

[1]: I can't exactly see why one would do this (it would let you have virtual machines in…the zfsbootmenu environment, that's it). Fortunately it doesn't change kexec-tools' commandline & instead just makes it use the Xen kexec hypercall instead of the kexec system call, so ZBM doesn't need to be specifically made aware of it. As kexec-tools comes from the system, the administrator can just rebuild it with `--with-xen` if they want to do this for whatever reason.
[2]: Example [here](https://xenbits.xenproject.org/docs/4.2-testing/misc/kexec_and_kdump.txt), section 1.a. NB: `DOM0_ARGS` is the Linux kernel cmdline, & `DOM0_IMAGE` is the Linux kernel image. I suggest passing `no-real-mode` (to [the Xen kernel](https://xenbits.xen.org/docs/unstable/misc/xen-command-line.html#no-real-mode-x86)) unconditionally if on x86. If it wasn't obvious, `--append` specifically is not needed & `--command-line` can be used as zfsbootmenu normally does.
[3] 4.14.1 may look like a Linux kernel version number, but it's not. It's the version of the Xen kernel.

ahesford:
To allow completely custom overrides of the final boot process, teardown hooks are called with three variables in their environment:

- ZBM_SELECTED_BE: filesystem holding the selected boot environment
- ZBM_SELECTED_KERNEL: path to the kernel within the selected BE
- ZBM_SELECTED_INITRAMFS: path to the corresponding initramfs

This is motivated by a desire to allow Xen users to inject a load/kexec into the following stage, effectively making ZBM responsible for selecting dom0 without making ZBM understand Xen at all.

Closes #231.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -782,8 +782,10 @@ kexec_kernel() {
   # Run teardown hooks, if they exist
   if [ -d /libexec/teardown.d ]; then
     for tdhook in /libexec/teardown.d/*; do
+      [ -x "${tdhook}" ] || continue
       zinfo "Processing hook: ${tdhook}"
-      [ -x "${tdhook}" ] && "${tdhook}"
+      env "ZBM_SELECTED_INITRAMFS=${initramfs}" \
+        "ZBM_SELECTED_KERNEL=${kernel}" "ZBM_SELECTED_BE=${fs}" "${tdhook}"
     done
     unset tdhook
   fi
--- a/man/zfsbootmenu.7
+++ b/man/zfsbootmenu.7
@@ -133,7 +133,7 @@
 .\" ========================================================================
 .\"
 .IX Title "zfsbootmenu 7"
-.TH zfsbootmenu 7 "2021-10-31" "1.11.0" "ZFSBootMenu"
+.TH zfsbootmenu 7 "2021-11-19" "1.11.0" "ZFSBootMenu"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
 .\" way too many mistakes in technical documents.
 .if n .ad l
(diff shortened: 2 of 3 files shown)


ahesford:
Note: I wavered between just running `ZBM_SELECTED_BE="${...}" "${tdhook}"`, between doing `export ZBM_SELECTED_BE=...` before the hook loop, or using `env` to set the environment as I did here. We force `env` into the image so it should always be valid. I'm open to suggestions in favor of the other options.

zdykstra:
> Note: I wavered between just running `ZBM_SELECTED_BE="${...}" "${tdhook}"`, doing `export ZBM_SELECTED_BE=...` before the hook loop, or using `env` to set the environment as I did here. We force `env` into the image so it should always be valid. I'm open to suggestions in favor of the other options.

I was waffling between creating a subshell and exporting the variables in it or doing it this way. Setting it via `env` seems perfectly reasonable.

zdykstra (approved):

PR #233, opened 2021-11-21
Support early/late hooks in generate-zbm

zdykstra:
Since generate-zbm only knows how to manage one ESP, it can be a challenge to keep multiple redundant ESPs up-to-date on a system. To help with this, Global.PreHooksDir and Global.PostHooksDir can now bedefined to point to a directory of executable hooks. These are executed without any additional environment context or arguments in the order a shell glob returns them.

`rsync` can be restricted to only manage ZFSBootMenu files through something like this:

`rsync --delete-after -avpP --include=zfsbootmenu\* --exclude=\* /boot/efi/EFI/void/ /tmp/second-efi`

`contrib/esp-sync.sh` has been added as an example post hook.

### DIFF
--- a/contrib/esp-sync.sh
+++ b/contrib/esp-sync.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+cleanup() {
+  if [ -n "${ESP_MNT}" ]; then
+    mountpoint -q "${ESP_MNT}" && umount -R "${ESP_MNT}"
+    [ -d "${ESP_MNT}" ] && rmdir "${ESP_MNT}"
+  fi
+}
+
+ESPS=(
+  "/dev/sdb1"
+  "/dev/sdc1"
+  "/dev/sdd1"
+)
+
+BMP="$( yq-go eval ".Global.BootMountPoint" /etc/zfsbootmenu/config.yaml )"
+if [ -z "${BMP}" ]; then
+  echo "Unable to determine BootMountPoint"
+  exit 1
+fi
+
+IMG_DIR="$( yq-go eval ".EFI.ImageDir" /etc/zfsbootmenu/config.yaml )"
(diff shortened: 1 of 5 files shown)


ahesford:
I could go one of two ways here:
1. If this in-config option is desirable, maybe we should make each hook an array of strings to allow proper handling of arguments with spaces:
    ```yaml
    LateHooks:
    - [ 'ls', '-lah', '/boot/efi/EFI/void' ]
    - [ 'rsync', '-avpP', '/boot/efi/EFI/void', '/supid path/to some other/void' ]
2. Just support `/etc/zfsbootmenu/generate-zbm.pre.d` and `/etc/zfsbootmenu/generate-zbm.post.d`, and `generate-zbm` can enumerate and run every executable in that directory at the right time, with no arguments.

If Perl can properly handle quoting of arguments in strings, so that something like
```yaml
LateHooks:
- 'rsync -avpP /boot/efi/EFI/void "/stupid path/to some other/void"'
```
will work, then I'm OK without using argv arrays.

contrib/esp-sync.sh
```diff
@@ -0,0 +1,45 @@
...
+IMG_REL="${IMG_DIR#"${BMP}"}"
+
+mount "${BMP}"
+
+if ! ESP_MNT="$( mktemp -d )"; then
+  echo "Unable to create temporary mountpoint"
+  exit
+fi
+
+trap 'rmdir ${ESP_MNT}' EXIT INT TERM
```
ahesford, line 30: You could add a check that ESP_MNT is a mount point and I mount if so. That should clean up if the script is interrupted between mount/unmount in the loop.

zdykstra: good call, done!

PR #234, opened 2021-12-01
Code removal / cleanup pass

zdykstra:
Skim hasn't seen a release since February 2021. It's starting to miss a
few features that we come to expect in fzf. Additionally, it has a few
rendering bugs / keybind issues that aren't easily resolved.

Removing skim support ends support for any platform not supported by Go,
but ZFS usage on those platforms isn't really common.

With all debug logging now running through the ztrace script, zlogtail
has been heavily simplified. It's now exclusively geared towards showing
warnning and error logs generated by the zwarn() and zerror() functions.

There is a `command -v fzf` test inside zfsbootmenu.sh - this can
possibly be removed. We test for fzf in module-setup.sh and critically
fail if it's missing.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -562,7 +562,7 @@ draw_kernel() {
 # returns: 130 on error, 0 otherwise
 
 draw_snapshots() {
-  local benv selected header expects sort_key snapshots multi
+  local benv selected header expects sort_key snapshots
 
   benv="${1}"
   if [ -z "${benv}" ]; then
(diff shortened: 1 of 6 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -586,12 +586,7 @@ draw_snapshots() {
 
   zdebug "snapshots: ${snapshots[*]}"
 
-  multi=( "--multi" )
-
-  # skim's --multi doesn't support any arguments
-  if [ "${FUZZYSEL}" == "fzf" ]; then
-    multi+=( "2" )
-  fi
+  multi=( "--multi" "2" )
```
ahesford, line 589: Just expand the array where it's used below, it will be clearer.

ahesford (approved):

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: doh, yes.

PR #236, opened 2021-12-16
Regardless of the source, always suppress root=

zdykstra:
More than one `root=` entry will result in the last one taking precedence - likely breaking the ability to boot into an environment. So regardless of the source, suppress that option on the BE's KCL.

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -1600,7 +1600,9 @@ load_be_cmdline() {
   # If a user-entered cmdline is found, it is not modified
   if [ -r "${BASE}/cmdline" ]; then
     zdebug "using ${BASE}/cmdline as commandline for ${zfsbe_fs}"
-    head -1 "${BASE}/cmdline" | tr -d '\n'
+
+    # root= is ALWAYS controlled by ZFSBootMenu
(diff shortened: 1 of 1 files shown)


90zfsbootmenu/zfsbootmenu-lib.sh
```diff
@@ -1600,7 +1600,11 @@ load_be_cmdline() {
   # If a user-entered cmdline is found, it is not modified
   if [ -r "${BASE}/cmdline" ]; then
     zdebug "using ${BASE}/cmdline as commandline for ${zfsbe_fs}"
-    head -1 "${BASE}/cmdline" | tr -d '\n'
+
+    # root= is ALWAYS controlled by ZFSBootMenu
+    zfsbe_args="$( suppress_kcl_arg root "$(head -1 "${BASE}/cmdline" | tr -d '\n')" )"
+
+    echo "${zfsbe_args}"
```
ahesford, line 1607: Why save a var to echo instead of just doing
```sh
suppress_kcl_arg root "$(head -1 "${BASE}/cmdline" | tr -d '\n')"
```

ahesford (approved):
Good idea, looks good to me.

90zfsbootmenu/zfsbootmenu-lib.sh
zdykstra: Good call.

PR #237, opened 2021-12-18
Add "settled" and "finished" hooks to force dracut event loop to fire

ahesford:
The only hooks guaranteed to run in the dracut event loop are "finished" hooks. If none are installed (or they all indicate success up front), no other "initqueue" or "settled" hooks will ever run. This includes, e.g., console initialization.

Rather than force console initialization in zfsbootmenu-preinit.sh, let's add a initqueue/settled and initqueue/finished hooks that, respectively, touch and watch for a file. This will force at least one iteration of the entire event loop (except for "timeout" hooks, which may still be short-circuited). If we want, we can make the settled hook more sophisticated and use it to stand up the necessary pre-requisites.

On real hardware, I have confirmed that the initqueue/finished hook runs several times (after other hook categories are run, the finish hooks are re-checked to determine whether to short-circuit the initqueue), the initqueue/settled hook runs at least once, and then the initqueue/finished hook runs a few more times (in my case, because other hooks are holding up the loop) before everything succeeds and ZBM starts.


zdykstra (approved):
It's FANTASTIC that you've cracked how to make networking with out `crypt-ssh` function. Well done!

PR #238, opened 2021-12-18
Configure the initramfs for SSH access

zdykstra:
If run.sh is invoked with -i or SSH_INCLUDE=1 is defined in
testing/.config, the wrapper script now does the following:

* Creates a dracut configuration file to enable crypt-ssh, listening on
  port 22 with ~/.ssh/authorized_keys set as the dropbear authentication
  ACL.
* Writes an SSH configuration file allowing ssh $test_environment to
  transparently connect you to the running ZFSBootMenu instance on port
  2222.
* Sets 'ip=dhcp rd.neednet' on the ZBM KCL.

If the option is not set, a dracut configuration file is written to omit
crypt-ssh.

The presence or absence of this flag does not control whether or not an
initramfs / EFI executable is generated.

### DIFF
--- a/testing/run.sh
+++ b/testing/run.sh
@@ -16,7 +16,7 @@ Usage: $0 [options]
   -s  Enable serial console on stdio
   -v  Set type of qemu display to use
   -D  Set test directory
-  -i  Write SSH config include
+  -i  Enable dropbear remote access via crypt-ssh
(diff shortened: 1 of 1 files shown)


ahesford (approved):
I wish this had been done long ago, it would have saved me some hassle.

PR #239, opened 2021-12-19
Allow rollbacks from the snapshot menu

ahesford:
Because BE rollbacks must be done before a system is booted, the justification for allowing rollback within ZBM is much the same as allowing checkpoint rewind. Let's offer this feature.

Missing from this implementation is a preliminary check to determine whether the snapshot being rolled back actually contains kernels. We're giving users a footgun, so I'm not sure we should try to stop them from blowing off the entire leg. (Really, I don't have a fantastic way to check and fail if kernels are missing. We can split `find_be_kernels` to avoid writing output---because we don't want to clobber the kernel list files if we abandon a rollback---and just return 0/1 depending on whether at least one kernel and initramfs appears to exist. I'm open to suggestions here.)

### DIFF
--- a/90zfsbootmenu/zfsbootmenu-lib.sh
+++ b/90zfsbootmenu/zfsbootmenu-lib.sh
@@ -575,10 +575,10 @@ draw_snapshots() {
 
   header="$( header_wrap "[RETURN] duplicate" "[CTRL+C] clone only" "[CTRL+X] clone and promote" "" \
     "[CTRL+N] create new snapshot" "[CTRL+J] jump into chroot" "[CTRL+D] show diff" "" \
-    "[CTRL+L] view logs" "[CTRL+H] help" "[ESCAPE] back" )"
+    "[CTRL+L] view logs" " " "[CTRL+H] help" "" "[ESCAPE] back" " " "[CTRL+R] rollback" )"
   context="Note: for diff viewer, use tab to select/deselect up to two items"
 
-  expects="--expect=alt-x,alt-c,alt-j,alt-o,alt-n"
+  expects="--expect=alt-x,alt-c,alt-j,alt-o,alt-n,alt-r"
 
   # ${snapshots} must always be defined so that the mod-n handler can be executed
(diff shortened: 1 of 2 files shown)


zdykstra:
I just tested this in various ways and it works exactly as expected. The footer text layout is a bit awkward, but I don't have any suggestions on how to improve that.

I'm torn on how many protections should be in place for a rollback. We don't have any in place when cloning / duplicating a snapshot into a full boot environment, so it's not unreasonable to skip that here as well. However, this IS a destructive action unlike cloning / duplicating a snapshot. If `find_be_kernels` can be reasonably chunked up to have a no-write / no-update mode and simply check for a kernel + initramfs, it wouldn't be bad to introduce that into the rollback procedure. We would have three options for how to handle a snapshot that can't boot:

1. If `find_be_kernels` returns 1, simply fail the rollback and log a reason via `zerror()`
2. If `find_be_kernels` returns 1, add an additional block of warning text to the rollback verification screen, explaining that unless you take additional action the BE will no longer be bootable.
3. If `find_be_kernels` returns 1, display a block of text explaining that the BE will be unbootable and if you really want to roll it back, here's how you can do it in the recovery shell

ahesford:
I didn't realize that `mount_zfs` preserves the snapshot name in the mount point `/zfsbootmenu/pool/path/to/filesystem@snapshot/mnt`. Likewise, `find_be_kernels`, `select_kernel` and `preload_be_cmdline` will all look in a path that includes the `@snapshot` part of the name. We can let `find_be_kernels` work as it does now without corrupting the kernel and cmdline files already produced for the parent filesystem. We waste some cycles, but things are kept simple. (Relatively speaking... there's a whole mess of stuff going on in cmdline and kernel handling.)

The force push checks for kernels and will refuse to roll back to a snapshot with no kernels, instead directing the user (without offering much help) to use the recovery shell.

zdykstra (approved):
I like this. It's not worth special-casing things in `find_be_kernels` when the work it does is so compartmentalized on a per-filesystem basis. If there's a pressing need for a shorter path through that function in the future, it can be added in easily - there are only three calls to `find_be_kernels` in the entire project.

PR #242, opened 2022-01-02
Support arbitrary colors in timed_prompt(), use getopts for arg parsing

zdykstra:
The function `decolorize()` has been added to strip color sequences from an arbitrary string and then echo the string for capture. With this function, we can now embed any number of color escape sequences in a string and then later strip them out to determine the 'on screen' size of the string for left pad purposes.

`timed_prompt()` now has arguments passed to it via `getopts`, so that we can more easily see exactly what is being passed to the function. I've semi-settled on a default color scheme for prompts and messages:

`[ESCAPE]` should be `red` in prompts/messages
`[RETURN]` should be `green` in prompts/messages

A ZFS pool or dataset should be `magenta`
A ZFS property should be `orange`
The countdown timer digits should be `yellow`

I've normalized all timed_prompt invocations to the new format, and tested the following prompts:

GRUB migration
Refusing to import r/w due to degraded status
Unable to find any kernels in a boot environment

### DIFF
--- a/90zfsbootmenu/bin/zfsbootmenu
+++ b/90zfsbootmenu/bin/zfsbootmenu
@@ -26,7 +26,9 @@ unset src sources
 mkdir -p "${BASE}"
 
 while [ ! -e "${BASE}/initialized" ]; do
-  if ! delay=5 prompt="Press [ESC] to cancel" timed_prompt "Waiting for ZFSBootMenu initialization"; then
+  if ! timed_prompt -d 5 -e "to cancel" \
+    -p "Waiting for ZFSBootMenu initialization"; then
+
     zdebug "exited while waiting for initialization"
     tput cnorm
     tput clear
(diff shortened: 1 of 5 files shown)


ahesford (commented):
I think this is generally fantastic. What do you think about adding special arguments for each of the RETURN and ESCAPE lines, so you can avoid repeating the colorization? Something like
```bash
timed_prompt -r "to engage return option" -e "to engage escape option"
```
The options will render "[RETURN]" and "[ESCAPE]" in the right colors and append the optarg to the line.

zdykstra:
> I think this is generally fantastic. What do you think about adding special arguments for each of the RETURN and ESCAPE lines, so you can avoid repeating the colorization? Something like
> ```bash
> timed_prompt -r "to engage return option" -e "to engage escape option"
> ```
> The options will render "[RETURN]" and "[ESCAPE]" in the right colors and append the optarg to the line.

That's a good idea! I'll get something hacked up based on this suggestion.

ahesford (approved):

PR #243, opened 2022-01-06
Add a shell utility to modify and review KCL properties

ahesford:
I think this utility is in a good state, but we should also have a manual page for it.

### DIFF
--- a/90zfsbootmenu/lib/echo-log-lib.sh
+++ b/90zfsbootmenu/lib/echo-log-lib.sh
@@ -9,25 +9,25 @@
 
 zdebug() {
   [ "${loglevel:-4}" -ge 7 ] || return 1
-  echo "DEBUG: $*" 
+  echo "DEBUG: $*" >&2
 }
 
 zinfo() {
   [ "${loglevel:-4}" -ge 6 ] || return 1
-  echo "INFO: $*"
+  echo "INFO: $*" >&2
 }
 
 znotice() {
   [ "${loglevel:-4}" -ge 5 ] || return 1
-  echo "NOTICE: $*"
+  echo "NOTICE: $*" >&2
 }
 
 zwarn() {
   [ "${loglevel:-4}" -ge 4 ] || return 1
-  echo "WARN: $*"
+  echo "WARN: $*" >&2
 }
 
 zerror() {
   [ "${loglevel:-4}" -ge 3 ] || return 1
-  echo "ERROR: $*"
+  echo "ERROR: $*" >&2
 }
--- a/90zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/90zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -902,7 +902,7 @@ read_kcl_prop() {
 
   zfsbe="${1}"
   if [ -z "${zfsbe}" ]; then
-    zerror "zfsbe is undefined"
+    zerror "boot environment is undefined"
     return 1
   fi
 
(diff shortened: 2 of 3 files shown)


bin/zbm_kcl
```diff
@@ -0,0 +1,264 @@
...
+
+edit() {
+  local bootenv kclfile oldsum newkcl newsum
+
+  bootenv="${1}"
+  kclfile="$(mktemp)"
+
+  trap 'rm -f "${kclfile}"' RETURN
+
+  if [ -z "${EDITOR}" ]; then
```
zdykstra, line 103: This conflicts with line 125. Can it default to `vi` here if undefined, and then only exit if `vi` isn't in $PATH ?

bin/zbm_kcl
```diff
@@ -0,0 +1,264 @@
...
+    zerror "failed to edit KCL for ${bootenv}"
+    return 1
+  fi
+
+  if ! newkcl="$(strip_kcl < "${kclfile}")"; then
+    zerror "failed to parse new KCL for ${bootenv}"
+    return 1
+  fi
+
+  if ! newsum="$(echo -n "${newkcl}" | md5sum | awk '{print $1}')"; then
```
zdykstra, line 135: `md5sum` is so old. Can we do `sha512sum` to future-proof this?

bin/zbm_kcl
```diff
@@ -0,0 +1,264 @@
...
+delete() {
+  local bootenv="${1}"
+
+  if ! zfs list -o name -H "${bootenv}" >/dev/null 2>&1; then
+    zerror "no boot environment specified"
+    usage
+    return 1
+  fi
+
+  if ! zfs inherit org.zfsbootmenu:commandline "${bootenv}"; then
```
zdykstra, line 64: How should things be handled in the event the property is set for `ztest/ROOT` ? delete() in this case will just restore the parent property to the BE.

bin/zbm_kcl
```diff
@@ -0,0 +1,264 @@
...
+    zerror "failed to compute checksum for new KCL"
+    return 1
+  fi
+
+  if  [ "${newsum}" = "${oldsum}" ]; then
+    znotice "modified KCL appears the same as original, will not change"
+    return 0
+  fi
+
+  zfs set org.zfsbootmenu:commandline="${newkcl}" "${bootenv}"
```
zdykstra, line 145: Errors here should be trapped like they are in the final `zfs set` at the end of the script.

zdykstra (changes requested):
I like the use/re-use of existing library functions here to do the heavy lifting. Those are very well-tested at this point, so their behavior is understood.

Any avoision to calling this `zbm-kcl` or `zbm-edit-kcl`? Underscores in program names are no bueno for me.

bin/zbm_kcl
ahesford: I meant to delete that when setting the default later, but I've now improved it to default to `vi` and fail early when `${EDITOR}` can't be found.

bin/zbm_kcl
ahesford: `md5sum` isn't as old as you. Changed to `sha512sum`.

bin/zbm_kcl
ahesford: That's the desired behavior. It's nice to be able to customize a specific boot environment and just use `-d` to clear the customization and restore the inherited behavior. I also think trying to be a bit craftier is likely to blow up strangely in some use cases.

bin/zbm_kcl
ahesford: Done.

ahesford:
I've renamed the file to `zbm-kcl`. I don't want to add `edit` in there because it's also useful just to query: `zbm-kcl` will just dump the KCL for the current root right to the console.

bin/zbm_kcl
zdykstra: I see, yup! Good call. I missed that you can specify a boot environment instead of just relying on finding the ZFS filesystem mounted at `/`.

zdykstra (approved):

PR #244, opened 2022-01-07
Improve KCL manipulations, drop dracut-lib dependency

ahesford:
I think this branch is ready for review and **thorough** testing.

Two things to consider:
1. We `mkdir -p ${BASE}` in a few places. We agree this redundancy is useful, but shoud we also *define* a default value `BASE=/zfsbootmenu` in all of these places? Right now `bin/zfsbootmenu` will make the directory but doesn't care if the value is set and this should probably change.
2. The new `zfsbootmenu-parse-commandline.sh` relies on the new `get_zbm_arg` and `get_zbm_bool` to write a cache file at `${BASE}/zbm.cmdline` that is also read by the `zbmcmdline` interactive helper. Should we manage the caching of the ZBM KCL in a different fashion?

These changes put us within striking distance of also supporting `mkinitcpio` or other initramfs generators. All we need are installation and "hook" scripts for each to populate the needed components and invoke -preinint as necessary. (At that point, we probably want to separate out the ZBM configuration part of `zfsbootmenu-parse-commandline.sh` to be independent of dracut assumptions.)

### DIFF
--- a/90zfsbootmenu/hook/zfsbootmenu-preinit.sh
+++ b/90zfsbootmenu/hook/zfsbootmenu-preinit.sh
@@ -21,7 +21,7 @@ export BASE="/zfsbootmenu"
 # shellcheck disable=SC2154
 cat >> "/etc/zfsbootmenu.conf" <<EOF
 # BEGIN additions by zfsbootmenu-preinit.sh
-export BASE="/zfsbootmenu"
+export BASE="${BASE}"
 export endian="${endian}"
 export spl_hostid="${spl_hostid}"
 export import_policy="${import_policy}"
--- a/90zfsbootmenu/hook/zfsbootmenu-preinit.sh
+++ b/90zfsbootmenu/hook/zfsbootmenu-preinit.sh
@@ -17,7 +17,6 @@ if ! [[ ${control_term} =~ ${tty_re} ]]; then
 fi
 
 export BASE="/zfsbootmenu"
-mkdir -p "${BASE}"
 
 # shellcheck disable=SC2154
 cat >> "/etc/zfsbootmenu.conf" <<EOF
@@ -37,8 +36,6 @@ export zbm_import_delay="${zbm_import_delay}"
 export control_term="${control_term}"
 # END additions by zfsbootmenu-preinit.sh
 EOF
-
-getcmdline | sed -e 's/^[ \t]*//' > "${BASE}/zbm.cmdline"
 
 # Set a non-empty hostname so we show up in zpool history correctly
 echo "ZFSBootMenu" > /proc/sys/kernel/hostname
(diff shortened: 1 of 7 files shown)


90zfsbootmenu/lib/zfsbootmenu-kcl.sh
```diff
@@ -0,0 +1,258 @@
...
+
+  for f in /etc/cmdline.d/*.conf; do
+    [ -r "${f}" ] || continue
+    kcl_tokenize < "${f}"
+  done
+
+  [ -r /proc/cmdline ] && kcl_tokenize < /proc/cmdline
+}
+
+# arg1..argN: argument to extract (multiples are alternatives, ordered by precedence)
```
zdykstra, line 176: `getarg()` from Dracut had the idea of a deprecated option. Supporting `-d option` would add excessive complexity / bloat to `get_zbm_arg()`. Would it instead be reasonable/doable to log in some fashion (either zdebug or zinfo level) when a non-primary argument is matched? It'd be of marginal value for us, and essentially no value to end-users.

zdykstra:
I wouldn't mind seeing `mkdir -p "${BASE:=/zfsbootmenu}"` as the pattern around that. If we're unsure if the directory exists, it's fair that `$BASE` might also be undefined.

When I was mulling this refactor over in my head, I envisioned that we'd collect all the of the different sources of KCL parameters and cache them in a file in some capacity. `${BASE}/zbm.cmdline` seems perfectly reasonable in that regard. Storing it as one token per line is both easy to read by a human, and simple to feed into the `kcl_` helper functions you've developed. I don't see there being any case where the cached file can be invalidated, so one-shot generating it as early as possible is fine.

90zfsbootmenu/lib/zfsbootmenu-kcl.sh
ahesford: Done. `get_zbm_arg` will now `zdebug` when the option that matches is not the first one specified.

90zfsbootmenu/hook/zfsbootmenu-preinit.sh
```diff
@@ -17,12 +17,11 @@ if ! [[ ${control_term} =~ ${tty_re} ]]; then
 fi
 
 export BASE="/zfsbootmenu"
-mkdir -p "${BASE}"
 
 # shellcheck disable=SC2154
 cat >> "/etc/zfsbootmenu.conf" <<EOF
 # BEGIN additions by zfsbootmenu-preinit.sh
-export BASE="/zfsbootmenu"
+export BASE="${BASE}"
```
zdykstra, line 24: For that final *chefs kiss* of protection, should this be `export BASE="${BASE:=/zfsbootmenu}"` as well?

ahesford: You're killing me, Smalls

zdykstra (approved):
I haven't put this PR through it's full paces - but as an aggregate I'm really liking this work.  Feel free to merge now - if I do happen to come across any small issues, I'll just commit a fix or run it by you on IRC.

PR #245, opened 2022-01-10
Profiling tools and initial optimizations

zdykstra:
If other languages can do performance profiling, why can't `bash` ? Our debug log format already had the bulk of the information we need to generate pre-function timing information. 

`run.sh -F` can now introduce a kernel command line option - `zbm.trace` which should point to a serial port. On the host side, the serial port is attached to a FIFO which dumps data into a log file in the test environment directory. On the ZBM side, a `debug` trap points to the serial device. Every command executed in the interpreter causes the trap to fire, which dumps the current function stack and source (file) stack along with the current unix TS with nanosecond precision to the serial port.

Once the VM exits, a small perl tool re-assembles the trace/performance data to the format expected by [Flamegraph](https://github.com/brendangregg/FlameGraph). `flamegraph.pl` should be in $PATH on your local install. It'll generate `${FLAMECHART}/flamechart.svg` and upload it if `webify` is on your system.

`/lib/profiling-lib.sh` can be sourced to enable the profiling/tracing trap if and only iff '$trace_term` is defined and the value points to a valid device.

Based on the flamechart generated, I've made the following changes:
* Omit `network-legacy` in the test environment, it spends a bunch of time messing with the KCL, etc
* Omit `rootfs-block`, `dm`, `dmraid` in the test environment - nothing we do needs what they provide and they're slow
* Add a dirty fast-fail to `get_zbm_arg`. When one lookup key is sent, use bashre to check if it's present in the KCL and return immediately if it's not.
* Below loglevel 7, replace `zdebug` with a function that executes `return 1` always
* Gate all zdebug calls in functions in the quick boot path with `zdebug &&`, avoiding building a string
* Replace a `get_zbm_bool primary secondary` call with two separate calls, to take advantage of fast fails

Before: https://somebits.link/u/p/d62bdcd0da/stream
After: https://somebits.link/u/p/94a6db104f/flamechart.svg

This is generally tested, but it could use a serious review of how the timing diff is generated and if it's valid. The downside to the profiler is that the trap, even as minimal as it is, adds a computation overhead. The timing values are consistent between runs, but they shouldn't be taken as absolute truth - they're inflated due to the time the trap takes to execute.


zdykstra:
Serial port speeds for both the control term and the trace term are now set via `stty`. I'm not sure if this works as intended on real hardware - I'm not sure how serial ports are used on physical systems.

ahesford (commented):
Nice work, this will be nice to have in the test setup.

90zfsbootmenu/lib/zfsbootmenu-kcl.sh
```diff
@@ -201,6 +201,13 @@ get_zbm_arg() {
 
   kcl="$( get_zbm_kcl )"
 
+  # Fail early for single arg lookup misses
+  if [ ${#@} -eq 1 ]; then
+    if ! [[ ${kcl} =~ ${1} ]]; then
+      return 1
+    fi
+  fi
+
```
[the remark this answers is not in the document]
zdykstra: Good idea. I've switched things up to operate in that fashion. The fastest invocations are a little slower now - but the slower invocations are now much faster for a ~100ms net gain.

90zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1221,17 +1224,13 @@ has_resume_device() {
...
-  prompt="Press $( colorize green "[RETURN]") or wait $( colorize yellow "%0.${#delay}d" ) seconds to continue"
-  message=()
+  local prompt delay message opt OPTIND
 
-  local opt OPTIND
   while getopts "d:p:m:r:e:" opt; do
     case "${opt}" in
       d)
         delay="${OPTARG}"
+        [ "${delay}" -gt 0 ] || return 0
```
ahesford, line 1233: If `$delay` is not numeric, this test will fail and 0 will be returned.
- What's the right behavior if the delay argument is bogus: using the default 30-second timer or dying with no prompt?
- If the answer is "dying with no prompt", is passing a bogus time more like pressing `RETURN` or `ESCAPE`?

If it's better to revert to the default 30-second timer when delay is bogus, it should be sufficient to switch to
```sh
[ "${delay}" -le 0 ] && return
```
If it's better to die but we want a bogus time to look like `ESCAPE` while a valid non-positive numeric time looks more like `RETURN`, consider the following:
```sh
if ! [ "${delay}" -gt 0 ]; then
  [ "${delay}" -le 0 ] || return 1
  return 0
fi
```
Any numeric comparison fails when the argument is not numeric, so moving the negation away from the operator means we do a single comparison in the most common case when `$delay` is a positive value.

90zfsbootmenu/module-setup.sh
```diff
@@ -162,6 +162,25 @@ install() {
...
   inst_hook initqueue/finished 99 "${moddir}/hook/zfsbootmenu-ready-chk.sh" || _ret=$?
 
+  # If tracing is enabled, build in the full-weight profiling library
+  if [ -n "${zfsbootmenu_enable_tracing}" ]; then
+    inst_simple "${moddir}/profiling/profiling-lib.sh" "/lib/profiling-lib.sh"
+
+    # shellcheck disable=SC2154
+    cat << EOF >> "${initdir}/etc/profiling.conf"
+export trace_term=${trace_term}
+export trace_baud=${trace_baud}
```
ahesford, line 172: This means that `trace_term` and `trace_baud` need to be configured at module setup time. Is that intended? If so, I recommend `zfsbootmenu_trace_term` and `zfsbootmenu_trace_baud`.

90zfsbootmenu/module-setup.sh
```diff
@@ -162,6 +162,25 @@ install() {
...
+EOF
+
+    # optionally enable early Dracut profiling
+    if [ -n "${dracut_early_tracing}" ]; then
+      inst_hook cmdline 00 "${moddir}/profiling/profiling-lib.sh"
+    fi
+  else
+    # Install the default profiling library that is a simple `return 0`
+    inst_simple "${moddir}/profiling/profiling-lib-noop.sh" "/lib/profiling-lib.sh"
```
ahesford, line 181: To avoid keeping some no-op file around, should this just be
```sh
echo "return 0" > "${whatever-the-installation-root-variable-is}/lib/profiling-lib.sh"
```

90zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh
```diff
@@ -30,9 +33,15 @@ fi
...
 
-if [ -z "${BYTE_ORDER}" ]; then
-  zwarn "unable to determine platform endianness; assuming little-endian"
-  BYTE_ORDER="le"
+# Only zwarn/zerror will log prior to loglevel being updated by a KCL argument
+min_logging=4
+loglevel=$( get_zbm_arg loglevel )
+if [ -n "${loglevel}" ]; then
+  # minimum log level of 4, so we never lose error or warning messages
+  [ "${loglevel}" -ge ${min_logging} ] || loglevel=${min_logging}
```
ahesford, line 41: The early source of `/lib/kmsg-logl-lib.sh` will assume the default log level and will have skipped the `traperror` stuff in that file. This isn't a huge deal because every other file should be souring the logger library, and this will redefine the functions on demand. However, we may wish to have proper logging for the rest of this file, so I recommend
```sh
if [ "${loglevel}" -ge "${min_logging}" ]; then
  source /lib/kmsg-log-lib.sh >/dev/null 2>&1
else
  loglevel="${min_logging}"
fi
```
or, if you adopt my include-guard recommendation for the logger library (see comment on kmsg-log-lib.sh), change the source line to
```sh
FORCE_RELOAD=1 source /lib/kmsg-log-lib.sh >/dev/null 2>&1
```

90zfsbootmenu/lib/kmsg-log-lib.sh
```diff
@@ -3,77 +3,78 @@
 
 : "${loglevel:=4}"
 
-# arg1: log level
-# arg2: log line
+# arg1..argN: log line
```
ahesford, line 6: If we're embracing include guards, let's do so here too because we keep re-sourcing this file. However, we probably want this file to be reloadable (see comment on zfsbootmenu-parse-commandline.sh, so I recommend the following:
```sh
if [ -n "${_ZFSBOOTMENU_KMSG_LOG_LIB}" ]; then
  [ -z "${FORCE_RELOAD}" ] && return
else
  readonly _ZFSBOOTMENU_KMSG_LOG_LIB=1
fi
```

We could make the same change in `echo-log-lib.sh`, but we don't actually use that anywhere except in the `zbm-kcl` utility.

ahesford (changes requested):

90zfsbootmenu/module-setup.sh
zdykstra: Done - I've normalized the control varriables here to `zfsbootmenu_trace_` .

90zfsbootmenu/module-setup.sh
zdykstra: Done - I originally had done this but then opted to use a file-on-disk for some unknown reason. Restored to this approach.

90zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh
zdykstra: Guard added, along with a re-import toggle.

90zfsbootmenu/lib/zfsbootmenu-core.sh
zdykstra: I opted to solve this in a slight different fashion, which I think covers all of the bases. I'm not dead-set on this course, but I think it offers the best compromise of validating user input vs time spent in `timed_prompt`. Thoughts?

PR #248, opened 2022-01-15
Generate release and recovery image builds

### LINKED ISSUE #227: zfs send | zfs recv/receive over a network connection?
Is it possible to bootstrap your filesystems over a network connection, or only via a local snapshot? If it's not possible to bootstrap over the network, are there any plans to implement such functionality?

zdykstra:
Add in support for a `recovery` style image which contains additional tools and networking support. Right now the additions are pretty minimal:

* gdisk
* parted
* mkfs.vfat / mkfs.ext4
* curl
* efibootmgr

The zfsbootmenu dracut module already installs mbuffer, lsblk and blkid into release images.  The changes behind the scenes allow for an arbitrary number of image styles to be made, with release and now recovery being made. EFI images are all made from a common container that must have all of the tooling available for every image style, so attention must be given to additional files/modules automatically being pulled in to the wrong image. I'm not convinced it's worth the effort though to make a new container per release style though.

Once this is merged and a new ZFSBootMenu release is cut, [redirect.pl](https://github.com/zbm-dev/get.zfsbootmenu.org/blob/master/redirect.pl) needs to be validated / updated to work with multiple EFI and .tar.gz files available per release.

This closes #227

### DIFF
--- a/etc/zfsbootmenu/recovery.conf.d/recovery.conf
+++ b/etc/zfsbootmenu/recovery.conf.d/recovery.conf
@@ -1,0 +1,16 @@
+zfsbootmenu_teardown+=" /zbm/contrib/xhci-teardown.sh "
+install_optional_items+=" /etc/zbm-commit-hash /bin/gdisk /bin/parted /bin/mkfs.vfat /bin/mkfs.ext4 /bin/curl /bin/efibootmgr"
+omit_drivers+=" amdgpu radeon nvidia nouveau i915 "
+
+# Network related modules
+add_dracutmodules+=" network network-legacy kernel-network-modules "
+omit_dracutmodules+=" crypt-ssh nfs lunmask "
+
+# qemu drivers
+omit_dracutmodules+=" qemu qemu-net "
+
+# filesystem and other related bits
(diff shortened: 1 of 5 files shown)


etc/zfsbootmenu/recovery.conf.d/recovery.conf
```diff
@@ -0,0 +1,7 @@
+zfsbootmenu_teardown+=" /zbm/contrib/xhci-teardown.sh "
+install_optional_items+=" /etc/zbm-commit-hash /bin/gdisk /bin/parted /bin/mkfs.vfat /bin/mkfs.ext4 /bin/curl /bin/efibootmgr"
```
zdykstra, line 2: I'm not really sure I like this, as it's very Void-specific. I've debated using a custom variable and looping through it in `module-setup.sh`. It could support things with an absolute path, or things expected to be in `$PATH` and install them as needed.

ahesford: Aren't you building this in a Void container?

zdykstra: > Aren't you building this in a Void container?

Ahh yeah - that is true. It's probably fine in that case.

PR #253, opened 2022-02-03
generate-zbm: remove INI migration support

zdykstra:
The migration to a YAML configuration file happened 2020-07-29. We've carried support for migrating from the legacy INI format long enough.


ahesford (approved):
A 2.0 release is a great time to drop legacy features. Don't forget to axe the forward declarations in lines 49, 50 as well.

PR #256, opened 2022-02-10
Use column to layout footer

zdykstra:
Realistically, we only need to print a footer in two modes:

1) using the explicit column layout designated by the caller - typically
   three columns with equal padding in columns

2) using a single column, due to a narrow screen

column supports both of these use cases by padding columnar data (as
indicated by embedded line returns in stdin).

item1:item2:item3\n
item4::item5

Would result in two rows with three columns. The second row would have
an empty middle column. By simply replacing ':' with a line return, we
can force everything into a single column.

New: https://somebits.link/u/p/5d0741fcb4/flamechart.svg?s=column_wrap
Old: https://somebits.link/u/p/37fd59c13e/flamechart.svg?s=header_wrap

`busybox` doesn't provide `column`, which is something to be considered. I'm generally okay with requiring tools from `util-linux`, though.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-lib.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-lib.sh
@@ -33,6 +33,7 @@ csv_cat() {
 }
 
 # arg1: colon-delimited string
+# arg2: short string used when column is missing / display is small
 # prints: string, columnized
 # returns: nothing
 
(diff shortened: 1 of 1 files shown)


zdykstra:
This supersedes #254 if it is merged.

ahesford (approved):
This cuts out a lot of complexity; looks good to me.

PR #258, opened 2022-02-12
Use an early-setup hook to force console init in release images

ahesford:
Setting rd.hostonly=0 in the release and recovery images causes dracut to throw out the initqueue hooks that ZBM installs to force the event loop to fire at least once. As a result, the console is often not initialized before ZBM takes control. ZFSBootMenu early-setup hooks provide a convenient place to force the console initialization, allowing things like fonts and keymaps to be set before password prompts might appear.

### DIFF
--- a/contrib/console-init.sh
+++ b/contrib/console-init.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+## This early-setup hook forces the dracut console initialization hook to run.
+##
+## ZFSBootMenu release builds embed rd.hostonly=0 in the kernel command line,
+## which causes the normal dracut initqueue to be thrown out; the ZFSBootMenu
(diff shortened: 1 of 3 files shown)


ahesford:
@Soulsuke If possible, can you test the EFI image built from this PR to see if it fixes your keymap problem? You should be able to download the UEFI build artifact at https://github.com/zbm-dev/zfsbootmenu/suites/5276094336/artifacts/163197353 (it packs the EFI image in a ZIP file).

ahesford:
Thanks for confirming. We will not bump the version for this fix, but I will recreate the release binaries with the new configuration. You should be able to just keep using the one you tested; it will be essentially the same.

Soulsuke:
@ahesford sounds fair, thanks for the support :)

PR #259, opened 2022-02-12
Force creation of the finished hook at runtime

zdykstra:
Setting rd.hostonly=0 in the release and recovery images causes dracut
to throw out the initqueue hooks that ZBM installs to force the event
loop to fire at least once. As a result, the console is often not
initialized before ZBM takes control.

Use a fake cmdline hook to create this for us after initqueue/finished/
has been destroyed.


zdykstra:
Closing this - #258 has been demonstrated to be correct. There is an interaction with QEMU, running with a GTK output and booting with an EFI BIOS that causes ZFSBootMenu to get the wrong console size, regardless of `console_init` running or not.

PR #260, opened 2022-02-14
zfsbootmenu-core.sh: drop preload_be_cmdline, improve KCL caching

ahesford:
I haven't actually tested this, but here's my proposal for improved KCL cache handling.

### DIFF
--- a/zfsbootmenu/bin/zfsbootmenu
+++ b/zfsbootmenu/bin/zfsbootmenu
@@ -275,11 +275,11 @@ while true; do
         def_args="${line}"
       done <<< "${BE_ARGS}"
 
-      echo -e "\nNew kernel command line"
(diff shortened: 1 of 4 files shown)


zdykstra (changes requested):
There are a few extra conditions under which a cache is no longer valid:

1. Any time after emergency_shell is called and the pool is `r/w`
2. Any time after zfs_chroot is called and the pool is `r/w`
3. Any time `/bin/zfsbootmenu` is started

These three things, and possibly others, could create `$BASE/cmdline.invalidate`. `validate_cmdline_cache` could check for the presence of that file, and if it's found, run through every BE defined in `$BASE/bootenvs` and remove the cached command line file there. After that, it could delete `$BASE/cmdline.invalidate`.

zdykstra (approved):
Fantastic compromise on caching / invalidating the cache. Nicely done!

PR #261, opened 2022-02-16
Replace 'tr' with bash string manipulation

zdykstra:
`tr` is only used in a few places, so this drops it from images without busybox. Might or might not be worth it. 

What might be more important is the additional use of `shopt -s extglob`. I'm debating turning that on globally and just fixing the fallout from it - if there is any. Thoughts?

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-lib.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-lib.sh
@@ -404,7 +404,10 @@ snapshot_dispatcher() {
 
     [ -n "${user_input}" ] || return
 
-    valid_name=$( echo "${user_input}" | tr -c -d 'a-zA-Z0-9-_.:' )
+    shopt -s extglob
+    valid_name="${user_input//+([!a-zA-Z0-9-_.:])/}"
+    shopt -u extglob
+
     if [[ "${user_input}" != "${valid_name}" ]]; then
       echo "${user_input} is invalid, ${valid_name} can be used"
       pre_populated="${valid_name}"
--- a/zfsbootmenu/install-helpers.sh
+++ b/zfsbootmenu/install-helpers.sh
@@ -20,7 +20,6 @@ zfsbootmenu_essential_binaries=(
   "sed"
   "grep"
   "tail"
-  "tr"
   "tac"
   "blkid"
   "awk"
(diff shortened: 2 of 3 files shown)


ahesford:
I fail to see how this is POSIX. We're never getting to suckless this way.

zdykstra:
> I fail to see how this is POSIX. We're never getting to suckless this way.

I'll PR the switch to dash next week.

zfsbootmenu/lib/zfsbootmenu-lib.sh
```diff
@@ -404,7 +404,10 @@ snapshot_dispatcher() {
 
     [ -n "${user_input}" ] || return
 
-    valid_name=$( echo "${user_input}" | tr -c -d 'a-zA-Z0-9-_.:' )
+    shopt -s extglob
+    valid_name="${user_input//+([!a-zA-Z0-9-_.:])/}"
```
ahesford, line 408: The set complement operator in bash is `!` and not `^`?!?

ahesford (approved):
I don't know enough about extglob to know the ramifications of turning it on globally. If it saves time, maybe it's worth looking into. On the flip side, I kind of like seeing it toggled exactly where needed to be explicit.

zfsbootmenu/lib/zfsbootmenu-lib.sh
zdykstra: `^` anchors the pattern to the start of the string. The best documentation I've found for extglob is at https://mywiki.wooledge.org/glob#extglob .

ahesford: Outside a character class, sure. But in character classes, usually a `^` immediately after the opening bracket means the class should be inverted.

zdykstra: https://wiki.bash-hackers.org/syntax/pattern#bracket_expressions

`^` and `!` are equivalent in this context. I can change it if you'd like.

ahesford: Not at all; I was just surprised.

zdykstra:
> I don't know enough about extglob to know the ramifications of turning it on globally. If it saves time, maybe it's worth looking into. On the flip side, I kind of like seeing it toggled exactly where needed to be explicit.

In my profiling test, there's effectively no time cost to the `shopt` on/off toggles. For now, I'd also like to explicitly enable/disable them. We can revisit this decision down the road if use of `extglob` increases.

PR #268, opened 2022-03-03
Tweaks to `zbm-builder.sh`

jrtoddy:
## Minor fixes to the `zbm-builder.sh` script.

ZBM-builder.sh script does not begin work on a fresh repository clone using Docker; however, it works fine using Podman. This is because Docker will not accept `.` or `./` as part of a `-v` command line argument, whereas Podman will.

`zbm-builder.sh -d` executes

`docker run --rm -v .:/build -w /build ghcr.io/zbm-dev/zbm-builder:latest -c /zbm/etc/zfsbootomenu/config.yaml`

Resulting in the following error...

`docker: Error response from daemon: create .: volume name is too short, names should be at least two alphanumeric characters.`

Expanding `.` to `./` is also not enough, as Docker reports.

`docker: Error response from daemon: create ./: "./" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.`

Thus, modifying `zbm-builder.sh` to use `$(pwd)` instead of `.` works for both Docker and Podman.

`$ docker/podman run --rm -v "$(pwd)":/build -w /build ghcr.io/zbm-dev/zbm-builder:latest -c /zbm/etc/zfsbootomenu/config.yaml`

Also, there was a little typo in the `BUILD_ARGS` line that passed an incorrect default `config.yaml` location to `zbm-build.sh` inside the container, causing build failure.

```
  BUILD_ARGS+=( "-c" "/zbm/etc/zfs**bootomenu**/config.yaml" )
```
Replacing boot**o**menu with bootmenu sorts this.

## Added option to use a local repository as opposed to a remote one.

As `zbm-builder.sh` is hard coded to pull the given tag from the zbm-dev/zfsbootmenu repository, this enables people to work from another repository using a local `git clone`. Combined with pulling images, this allows an offline edit-build-test workflow (i.e. it allows newbies such as myself to clone/pull and tinker locally. :baby:)

### DIFF
--- a/zbm-builder.sh
+++ b/zbm-builder.sh
@@ -95,7 +95,7 @@ fi
 
 # If no config is specified, use in-tree default
 if ! [ -r ./config.yaml ]; then
-  BUILD_ARGS+=( "-c" "/zbm/etc/zfsbootomenu/config.yaml" )
+  BUILD_ARGS+=( "-c" "/zbm/etc/zfsbootmenu/config.yaml" )
 fi
 
 # Make `/build` the working directory so relative paths in a config file make sense
--- a/zbm-builder.sh
+++ b/zbm-builder.sh
@@ -45,7 +45,7 @@ VOLUME_ARGS=()
 while getopts "hHLCdt:B:" opt; do
   case "${opt}" in
     L)
-      VOLUME_ARGS+=("-v" "$(pwd):/zbm")
+      VOLUME_ARGS+=("-v" "${PWD}:/zbm")
      ;;
     H)
       SKIP_HOSTID="yes"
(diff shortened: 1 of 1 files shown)


jrtoddy:
Hello everyone! :smile: This is my first pull request, so if anything is amiss please let me know.

ahesford:
Thanks for the contribution! Bash always defines `${PWD}`, so you should be able to replace the calls to `$(pwd)` with the variable to avoid shelling out to an external helper.

jrtoddy:
You are absolutely right, fixed.

zdykstra:
> You are absolutely right, fixed.

Please squash this PR down to a single commit after rebasing on master - no merge commits. Thanks!

PR #272, opened 2022-03-05
Bug fixes and use local repository option for zbm-builder.sh

jrtoddy:
### Fixes for the `zbm-builder.sh` script.

Unlike Podman, Docker does not accept `.` or `./` as arguments for bind volumes at the command line. Modifying `zbm-builder.sh` to use `${PWD}` instead of `.` works for both Docker and Podman.

A typo in the default `config.yaml` causes `zbm-build.sh` to fail, fixed.

### Added option to use a local repository as opposed to a remote one.

Currently, `zbm-build.sh` is hard-coded to pull the required tag from `zbm-dev/zfsbootmenu`. Adding this command line option allows builds using a local repository tree.

### DIFF
--- a/zbm-builder.sh
+++ b/zbm-builder.sh
@@ -7,6 +7,9 @@ Usage: $0 [options]
   OPTIONS:
   -h Display help text
 
+  -L Use local './' source tree instead of remote.
+     (Default: upstream master.)
+
   -H Do not copy /etc/hostid into image
      (Has no effect if ./hostid exists)
 
@@ -37,9 +40,13 @@ SKIP_CACHE=
 BUILD_TAG="ghcr.io/zbm-dev/zbm-builder:latest"
 
 BUILD_ARGS=()
+VOLUME_ARGS=("-v" "${PWD}:/build")
 
-while getopts "hHCdt:B:" opt; do
+while getopts "hHLCdt:B:" opt; do
   case "${opt}" in
+    L)
+      VOLUME_ARGS+=("-v" "${PWD}:/zbm")
+     ;;
     H)
       SKIP_HOSTID="yes"
       ;;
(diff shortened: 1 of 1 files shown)


jrtoddy:
Oops, my apologies @zdykstra. Hopefully this is better? :sweat_smile:

ahesford:
He means squash the commits, which you can do on your end by running

    git rebase -i master

This should open up an editor with one line for each commit, starting with `pick` and followed by the hash. Change the `pick` on every line **except the first** to `squash`, then save and quit the editor. Another editor should open up with all of the commit messages concatenation together; just remove them and write a combined message that makes sense. (The text you have here is fine as the commit message for the squashed commit.) Finally,

    git push -f

should force-push the new single-commit branch to your fork and be reflected in this PR.

jrtoddy:
Thanks and apologies @ahesford, let me know if this is OK. `rebase -i` was playing up but `git reset --soft HEAD~4` followed by `git push -f` did the trick.

zbm-builder.sh
```diff
@@ -95,8 +102,8 @@ fi
 
 # If no config is specified, use in-tree default
 if ! [ -r ./config.yaml ]; then
-  BUILD_ARGS+=( "-c" "/zbm/etc/zfsbootomenu/config.yaml" )
+  BUILD_ARGS+=( "-c" "/zbm/etc/zfsbootmenu/config.yaml" )
 fi
 
 # Make `/build` the working directory so relative paths in a config file make sense
-"${PODMAN}" run --rm -v ".:/build" -w "/build" "${BUILD_TAG}" "${BUILD_ARGS[@]}"
+"${PODMAN}" run --rm "${VOLUME_ARGS[@]}" -v "${PWD}:/build" -w "/build" "${BUILD_TAG}" "${BUILD_ARGS[@]}"
```
zdykstra, line 109: Instead of having two different ways to specify volumes (`$VOLUME_ARGS` and then hard-coded), you could instead push all volumes to `$VOLUME_ARGS`. The default value of `-v "${PWD}:/build"` can be pushed to the array when it's defined.

zdykstra (changes requested):
One last brief change and then this is good to go!

jrtoddy:
Ready to go, I believe. :+1:

zdykstra (approved):

zdykstra:
Thank you for the improvements / cleanup!

PR #282, opened 2022-03-20
zbm-builder.sh: various changes

jrtoddy:
**As discussed, following features are now present:**

- Specify local source directory (rather than assuming `$PWD`)
- Specify hook scripts for various stages
- `EFI` and `kernel/initramfs` pair will be generated if default `config.yaml` is used.

---

`zbm-builder.sh`: **Added `-l`ocal source option.**

Added `-l`ocal source path option (as lowercase, -C and -H being inhibitory.)
Added simple sanitise_path() function to help with docker path issues.

`zbm-builder.sh` - **Clarified help text**

Added short paragraph describing function of script and directing users to further information.

Clarified options `-B` and `-t`, which both used `tag` terminology.

The term `<tag>` now only refers to a release build tag, `<image>` replacing `<tag>` with regard to specifying an alternate container image.

`zbm-builder.sh`: **Kernel and EFI generation**

While not using a custom `config.yaml`, `zbm-builder.sh` will produce both kernel/initramfs pair and EFI bundle unless inhibited with the `-K` and `-E` options respectively.

`zbm-builder.sh`: **Collated all podman/docker checks**

Collated all docker/podman checks into one location in script.

`zbm-builder.sh`: **Rearranged/tweaked command line.**

Further re-wording of help text.

Switched `B` to `b` to conform to custom that capital letters prohibit actions.

Minor typo tweaks to error messages.

Re-arranged command line options in help text and script to alphabetical order, displaying augment/modify switches before prohibitive switches.

`zbm-builder.sh`: **Added `-i` switch to set build directory**

The `-i` switch allows user to choose an alternate build directory.

`zbm-builder.sh`: **Removed `-E` and `-K` switches**

As suggested, -E and -K switches are no longer exposed, but the script will still override default `config.yaml` (when used) to produce both EFI bundle and kernel/initramfs pair.

`zbm-builder.sh`: **swapped `-i` and `-b` options, typos.**

Swapped -i and -b switches; `-i` dictating image and `-b` dictating build directory makes more sense.

Typo fixes to help text, restored -h option text.

`zbm-builder.sh`: **switches to add hooks scripts**

New command line switches `-1`, `-2` and `-3` automate integration of hook scripts for the `zbm_early_setup`, `zbm_setup` and `zbm_teardown` stages respectively.

Added `sanitise_file()` function to check input is suitable hook script.

### DIFF
--- a/zbm-builder.sh
+++ b/zbm-builder.sh
@@ -80,7 +80,6 @@ fi
 while getopts "b:dhi:l:t:CH" opt; do
   case "${opt}" in
     b)
-      sanitise_path
       BUILD_DIRECTORY="${OPTARG}"
       ;;
     d)
@@ -94,7 +93,6 @@ while getopts "b:dhi:l:t:CH" opt; do
       BUILD_TAG="${OPTARG}"
       ;;
     l)
-      sanitise_path 
       VOLUME_ARGS+=( "-v" "${OPTARG}:/zbm" )
      ;;
     t)
@@ -150,4 +148,4 @@ if ! [ -r "${BUILD_DIRECTORY}"/config.yaml ]; then
 fi
 
 # Make `/build` the working directory so relative paths in a config file make sense
-echo "${PODMAN}" run --rm "${VOLUME_ARGS[@]}" -w "/build" "${BUILD_TAG}" "${BUILD_ARGS[@]}"
+"${PODMAN}" run --rm "${VOLUME_ARGS[@]}" -w "/build" "${BUILD_TAG}" "${BUILD_ARGS[@]}"
--- a/zbm-builder.sh
+++ b/zbm-builder.sh
@@ -66,7 +66,7 @@ Usage: $0 [options]
      (By default, a source archive is pulled by the container from GitHub.)
 
   -t <tag>
-     Build a specific ZFSBootMenu release tag (e.g. v1.12.0, v1.10.1, v1.9.0).
+     Build a specific ZFSBootMenu commit or tag (e.g. v1.12.0, v1.10.1, d5594589).
      (By default, the current upstream master is used.)
 
   -C Do not integrate /etc/zfs/zpool.cache into build image image from host.
--- a/zbm-builder.sh
+++ b/zbm-builder.sh
@@ -10,15 +10,6 @@ sanitise_path() {
   fi
 }
 
-sanitise_file() {
-  if [ -x "${OPTARG}" ] && ! [ -d "${OPTARG}" ]; then
-    OPTARG=$(readlink -f "${OPTARG}")
-  else
-    echo "Error: ${OPTARG} is not a valid, executable file."
-    exit 1
-  fi
-}
-
 usage() {
   cat << EOF
 
@@ -35,20 +26,6 @@ Usage: $0 [options]
 
   OPTIONS:
  
-  -1 <file>
-     Fetch inidcated file to build directory and mark for integration as
-     a 'zfsbootmenu_early_setup' hook. This command can be invoked multiple
-     times.
-
-  -2 <file>
-     Fetch inidcated file to build directory and mark for integration as
-     a 'zfsbootmenu_setup' hook. This comand can be invoked multiple times.
-  
-  -3 <file>
-     Fetch inidcated file to build directory and mark for integration as
-     a 'zfsbootmenu_teardown' hook. This command can be invoked multiple
-     times.
-
   -b <path>
      Use an alternate build directory for the container.
      (By default, the directory containing 'zbm-builder.sh' is used.)
@@ -100,20 +77,8 @@ else
   exit 1
 fi
 
-while getopts "1:2:3:b:dhi:l:t:CH" opt; do
+while getopts "b:dhi:l:t:CH" opt; do
   case "${opt}" in
-    1)
-      sanitise_file "${OPTARG}"
-      HOOKS_EARLY_SETUP+=( "${OPTARG}" )
-    ;;
-    2)
-      sanitise_file "${OPTARG}"
-      HOOKS_SETUP+=( "${OPTARG}" )
-    ;;
-    3)
-      sanitise_file "${OPTARG}"
-      HOOKS_TEARDOWN+=( "${OPTARG}" )
-    ;;
     b)
       sanitise_path
       BUILD_DIRECTORY="${OPTARG}"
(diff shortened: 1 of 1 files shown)


jrtoddy:
As `basename` is POSIX, I'm assuming it's acceptable in the `zbm-builder` script?

zbm-builder.sh
```diff
@@ -1,70 +1,145 @@
...
+     (If ./hostid exists, this switch will be ignored.)
 
+EOF
+}
+  
 SKIP_HOSTID=
 SKIP_CACHE=
 
 BUILD_TAG="ghcr.io/zbm-dev/zbm-builder:latest"
+BUILD_DIRECTORY=
```
ahesford, line 85: The build directory should have a sensible default and should be sanitized after all argument processing. Also, the local repo mount should be handled similarly, so
```suggestion
BUILD_DIRECTORY="."
LOCAL_REPO=""
```
Here, `$LOCAL_REPO` may remain empty to indicate that no local mount is desired, but a local build directory will always be used.

zbm-builder.sh
```diff
@@ -1,70 +1,145 @@
...
+      sanitise_file "${OPTARG}"
+      HOOKS_SETUP+=( "${OPTARG}" )
+    ;;
+    3)
+      sanitise_file "${OPTARG}"
+      HOOKS_TEARDOWN+=( "${OPTARG}" )
+    ;;
+    b)
+      sanitise_path
+      BUILD_DIRECTORY="${OPTARG}"
```
ahesford, line 119:
```suggestion
      BUILD_DIRECTORY="${OPTARG}"
```
Sanitizing and appending is not appropriate, because we don't want multiple specifications of `-b` to add multiple competing bind mounts. Just store the last argument encountered for later sanitization and mount configuration.

zbm-builder.sh
```diff
@@ -73,15 +148,17 @@ while getopts "hHLCdt:B:" opt; do
...
 
-if ! command -v "${PODMAN}" >/dev/null 2>&1; then
-  echo "ERROR: container front-end ${PODMAN} not found"
-  exit 1
+if [ -n "${BUILD_DIRECTORY}" ]; then
+  VOLUME_ARGS+=( "-v" "${BUILD_DIRECTORY}:/build" )
+else  
+  BUILD_DIRECTORY="${PWD}"
+  VOLUME_ARGS+=( "-v" "${PWD}:/build" )
 fi
```
ahesford, line 156: Now any configured mounts must be handled.
```suggestion
if ! BUILD_DIRECTORY="$( sanitise_path "${BUILD_DIRECTORY}" )"; then
  echo "ERROR: build directory does not exist"
  exit 1
fi
VOLUME_ARGS+=( "-v" "${BUILD_DIRECTORY}:/build" )

if [ -n "${LOCAL_REPO}" ]; then
  if ! LOCAL_REPO="$( sanitise_path "${LOCAL_REPO}" )"; then
    echo "ERROR: local repository does not exist"
    exit 1
  fi
  VOLUME_ARGS+=( "-v" "${LOCAL_REPO}:/zbm" )
fi
```

zbm-builder.sh
```diff
@@ -1,70 +1,145 @@
...
+    h)
+      usage
+      exit 0
+      ;;
+    i)
+      BUILD_TAG="${OPTARG}"
+      ;;
+    l)
+      sanitise_path 
+      VOLUME_ARGS+=( "-v" "${OPTARG}:/zbm" )
```
ahesford, line 133: As with `-b`, just overwrite the desired local repository.
```suggestion
      LOCAL_REPO="${OPTARG}"
```

ahesford (approved):
The build and repo support looks good, but I have a few suggestions for improvement.

I'm not convinced about supporting hooks in this fashion:
1. This only works for dracut, but we want to de-emphasize dracut and eventually build images with mkinitcpio by default.
2. Adding awareness of mkinitcpio as an alternative to work properly regardless of the initramfs generator gets even clumsier if we start supporting, *e.g.*, initramfs-tools.
3. The `-1`, `-2` and `-3` options are not obviously connected to the three stages of hooks.

I'm inclined to say that adding custom hooks is specialized enough that users should be writing configs and including a custom `dracut.conf.d` or `mkinitcpio.conf` in the build directory for this. However, I am open to arguments. @zdykstra do you have any comments?

If we do adopt support for hooks in this fashion, it might be good to avoid repetition by defining a function that takes a hook level (`setup`, `early_setup`, `teardown`) and a list of paths for that level, copies all files into the build directory and writes the appropriate configuration. The best way to loop over the list of paths and extract the basename would be
```sh
build_hooks=()
for hook in "${HOOKS[@}}"; do
  cp "${hook}" "${BUILD_DIRECTORY}"
  build_hooks+=( "/build/${hook##*/}" )
done
```

ahesford (changes requested):
Whoops, ticked the wrong box.

zbm-builder.sh
```diff
@@ -1,70 +1,145 @@
 #!/bin/bash
 # vim: softtabstop=2 shiftwidth=2 expandtab
 
+sanitise_path() {
+  if [ -d "${OPTARG}" ]; then
+    OPTARG=$(readlink -f "${OPTARG}")
+  else
+    echo "Error: ${OPTARG} is not a valid path."
+    exit 1
+  fi
```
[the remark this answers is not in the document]
jrtoddy: No problemo. :+1: I'll implement this.

zbm-builder.sh
jrtoddy: :+1:

zbm-builder.sh
jrtoddy: Understood. :+1:

zbm-builder.sh
jrtoddy: :+1:

zbm-builder.sh
jrtoddy: :+1:

jrtoddy:
> The -1, -2 and -3 options are not obviously connected to the three stages of hooks.

Admittedly, numerical options aren't the most elegant solution, but at least there are parallels with the three (possible) hook stages, rather than using whatever tortured single letter `-a`nologies I could contrive.

It seems that `getopts` supports `--long-and-friendly` options, would it be worthwhile adding these for all switches? `--hook-early`, `--hook-setup`, `--hook-teardown` would make sense.

> If we do adopt support for hooks in this fashion, it might be good to avoid repetition by defining a function that takes a hook level (`setup`, `early_setup`, `teardown`) and a list of paths for that level, copies all files into the build directory and writes the appropriate configuration.

I had thought about putting the hooks feature in a function much as you describe, and making the fetching of files another function too, but I wasn't sure how much  would be over-the-top. If desired, I'll get on it. :+1:

ahesford:
I think it would be better to just modify the script to look in `${BUILD_DIR}/{early_setup,setup,teardown}.d` or something like that for executable scripts. Command-line arguments add complexity just to copy scripts to the build directory. Furthermore, if people want to keep the same long-lived build directory around to update ZBM images using this script, it becomes unwieldy and the build directory becomes polluted with scripts.

The idea would be to determine whether each of these directories contains any executable scripts and, if so, write a `${BUILD_DIR}/dracut.conf.d/hooks.conf` file that points to the right paths in the containers. We can also write a corresponding `${BUILD_DIR}/mkinitcpio.d/hooks.conf` file for mkinitcpio users, then modify our default `mkinitcpio.conf` to source files in a `mkinitcpio.d` path. (This would involve modifying the container and can be a task for later.)

I've implemented this approach on top of your build directory and local repo support in the branch https://github.com/zbm-dev/zfsbootmenu/tree/corners-on-rails

jrtoddy:
> I've implemented this approach on top of your build directory and local repo support in the branch https://github.com/zbm-dev/zfsbootmenu/tree/corners-on-rails

@ahesford thank you for all your advice and taking the time to write at length for my benefit.  The directory-based approach does seem more elegant, from the point of view of a user that is re-building ZBM repeatedly over time but likely requiring the same hooks.

 I'll peel away the hook support from this pull request, if that suits? :+1:

ahesford:
Thanks, but don't worry about fixing up these commits; I will merge the modifications from the other branch that include your path improvements.

jrtoddy:
@ahesford Ah, got your message too late. :innocent: Feel free to close this pull request down anyhow. :+1:

PR #298, opened 2022-04-27
Basic EFI KCL editor

zdykstra:
This is intended to be a stand-alone tool that can edit the `.cmdline` segment in the release and recovery EFI files.  It's marginally tested. Some of the 'cleanup' behavior might need to be modified or improved upon.

### DIFF
--- a/bin/zbm-efi-kcl
+++ b/bin/zbm-efi-kcl
@@ -1,0 +1,139 @@
+#!/bin/bash
+# vim: softtabstop=2 shiftwidth=2 expandtab
+
+cleanup() {
+  if [[ -f "${TEMP_EFI}" && -z "${PRESERVE_TEMP_EFI}" ]]; then
+    rm "${TEMP_EFI}"
+  fi
+  [ -f "${TEMP_KCL}" ] && rm "${TEMP_KCL}"
+}
+
+usage() {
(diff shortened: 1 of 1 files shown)


bin/zbm-efi-kcl
```diff
@@ -0,0 +1,134 @@
...
+
+# Remove existing null terminators
+sed -i 's/\x0//g' "${TEMP_KCL}"
+
+# Collapse multiple lines into one
+sed -i -z 's/\n/ /g' "${TEMP_KCL}"
+
+# Pad with a space; replicating Dracut behavior.
+# TODO: Is this strictly needed?
+sed -i 's/^/ /' "${TEMP_KCL}"
```
ahesford, line 118: This could be replaced with `strip_kcl` in `bin/zbm-kcl`; maybe that could become a library function for reuse in both places.

ahesford (approved):
This looks great. If you want to generalize `strip_kcl` from `zbm-kcl`, have at it. Otherwise, merge as is.

bin/zbm-efi-kcl
zdykstra: For now I'd like to avoid any external libraries in this tool. Ideally it should be a "just download it" style script.

PR #300, opened 2022-06-01
Update generate-zbm

derzahla:
add aarch64 stub locations


zdykstra:
Does `/usr/lib/gummiboot/linuxaa64.efi.stub` exist in the wild anywhere? On `Void`, we only provide a stub for `x86_64` systems. Since most distributions use the systemd stub (formerly Gummiboot), I'm not sure we'll ever come across `linuxaa64.efi.stub`. Can you shed some light on that, since you're coming from an aarch64 system?

derzahla:
Maybe no need for the gummiboot path then.  I only have  /usr/lib/systemd/boot/efi/linuxaa64.efi.stub
 on Asahi.

PR #307, opened 2022-06-25
Dev/remote ssh access

cbreak-black:
Extend the image-build.ssh script to install dependencies, and zbm-build.sh / zbm-builder.sh to bundle keys and configuration as described in [0]. The created image will contain ssh host private keys, and user public keys.

[0] https://github.com/zbm-dev/zfsbootmenu/wiki/Remote-Access-to-ZBM

### DIFF
--- a/docs/REMOTE-SSH.md
+++ b/docs/REMOTE-SSH.md
@@ -1,0 +1,56 @@
+# Remote SSH Support
+
+There are several ways to do this, below is only one possible option (which has
+worked at least once in the past). The script changes are heavily based on the
+instructions in the [Remote-Access-to-ZBM](https://github.com/zbm-dev/zfsbootmenu/wiki/Remote-Access-to-ZBM)
+
+
+## Preparation
+
+The ZBM initrd will contain its own SSH host keys, which will be created by
+`zbm-builder.sh` if they are not already present in `etc/dropbear` as
+`ssh_host_ecdsa_key` and `ssh_host_rsa_key`.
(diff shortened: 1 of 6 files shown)


ahesford:
I think supporting an SSH-capable build within the container is definitely worthwhile. Supporting this via dedicated flags brings too much special-purpose logic into the build scripts, which should be as generic as possible. (For example, forcing `ip=dhcp` in `/etc/cmdline.d/dracut-network.conf` precludes static IP configurations.)

As an alternative, I have put together https://github.com/zbm-dev/zfsbootmenu/pull/308 to provide, in a general manner, the major knobs necessary for customizations required to include a working dropbear configuration in ZBM images. This is done via the addition of `-p` command-line arguments to the `image-build.sh` and in-container `zbm-build.sh` scripts that allow container images *or* instantiated containers to be augmented with custom package lists; the `-p` argument is also exposed in the user tool `zbm-builder.sh` and forwards this to the container build script, as well as a `-v` argument that is forward to `podman` or `docker` to allow custom bind mounts in the build container.

The example in the PR details the process for including a working dropbear configuration in ZBM images produced with the enhanced container utilities. Bind-mounting `/etc/cmdline.d` to include the desired Dracut network configuration, adding the `dracut-crypt-ssh` package to the instantiated container, and adding a few configuration files to the build directory are sufficient to pull in all the bits outlined in my wiki entry.

cbreak-black:
See https://github.com/zbm-dev/zfsbootmenu/pull/311

PR #308, opened 2022-06-27
Enhance configurability of containerized builds

ahesford:
- Allow container images to include custom Void packages via image-build.sh
- Allow container instances to include custom Void packages via zbm-build.sh
- Allow specification of custom Void packages and volume mounts via user-facing zbm-builder.sh

This can be used, *e.g.*, to build a custom dropbear-enabled image via `zbm-builder.sh` with a few simple modifications to the build directory used in the containerized build process:
1. Create the file `cmdline.d/dracut-network.conf` with contents `ip=dhcp rd.neednet=1`
2. Create the directory `dropbear` and place desired OpenSSH host keys (public *and* private) as well as an `authorized_keys` file with keys allowed to connect to ZBM
3. Create the file `dracut.conf.d/dropbear.conf` with contents, *e.g.*,
```sh
add_dracutmodules+=" crypt-ssh "
install_optional_items+=" /etc/cmdline.d/dracut-network.conf "
dropbear_keytypes="<your-keytypes-here>"
dropbear_acl=/build/dropbear/authorized_keys
dropear_<keytype>_key=/build/dropbear/ssh_host_<keytype>_key
dropbear_port=2222
```
4. Run the `zbm-builder.sh` script with extra arguments to make `dracut-crypt-ssh` and the `cmdline.d` available to the ZBM build container:
```sh
zbm-builder.sh [-i zbuild] -p dracut-crypt-ssh -v ./cmdline.d:/etc/cmdline.d:ro
```
> Note: for now, `-i zbuild` points to a locally generated `zbuild` container image that includes the new `zbm-build.sh` which recognizes the `-p` argument to specify packages; once the new image is published on GHCR, the default public image can be used instead.

### DIFF
--- a/releng/docker/image-build.sh
+++ b/releng/docker/image-build.sh
@@ -1,12 +1,51 @@
 #!/bin/sh
 # vim: softtabstop=2 shiftwidth=2 expandtab
 
+usage() {
+  cat <<-EOF
+	USAGE: $0 [-h] [-p <pkg>] <tag> [zbm-commit-like]
+	Create a container image for building ZFSBootMenu
+	
+	-h: Display this message and exit
+	
+	-p <pkg>:
+	   Include the specified Void Linux package in the image
+	   (One package per argument; may be repeated as needed)
+	
+	<tag>:
+	   Tag assigned to container image
+	
+	[zbm-commit-like]:
+	   ZFSBootMenu commit hash, tag or branch name used by
+	   default to build ZFSBootMenu images (default: master)
+	EOF
+}
+
 set -o errexit
+
+extra_pkgs=""
+while getopts "hp:" opt; do
+  case "${opt}" in
+    p)
+      extra_pkgs="${extra_pkgs} ${OPTARG}"
+      ;;
+    h)
+      usage
+      exit
+      ;;
+    *)
+      usage
+      exit 1
+      ;;
+  esac
+done
+
+shift $((OPTIND-1))
 
 # A tag for the image is required
 tag="${1}"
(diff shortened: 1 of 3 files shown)


cbreak-black:
This looks a lot more flexible than my solution, but less easy to use. Making a wrapper script around the new `zbm-builder.sh` that handles the key creation and other integrity checks might be useful. But I'm not sure how generic this can be made.

releng/docker/image-build.sh
```diff
@@ -1,12 +1,51 @@
...
+	
+	[zbm-commit-like]:
+	   ZFSBootMenu commit hash, tag or branch name used by
+	   default to build ZFSBootMenu images (default: master)
+	EOF
+}
+
 set -o errexit
 
+xtra_pkgs=""
```
zdykstra, line 26: Why this is `xtra_pkgs` instead of `extra_pkgs` ?

releng/docker/image-build.sh
```diff
@@ -1,12 +1,51 @@
 #!/bin/sh
```
zdykstra, line 1: This is unrelated to the PR, but why are we using `/bin/sh` in this script, with `/bin/bash` used virtually everywhere else?

zdykstra (approved):
This a good general-purpose change set. As @cbreak-black noted, it's not quite as directly usable as his laser-focused PR. A good compromise would be a `/contrib` script that wraps `zbm-builder.sh`. It'd need to be fairly user-friendly, with embedded documentation, though.

PR #309, opened 2022-06-28
Exploit common configurations for recovery/release images

ahesford:
This is tested only to the point that I prepared some release assets with `make-binary.sh` and did an `lsinitrd` on the release and recovery EFI bundles to confirm that some contents are expected.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,12 +10,18 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
+      - name: Prepare repository for build
+        run: |
+          rm -rf /zbm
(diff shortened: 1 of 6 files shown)


zdykstra (changes requested):
These changes look good. The `.github/workflows/build.yml` configuration file will need a few small changes to it to understand the new file layout.

Might be a reasonable time to investigate omitting the `drm` Dracut module as well, it seems to be strictly focused to KMS-related additions.

ahesford:
The artifacts need testing, but I revamped the workflow and dropped `drm`.

zdykstra (approved):

PR #311, opened 2022-07-03
remote ssh contrib

cbreak-black:
This branch contains a script using the new `zbm-builder.sh` functionality to build an ssh-enabled ZBM. It replaces https://github.com/zbm-dev/zfsbootmenu/pull/307.


contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+# WARNING: These private keys will be part of the final image!
+RS_PK_RSA="${RS_KEYDIR}"/ssh_host_rsa_key
+RS_PK_ECDSA="${RS_KEYDIR}"/ssh_host_ecdsa_key
+RS_KEYGEN_OPT=(-m PEM -N '' -C "zbm-server-key")
+if [ ! -f "${RS_PK_RSA}" ]; then
+  ssh-keygen -t rsa "${RS_KEYGEN_OPT[@]}" -f "${RS_PK_RSA}"
+fi
+if [ ! -f "${RS_PK_ECDSA}" ]; then
+  ssh-keygen -t ecdsa "${RS_KEYGEN_OPT[@]}" -f "${RS_PK_ECDSA}"
+fi
```
ahesford, line 38: Make this more general to allow easty and arbitrary customization of the keytypes to be generated.
```suggestion
RS_SSH_KEYTYPES=(rsa ecdsa)
RS_KEYGEN_OPT=(-m PEM -N '' -C "zbm-server-key")
for keytype in "${RS_SSH_KEYTYPES[@]}"; do
  RS_PK_FILE="${RS_KEYDIR}/ssh_host_${keytype}_key"
  [ -f "${RS_PK_FILE}" ] && continue
  ssh-keygen -t "${keytype}" "${RS_KEYGEN_OPT[@]}" -f "${RS_PK_FILE}"
done
```

contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+
+
+## PREPARE AUTHORIZED KEYS
+
+RS_AUTH="${RS_KEYDIR}/authorized_keys"
+if [ ! -f "${RS_AUTH}" ]; then
+  echo "Cannot find ${RS_AUTH}, copying from current user"
+  cp -v ~/.ssh/authorized_keys "${RS_AUTH}"
+fi
```
ahesford, line 47:
```suggestion
if [ ! -f "${RS_AUTH}" ] && [ -r "${HOME}/.ssh/authorized_keys" ]; then
  echo "Cannot find ${RS_AUTH}, copying from current user"
  cp -v ${HOME}/.ssh/authorized_keys "${RS_AUTH}"
fi
```

contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+
+
+## PREPARE SETTINGS
+
+RS_DNC="${RS_DIR}/dracut-network.conf"
```
ahesford, line 52: If the entire `cmdline.d` is mounted rather than a single-file bind, it becomes easy to inject other customized cmdline.d snippets and avoids polluting the build directory with a bunch of files that have different conceptual purposes.

```suggestion
mkdir -p "${RS_DIR}/cmdline.d"
RS_DNC="${RS_DIR}/cmdline.d/dracut-network.conf"
```

contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+if [ ! -f "${RS_DDC}" ]; then
+  cat >> "${RS_DDC}" <<EOF
+# Enable dropbear ssh server and pull in network configuration args
+add_dracutmodules+=" crypt-ssh "
+install_optional_items+=" /etc/cmdline.d/dracut-network.conf "
+dropbear_rsa_key=/etc/dropbear/ssh_host_rsa_key
+dropbear_ecdsa_key=/etc/dropbear/ssh_host_ecdsa_key
+dropbear_acl=/etc/dropbear/authorized_keys
+EOF
+fi
```
ahesford, line 70: Just dump this right in `dracut.conf.d`, and why bother mounting stuff into `/etc/dropbear` when `./dropbear` is visible inside the container as `/build/dropbear`? We can point right to that.

Also, add tabs to make things more readable. (They have to be hard tabs, not the two spaces added by the modeline, so `Ctrl-v <tab>` it is.
```suggestion
mkdir -p dracut.conf.d
RS_DDC="${RS_DIR}/dracut.conf.d/dracut-dropbear.conf"
if [ ! -f "${RS_DDC}" ]; then
  cat >> "${RS_DDC}" <<-EOF
	# Enable dropbear ssh server and pull in network configuration args
	add_dracutmodules+=" crypt-ssh "
	install_optional_items+=" /etc/cmdline.d/dracut-network.conf "
	dropbear_rsa_key=/build/dropbear/ssh_host_rsa_key
	dropbear_ecdsa_key=/build/dropbear/ssh_host_ecdsa_key
	dropbear_acl=/build/dropbear/authorized_keys
	EOF
fi
```

contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+
+# zbm-build.sh will copy this into the right place since mounting it into
+# /etc/zfsbootmenu/dracut.conf.d breaks the build process
+mkdir -p "dracut.conf.d"
+cp -f "${RS_DDC}" "dracut.conf.d/"
```
ahesford, line 75: No longer necessary with changes above.

contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+
+
+## ZBM BUILDING
+
+./zbm-builder.sh \
+  -p dracut-crypt-ssh -p dropbear \
+  -v "${RS_KEYDIR}":/etc/dropbear \
+  -v "${RS_DNC}":/etc/cmdline.d/dracut-network.conf \
```
ahesford, line 83: Volume mounts can be reworked with changes above.
```suggestion
  -v "${RS_DIR}/cmdline.d":/etc/cmdline.d:ro \
```

contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+  "${@}"
+
+
+# CLEANUP
+
+rm -f "dracut.conf.d/dracut-dropbear.conf"
+rmdir --ignore-fail-on-non-empty "dracut.conf.d"
```
ahesford, line 90: Don't clean up when `dracut.conf.d/dracut-dropbear.conf` becomes the only source of truth.

contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+# zbm-build.sh will copy this into the right place since mounting it into
+# /etc/zfsbootmenu/dracut.conf.d breaks the build process
+mkdir -p "dracut.conf.d"
+cp -f "${RS_DDC}" "dracut.conf.d/"
+
+
+## ZBM BUILDING
+
+./zbm-builder.sh \
+  -p dracut-crypt-ssh -p dropbear \
```
ahesford, line 81: XBPS should pick up `dropbear` as a dependency to `dracut-crypt-ssh`, so its specification isn't necessary here. (It doesn't harm anything to leave it.)

contrib/remote-ssh-build.bash
```diff
@@ -0,0 +1,90 @@
...
+# working directory by default). Its contents can be customized as needed.
+# They are regenerated if missing.
+
+# Server private keys are generated automatically as needed. It is
+# recommended to not reuse existing private keys.
+# User authentication for the root account is done via 'authorized_keys'.
+# The current users' authorized keys are copied if it is not present.
+
+RS_DIR=$(realpath "${RS_DIR:-ssh-data}")
+RS_KEYDIR="${RS_DIR}/keys"
```
ahesford, line 20: Calling this `dropbear` is a bit more descriptive.
```suggestion
RS_KEYDIR="${RS_DIR}/dropbear"
```

docs/REMOTE-SSH.md
```diff
@@ -0,0 +1,46 @@
...
+```
+mkdir -p -m 700 /etc/zfs/keys
+cp passwordfile /etc/zfs/keys/rpool.key
+zfs change-key \
+    -o keylocation=file:///etc/zfs/keys/rpool.key \
+    -o keyformat=passphrase rpool
+update-initramfs -u
+chmod go= /boot
+```
+
```
ahesford, line 36: This whole section is beyond the scope of remote unlocking, and is described in general setup docs anyway.

ahesford (changes requested):
Thanks for putting this together. This is a nice script. I have a few suggestions to keep the build directory a bit neater and simplify the volume mounts in the container.

Please change the extension on the script from `.bash` to `.sh`. Yes, it is a bash script and uses bashisms; but that is the case eveywhere in the repo and we don't adopt the `.bash` extension anywhere else.

As for the documentation, I would rather the contents of `REMOTE-SSH.md` make their way into the remote access wiki page as an addendum instead of residing in the docs tree. Alternatively (or in addition to content in the wiki), merge the description in `REMOTE-SSH.md` with comments at the head of `remote-ssh-build.sh` so the script and its documentation are self-contained.

contrib/remote-ssh-build.bash
cbreak-black: I like this improvement, but I don't think any other key formats will actually be useful (not without also changing the config file for dropbear in sync, and I'm not sure if it supports more formats)

ahesford: We support three key types in [dracut-crypt-ssh](https://github.com/dracut-crypt-ssh/dracut-crypt-ssh/commit/6ae8c358d24989ac7c6e625609bd1af4cf8ce1de); it is probably equally useful to allow people to strip out unwanted keys as it is to allow them to add something like ed25519.

contrib/remote-ssh-build.bash
cbreak-black: I improved this section by incorporation your suggestion, and also searching in the home of `${SSH_USER}` before `${HOME}`.

contrib/remote-ssh-build.bash
cbreak-black: I was worried that I'd override files created by other stages in the build pipeline.

docs/REMOTE-SSH.md
cbreak-black: Booting from a remotely unlocked pool won't work without also doing this though. It's an essential step that should be mentioned somewhere easily found.

contrib/remote-ssh-build.bash
cbreak-black: This won't work as-is, since `${RS_DIR}` is not usually the current working directory. I either have to ignore `${RS_DIR}`, and dump some of the configuration into the current working directory (the default build directory), or maybe change the build directory to `$RS_DIR`.

contrib/remote-ssh-build.bash
cbreak-black: I think the fact that we use dropbear inside might not be super relevant for users, but that the directory contains the keys is important.

contrib/remote-ssh-build.bash
cbreak-black: I've added this functionality, but somehow dropbear doesn't like ed25519. Maybe in the future.

docs/REMOTE-SSH.md
cbreak-black: I moved the whole readme into the script.

ahesford:
Thanks for making these changes. I did some basic tests and the image produced by your script works as expected. I squashed the commits and also made two minor changes (to silence shellcheck and allow the zbm-builder script to be found somewhere other than the PWD).

If you are interested, it would be nice to update my remote access wiki page to document using this script.

PR #322, opened 2022-07-20
Include the OpenSSH client in recovery images

zdykstra:
No other viable packaged SSH clients on Void - so this one is it.


ahesford (approved):
Could always install Microsoft (R) Win32 Platform Services for UNIX (TM) and go with PuTTY, but I guess this is fine in a pinch.

ahesford:
I just checked the build artifacts for this PR and see `/usr/lib/libssh.2.so.1*` but no `/bin/ssh` or `/usr/bin/ssh`.

zdykstra:
I tested this locally by way of `testing/` - `/usr/sbin/ssh` makes it into the initramfs. The build artifacts are incomplete because the `openssh` package isn't available in the build container.

PR #338, opened 2022-09-27
Add recursion (-R)

ahesford:


zdykstra (approved):
I guess recursion really was that easy.

PR #339, opened 2022-09-27
Add a top-row header showing the currently selected page

zdykstra:
The `draw_page` function is a bit of a hack, it has to write directly to the controlling terminal. This has been tested successfully on both serial outputs and virtualized physical outputs. To avoid weirdness, escape sequences are used to write the string directly to the required rows/columns, overwriting the existing header. Since the headers are always the same length, this just changes the highlighted page. A line return is NOT added; `fzf` does the right thing and starts drawing correctly on the next row, at column 0. `fzf` has to be told that the screen is 1 row shorter than it actually is, since it'll by default run in full screen mode.

Also add a special character `^` to help key text, indicating that the line should be centered on the screen. 

image

### DIFF
--- a/zfsbootmenu/bin/zfsbootmenu
+++ b/zfsbootmenu/bin/zfsbootmenu
@@ -125,7 +125,6 @@ while true; do
 
     bootenv="$( draw_be "${BASE}/bootenvs" )"
     ret=$?
-    tput clear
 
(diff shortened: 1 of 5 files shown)


zdykstra:
This solution is unworkable with the linux console. Abandoning for now!

PR #340, opened 2022-09-28
Center key bind text on horizontal layout fzf screens

zdykstra:
`column_wrap` executes in ~15ms from master and ~19ms with these changes (under the flamechart profiler).

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-lib.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-lib.sh
@@ -38,7 +38,7 @@ csv_cat() {
 # returns: nothing
 
 column_wrap() {
-  local footer
+  local footer max lpad pad
   footer="${1}"
 
   if [ "${COLUMNS:-0}" -lt 80 ] || [ -z "${HAS_COLUMN}" ] ; then
@@ -53,9 +53,18 @@ column_wrap() {
     footer="$( echo -e "${footer}" | column -t -s ':' )"
   fi
 
+  # Find the number of characters in the longest line
+  max="$( echo -e "${footer}" | awk 'BEGIN{l=0} length>l {l=length}; END{print l}' )"
+
+  # remove an extra 3 - 1 for '^', 2 for the fzf gutter
+  lpad="$(( (COLUMNS - max - 3) / 2 ))"
+
+  [[ ${lpad} -gt 0 ]] && printf -v pad "%*s" "${lpad}" ''
+
   footer="${footer//\[/\\033\[0;32m\[}"
   footer="${footer//\]/\]\\033\[0m}"
-  echo -e "${footer}"
+
+  echo -e "${footer//^/${pad}}"
 }
 
 # arg1: Path to file with detected boot environments, 1 per line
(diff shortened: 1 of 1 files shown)


zfsbootmenu/lib/zfsbootmenu-lib.sh
```diff
@@ -53,9 +53,16 @@ column_wrap() {
     footer="$( echo -e "${footer}" | column -t -s ':' )"
   fi
 
+  # Find the number of characters in the longest line
+  max="$( echo -e "${footer}" | awk -v l=0 'length>l {l=length}; END{print l}' )"
```
ahesford, line 57:
```suggestion
  max="$( echo -e "${footer}" | awk 'BEGIN{l=0} length>l {l=length} END{print l}' )"
```

zfsbootmenu/lib/zfsbootmenu-lib.sh
```diff
@@ -53,9 +53,16 @@ column_wrap() {
...
+
+  # remove an extra 3 - 1 for '^', 2 for the fzf gutter
+  printf -v pad "%*s" $(( ( COLUMNS - max - 3 ) / 2 )) ''
```
ahesford, line 60:
```suggestion
  local lpad=$(( (COLUMNS - max - 3) / 2))"
  [[ lpad -gt 0 ]] && printf -v pad "%*s" "${lpad}''
```

zfsbootmenu/lib/zfsbootmenu-lib.sh
```diff
@@ -53,9 +53,16 @@ column_wrap() {
...
+
   footer="${footer//\[/\\033\[0;32m\[}"
   footer="${footer//\]/\]\\033\[0m}"
-  echo -e "${footer}"
+
+  echo -e "${footer//^/${pad}}"
```
ahesford, line 65: Could make this auto-detect to avoid the carets and do, *e.g.*,
```suggestion
  if [ -z "${pad}" ]; then
    echo -e "${footer}"
    return
  fi
  
  echo -e "${footer}" | while IFS=$'\n' read -r line; do echo -e "${pad}${line}"; done
```

zdykstra: I'm torn on this. I was originally thinking it should be opt-in, for the vertical menu screens (help browser, pool status) - but with reasonable padding detection, it could work with out a problem on those too.

zfsbootmenu/lib/zfsbootmenu-lib.sh
zdykstra: Good call to sanity check that

zfsbootmenu/lib/zfsbootmenu-lib.sh
ahesford: I was kind of hoping you'd work some flamegraph magic on the while-loop version and tell me it isn't more than trivially more expensive than the search-and-replace...

ahesford: Also, if it isn't more expensive to do the while loop, and we still want opt-in behavior, there is always `CENTER=1 column_wrap [...]`

zdykstra: >  I was kind of hoping you'd work some flamegraph magic on the while-loop version and tell me it isn't more than trivially more expensive than the search-and-replace...

How do I do that in `ish`?

ahesford: I think you first have to install qemu-kvm so you can emulate x86_64 in your x86 emulator on your ARM phone.

zdykstra: Replacing `^` with a pad string: https://somebits.link/u/p/7389d74285/flamechart.svg?s=column_wrap
Padding the string line-by-line: https://somebits.link/u/p/15a20f02f8/flamechart.svg?s=column_wrap

PR #345, opened 2022-10-26
Add support for !!, so ONLY the preferred pool is imported

zdykstra:
Extend the modifiers for `zbm.prefer` to include `!!`. The single `!` behavior is preserved, requiring the named pool to be imported. Once that import has completed successfully, no other imports are attempted. This has been minimally tested on a two pool test environment. There might be a nicer way to do the `!` / `!!` check, but this seems to work well enough.

### DIFF
--- a/docs/pod/zfsbootmenu.7.pod
+++ b/docs/pod/zfsbootmenu.7.pod
@@ -18,9 +18,25 @@ These options are set on the kernel command line when booting the initramfs or U
 
 When creating an initramfs or UEFI bundle, the I</etc/hostid> from the system is copied into the target. If this image will be used on another system with a different hostid, replace B<E<lt>hostidE<gt>> with the desired hostid, as an eight-digit hexadecimal number, to override the value contained within the image.
 
+=item B<zbm.prefer>
+
+ZFSBootMenu will attempt to import as many pools as possible to identify boot environments and will, by default, look for the I<bootfs> property on the first imported pool (sorted alphabetically) to select the default boot environment. This option controls this behavior.
+
+=over 2
+
 =item B<zbm.prefer=E<lt>poolE<gt>>
 
-ZFSBootMenu will attempt to import as many pools as possible to identify boot environments and will, by default, look for the I<bootfs> property on the first imported pool (sorted alphabetically) to select the default boot environment. If you have multiple pools, replace B<E<lt>poolE<gt>> with the name of your preferred pool to override the default. If B<E<lt>poolE<gt>> is the name of a pool followed by a literal I<!> character, ZFSBootMenu will insist on successfully importing the named pool before attempting to import any others.
+The simplest form attempts to import B<E<lt>poolE<gt>> before any other pool. The I<bootfs> value from this pool will control the default boot environment.
+
+=item B<zbm.prefer=E<lt>poolE<gt>I<!>>
+
+If a literal I<!> has been appended to the pool name, ZFSBootMenu will insist on successfully importing the named pool before attempting to import any others.
+
+=item B<zbm.prefer=E<lt>poolE<gt>I<!!>>
+
+If a literal I<!!> has been appended to the pool name, ZFSBootMenu will insist on successfully importing the named pool and no others. 
+
+=back
 
 =item B<zbm.import_delay=E<lt>timeE<gt>>
 
--- a/zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh
+++ b/zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh
@@ -231,13 +231,23 @@ case "${root}" in
   ;;
 esac
 
-# Pool preference ending in ! indicates a hard requirement
-bpool="${root%\!}"
-if [ "${bpool}" != "${root}" ]; then
-  # shellcheck disable=SC2034
-  zbm_require_bpool=1
-  root="${bpool}"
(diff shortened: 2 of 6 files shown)


docs/pod/zfsbootmenu.7.pod
```diff
@@ -18,9 +18,25 @@ These options are set on the kernel command line when booting the initramfs or U
...
+=over 2
+
 =item B<zbm.prefer=E<lt>poolE<gt>>
 
-ZFSBootMenu will attempt to import as many pools as possible to identify boot environments and will, by default, look for the I<bootfs> property on the first imported pool (sorted alphabetically) to select the default boot environment. If you have multiple pools, replace B<E<lt>poolE<gt>> with the name of your preferred pool to override the default. If B<E<lt>poolE<gt>> is the name of a pool followed by a literal I<!> character, ZFSBootMenu will insist on successfully importing the named pool before attempting to import any others.
+The simplest form attempts to import B<E<lt>poolE<gt>> before any other pool. The I<bootfs> value from this pool will control the default boot environment.
+
+=item B<zbm.prefer=E<lt>poolE<gt>I<!>>
+
+If a literal I<!> has been appended to the pool name, ZFSBootMenu will insist on successfully importing the named pool before attempting to import any others
```
ahesford, line 33: Missing period at the end of this sentence

zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh
```diff
@@ -231,11 +231,18 @@ case "${root}" in
...
+
+bpool="${root//[\!]/}"
+if [ "${root:(-1)}" = "!" ]; then
   # shellcheck disable=SC2034
   zbm_require_bpool=1
+
+  if [ "${root:(-2):1}" = "!" ]; then
+    # shellcheck disable=SC2034
+    zbm_only_bpool=1
+  fi
```
ahesford, line 245: What about (it seems "!" doesn't need to be escaped in these contexts)
```sh
# pool! : this pool must be imported before all others
# pool!!: this pool, and no others, must be imported

case "${root}" in
  *!!)
    zbm_require_bpool=yes
    root="${root%!!}"
    ;;
  *!)
    zbm_require_bpool=only
    root="${root%!}"
    ;;
  *)
    zbm_require_bpool=""
    ;;
esac
```
...and corresponding changes to preinit and init.

zfsbootmenu/hook/zfsbootmenu-preinit.sh
```diff
@@ -29,6 +29,7 @@ export menu_timeout="${menu_timeout}"
 export loglevel="${loglevel}"
 export root="${root}"
 export zbm_require_bpool="${zbm_require_bpool}"
+export zbm_only_bpool="${zbm_only_bpool}"
```
ahesford, line 32: Not necessary if `zbm_require_bpool` is one of empty, `yes` or `only`.

zfsbootmenu/libexec/zfsbootmenu-init
```diff
@@ -105,7 +105,10 @@ while true; do
 
   # shellcheck disable=SC2154
   if check_for_pools; then
-    if [ -n "${try_pool}" ]; then
+    if [ "${zbm_only_bpool:-0}" -eq 1 ]; then
```
ahesford, line 108:
```suggestion
    if [ "${zbm_only_bpool}" = "only" ]; then
```
and also on line 133 in this diff ([line 130 in the original](https://github.com/zbm-dev/zfsbootmenu/blob/master/zfsbootmenu/libexec/zfsbootmenu-init#L130)) would replace
```sh
[ "${zbm_require_bpool:-0}" -ne 1]
```
with
```sh
[ -z "${zbm_require_bpool}" ]

zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh
zdykstra: Agreed, that's a lot easier to quickly read and understand what is intended.

docs/pod/zfsbootmenu.7.pod
zdykstra: fixed

zdykstra:
Testing was performed in an Alpine VM with two zpools, `ztest` and `ztest-01`.

* Booting with `zbm.prefer=ztest-01`
Expected: both pools will be imported, ZFSBootMenu will default to the environment specified by the value of `bootfs` on `ztest-01`.
Result: as expected

* Booting with `zbm.prefer=ztest-01!`
Expected: both pools will be imported, ZFSBootMenu will default to the environment specified by the value of `bootfs` on `ztest-01`.
Result: as expected

* Booting with `zbm.prefer=ztest-01!!`
Expected: both pools will be imported, ZFSBootMenu will default to the environment specified by the value of `bootfs` on `ztest-01`.
Result: as expected

* Booting with `zbm.prefer=ztest-asdf!`
Expected: ZFSBootMenu will enter a try/catch loop attempting to import `ztest-asdf`, allowing access to the recovery shell after the counter runs out.
Result: as expected

* Booting with `zbm.prefer=ztest-asdf!!`
Expected: ZFSBootMenu will enter a try/catch loop attempting to import `ztest-asdf`, allowing access to the recovery shell after the counter runs out.
Result: as expected

ahesford (approved):

PR #346, opened 2022-10-27
Allow imports of hooks from external sources at runtime

ahesford:
Setting `zbm.hookroot=<filesystem>//<path>` will cause ZBM to mount the named filesystem and copy any hooks (files) in the path

    <path>/{early-setup,setup,teardown}.d

into the corresponding hook directory under `/libexec`.

I ran through some tests and this seems to do what I expect; adding a hook root at `/EFI/zfsbootmenu/hooky` and some hook subdirectories with test hooks to my EFI system partition and specifying `zbm.hookroot=UUID=XXXX-YYYY//EFI/zfsbootmenu/hooky` causes the hooks to be copied into the expected locations in the initramfs at boot. The hooks are then executed by the normal flow. Failures may log messages but otherwise do not impede operation.

One caveat is that any modules necessary to mount the hook filesystem must be available for loading. I learned this the hard way with my `mkinitcpio` image, which lacked the `vfat.ko`, `fat.ko`, `nls_iso8859_1.ko` and `nls_cp437.ko` modules. Adding these modules explicitly to the `MODULES` array in `mkinitcpio.conf` resolved the issue.

If we're going to ship prebuilt EFI executables that can load external hooks, we'll need to be sure that dracut includes support for vfat and any other common filesystems we wish to support.


zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1815,3 +1815,49 @@ zreport() {
...
+    zerror "failed to mount hook filesystem ${hook_fs}"
+    return 1
+  fi
+
+  for hdir in early-setup.d setup.d teardown.d; do
+    hsrc="${hook_mount}/${hook_path}/${hdir}"
+    [ -d "${hsrc}" ] || continue
+    mkdir -p "/libexec/${hdir}"
+    for hfile in "${hsrc}"/*; do
+      [ -f "${hfile}" ] && cp "${hfile}" "/libexec/${hdir}"
```
zdykstra, line 1856: Is it worth handling a failure to copy, or a user-provided hook having the same name as one we ship by default?

zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh
```diff
@@ -193,6 +193,13 @@ else
   zinfo "defaulting sort key order to ${zbm_sort}"
 fi
 
+# Allow hooks to be imported from another filesystem
+
+# shellcheck disable=SC2034
+if ! zbm_hook_root=$( get_zbm_arg zbm.hookroot ) ; then
```
zdykstra, line 199: Can this be changed to more closely match the existing `get_zbm_arg` usage pattern?
```
zbm_hook_root=
if zbm_hook_root=$( get_zbm_arg zbm.hookroot) ; then
  zinfo "setting user hooks to ${zbm_hook_root}"
fi
```

zdykstra (commented):
This looks good, and should prove quite useful for people that are only rebuilding just so that they can add in custom hooks. A documentation bump will be needed - I can handle this if you'd like.

zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh
ahesford: Done

zfsbootmenu/lib/zfsbootmenu-core.sh
ahesford: I now trap failures to copy and print a warning. Allowing hooks with the same name as defaults is a nice way to allow users to mask the default hooks.

zdykstra (approved):

ahesford:
7eb77804cbb

PR #348, opened 2022-10-28
Move boot environments to ${BASE}/environments/

zdykstra:
It's only a matter of time until a pool name conflicts with a file that we create in `/zfsbootmenu`. Moving the pool/dataset directory layout to `/zfsbootmenu/environments` puts them in a directory free of potential conflicts. A helper function, `be_location` has been added that returns the on-disk path for where a filesystem _should_ be mounted/live. This avoids having to manually construct this string when `mount_zfs` isn't in use.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -240,7 +240,7 @@ mount_zfs() {
     return 1
   fi
 
-  mnt="${BASE}/${fs}/mnt"
+  mnt="$( be_location "${fs}" )/mnt"
   mkdir -p "${mnt}"
 
   # filesystems are readonly by default, but read-write mounts may be requested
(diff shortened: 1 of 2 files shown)


zdykstra:
This has been minimally tested - it's possible that I've missed some uses where `${BASE}/${fs}/...` was crafted directly, but a quick and dirty grep doesn't show anything.

zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1743,6 +1745,22 @@ load_key() {
...
+be_location() {
+  local fs
+  fs="${1}"
+  if [ -z "${fs}" ]; then
+    zerror "fs is undefined"
+    return 1
+  fi
+  zdebug "fs set to ${fs}"
+
+  echo "${BASE}/environments/${fs}"
```
ahesford, line 1761: Could do
```suggestion
  local beloc="${BASE}/environments/${fs}"
  mkdir -p "${beloc}"
  echo "${beloc}"
```
to make sure the path always exists when a path is returned.

ahesford (approved):
I haven't done an extensive search for lingering paths, but this seems like a good idea.

zfsbootmenu/lib/zfsbootmenu-core.sh
zdykstra: Good idea. Done!

Issue #335, opened 2022-09-10
`//lib/dracut/hooks/cmdline/95-zfsbootmenu-parse-commandline.sh: Syntax error: "(" unexpected` when booting


lediur:
I recently set up a Debian system with zfsbootmenu and rEFInd using the instructions in the wiki. After a few days, I updated the system from bullseye with zfs 2.0 to testing with zfs 2.1.5. The next time I booted, zfsbootmenu presented a warning and said I should rerun `generate-zbm`, but still allowed me to boot into Debian. In my Debian testing install, I reran `generate-zbm` and rebooted.

Now, if I try to use rEFInd to boot the most recently generated entry, the boot process fails with the error:

```
[ ... ] Run /init as init process 
/init: 5: //lib/dracut/hooks/cmdline/95-zfsbootmenu-parse-commandline.sh: Syntax error: "(" unexpected
dracut Warning: Signal caught!
```

![signal-2022-09-09-191305](https://user-images.githubusercontent.com/1077648/189464979-18b291ad-fd04-419d-ace8-965855b7a129.jpeg)

It drops to the emergency shell, but the shell is unresponsive to USB keyboard input.

I've tried adding `init=/bin/bash` to the kernel arguments with no change.

I'm not very familiar with dracut or the surrounding boot process, but I did a bit of digging nevertheless. The error seems to suggest that the script isn't getting run as a bash script, but by `sh` or `dash`? This appears to be [line 5 of the affected file](https://github.com/zbm-dev/zfsbootmenu/blob/master/zfsbootmenu/hook/zfsbootmenu-parse-commandline.sh#L5).

Any idea how I can recover my ZBM setup? I've tried fiddling a bit with a chrooted live USB, but I'm not really sure what I need to fix in order to execute this script properly short of rewriting the hook to not use Bash arrays and rebuilding.

zdykstra:
When you're on the `rEFInd` screen and you hit `F2`, do you have access to previous versions of ZFSBootMenu? If you do, you can simply boot one of those for the time being. The warning to upgrade your version of ZFS in ZFSBootMenu isn't fatal; it'll just prevent you from doing any read-write operations inside ZBM. You can continue to boot normally.

[...]

PR #349, opened 2022-10-29
Explicitly require bash, blacklist dash

ahesford:
I don't know what broke the Debian and Ubuntu installations in https://github.com/zbm-dev/zfsbootmenu/issues/335 and I can't reproduce the issue in our testbed, even when I bump Ubuntu from "jammy" to "kinetic". However, clearly something has caused the affected images to link `/bin/sh` to dash. This breaks our KCL parsing, which is done in a sourced dracut hook before we explicitly exec into bash and take control.

The dracut `bash` module will force installation of `/bin/bash` and, if `/bin/sh` isn't linked, link to bash. Adding `bash` to the required dependencies of the ZBM dracut module seems like a no-brainer. The `bash` module was added to dracut in 2013 so we should not have any issues.

When I spun up a testbed, I noticed that the initramfs generated by my system also included the `dash` module. Like `bash`, this module installs dash and, if `/bin/sh` isn't linked, link to dash. This creates the possibilityh of an ordering problem where `dash` is processed first; they are currently named `00bash` and `00dash` so lexicographical ordering should ensure `bash` is processed first. However, I don't know if there are other means for sorting and processing modules that could cause problems.

In response, I have blacklisted the `dash` module. I don't know if there are adverse consequences to this and am on the fence. Maybe we just hard depend on `bash` and wait for additional reports of broken systems to take more aggressive action.

Another alternative is to defer ZBM KCL parsing until we exec into our own preinit. This probably doesn't change anything from our perspective and means we don't have to care about whether dash is `/bin/sh` or not.

Thoughts?

### DIFF
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -7,7 +7,7 @@ check() {
 }
 
 depends() {
-  echo udev-rules
+  echo bash udev-rules
   return 0
 }
 
--- a/etc/zfsbootmenu/dracut.conf.d/zfsbootmenu.conf
+++ b/etc/zfsbootmenu/dracut.conf.d/zfsbootmenu.conf
@@ -1,3 +1,3 @@
 nofsck="yes"
 add_dracutmodules+=" zfsbootmenu "
-omit_dracutmodules+=" btrfs zfs resume systemd systemd-initrd dracut-systemd plymouth "
+omit_dracutmodules+=" btrfs zfs resume systemd systemd-initrd dracut-systemd plymouth dash "


Sithuk:
I just ran a test with omit_dracutmodules+=" dash " in /etc/zfsbootmenu/dracut.conf.d/zfsbootmenu.conf

Zfsbootmenu failed to start. I was dropped to the emergency shell. I could see that dash and bash are both present in /usr/bin. /bin/sh is symlinked to /bin/dash. It doesn't look like using omit_dracutmodules succeeded?

ahesford:
In my tests, omitting `dash` caused it to be omitted from the initramfs. Something is wrong with your setup. Did you also carry forward the hard bash dependency I added to `module-setup.sh`? That should have the same effect as your earlier successful attempt to `add_dracutmodules+=" bash "`.

Can you grep through *all* of your dracut configs---`/etc/dracut*`, `/etc/zfsbootmenu/dracut*`, `/usr/lib/dracut*` to see what might be pulling in the `dash` module? Also confirm that your `config.yaml` refers to the right dracut configuration directory.

Sithuk:
Sorry, I missed the change to include bash in the /usr/lib/dracut/modules.d/90zfsbootmenu/module-setup.sh file. I've done that change too. Then re-ran "update-initramfs -c -k all" and "generate-zbm". Everything works fine. Please ignore my previous comment.

zdykstra:
I'm okay with blacklisting `dash`. Bash, when invoked as `/bin/sh` emulates a POSIX compliant `sh`. With `bash` as the only shell module pulled in for Dracut, `/bin/sh` is a symlink to `bash`.

Moving all KCL parsing to our pre-init script and making `zfsbootmenu-parse-commandline.sh` be a super minimal stub could be useful. I think the only thing that has to be set there is `rootok=1` and then a symlink from `/dev/null` to /dev/root`. Before we do this, though, it might be a good idea to dig into the structure of Debian's initramfs generator works to make sure we aren't backing ourselves into a corner. Though again, we generally just need to `exec` into our codebase and throw away anything else the underlying initramfs scripts/tooling are doing.

zdykstra (approved):

Sithuk:
It would be good to know what changed too. What did Debian change that caused this issue? The 95-zfsbootmenu-parse-commandline.sh script has #!/bin/bash at the top. Bash is present in /bin/ in the initramfs. It should run using bash not, /bin/sh ? So it shouldn't matter that /bin/sh is a link to dash?

zdykstra:
The parse command line script is sourced by the Dracut event loop, not executed as a stand-alone script. The interpreter that manages to execute `/init` is what will be used to execute the ZBM command line parsing script.

The interpreter shebang at the top is largely there to provide hints to developers/tools that it's Bash and not POSIX sh.

ahesford:
I forgot about the rootok stuff. Let's defer restructuring known control flow and just force the dracut environment to behave the way we expect.

PR #351, opened 2022-11-01
generate-zbm: deprecate syslinux support

ahesford:
Support for syslinux is poorly tested and, in fact, still doesn't work properly. When `Components.Versions` is set to `false`, `generate-zbm` will properly roll `vmlinuz-bootmenu`, `initramfs-bootmenu.img` as well the backup files `vmlinux-bootmenu-backup` and `initramfs-bootmenu-backup.img`. However, the syslinux configuration generator doesn't build the proper backup entries. Although the backup entry has `KERNEL` line that points to `vmlinuz-bootmenu-backup` as expected, the `LABEL` for this entry is still `ZFSBootMenu-bootmenu` (the same as the entry for the newer `vmlinuz-bootmenu` entry) and, even worse, the `INITRD` points to `initramfs-bootmenu.img` in *both* entries. Not only will the entry *not* boot the known-good backup version, it is likely not to boot at all, since the backup kernel and newer initramfs may well not have matching kernel and kmod versions.

The assumption that `generate-zbm` is the sole owner of the syslinux configuration is also bad, as we see on IRC and, indeed in my own use: I require a `SERIAL` declaration that is not added by `generate-zbm`.

My preferred approach is to disable syslinux support and set `Components.Versions = false` to give me a current kernel/initramfs pair and a single `-backup`, then maintain my own static syslinux configuration that references each of these. I suggest we make this the official position and just drop syslinux support altogether.

If this is merged, I will update the wiki page to document the new procedure and we can keep the legacy support around for one more release cycle.

### DIFF
--- a/contrib/syslinux-update.sh
+++ b/contrib/syslinux-update.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+
+## This script can be used by generate-zbm to update a syslinux configuration
+## after a ZFSBootMenu initramfs and kernel pair is created. The file system
+## from which syslinux will run, which can be ext2/3/4 (make sure to disable
+## the `64bit` feature), NTFS, UFS/FFS, XFS or btrfs, must be mounted when this
+## hook runs. To ensure that syslinux can find the ZFSBootMenu kernel and
+## initramfs, make sure to set the `Components.ImageDir` option in your
+## ZFSBootMenu configuration to a path within the syslinux file system.
+##
+## Ensure the file is executable and place it in the directory specified by the
+## `Global.PostHooksDir` option in your ZFSBootMenu configuration file.
+##
+## This script uses the SYSLINUX_ROOT to refer to the mount point of the
+## syslinux filesystem and KERNEL_PATH to be the subdirectory, relative to
+## SYSLINUX_ROOT, where ZFSBootMenu will install initramfs and kernel pairs.
+## Kernels are expected to have file names of the form
+##
+##     ${KERNEL_PREFIX}[-<version>]
+##
+## and a corresponding initramfs image
+##
+##     initramfs[-<version>].img
+##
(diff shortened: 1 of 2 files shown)


ahesford:
After the deprecation cycle, we will have to decide between silently ignoring any `syslinux` configuration options or throwing a hard error to force people to address the configuration change.

contrib/syslinux-update.sh
```diff
@@ -0,0 +1,116 @@
...
+		DEFAULT_SET="yes"
+	fi
+
+	cat >> "${SYSLINUX_CFG}" <<-EOF
+
+	LABEL ${zbmlabel}
+	MENU LABEL ZFSBootMenu (${version#-})
+	LINUX ${KERNEL_PATH:+/${KERNEL_PATH}}/${kern}
+	INITRD ${KERNEL_PATH:+/${KERNEL_PATH}}/${initramfs}
+	APPEND ${ZBM_KCL_ARGUMENTS}
```
zdykstra, line 103: This doesn't look like it's defined anywhere / is superseded by `ZBM_KCL_ARGS`

zdykstra (changes requested):
Moving `syslinux.cfg` to a contrib script is the right thing to do. `generate-zbm` should only be focused on creating images. We don't have integrations with `efibootmgr` baked into `generate-zbm`, so we shouldn't have `syslinux` support baked in either.

I'd like to see a slight extension to the contrib script - have it blindly dump the contents of files from a `SYSLINUX_SNIPPET` directory into the final configuration file. They'd be inserted in the order that a general file glob found them, with no attempt made to validate them / verify them.

contrib/syslinux-update.sh
ahesford: Thanks, fixed

zdykstra (approved):

PR #356, opened 2022-11-13
Use fzf border labels

zdykstra:
`fzf` 0.35 introduced border/preview labels. Global (e.g. top row) text can now be set for any invocation of `fzf`. This means that we can now have a header that displays which screen you're on when in the main `zfsbootmenu` interface. The recovery shell and zdebug/trace screen, and any screens requiring custom input (cloning a snapshot, pool rollback, etc) do not have a header.

image

Included in this WIP are the following changes:

* Test if the border label option exists, gate it behind `HAS_BORDER_LABEL`
* Add a `draw_page`  (would `global_header` make more sense?) function that determines which page to highlight based on the name of the caller. Easy to extend / keep all of the logic central to one place.
* Changes the preview border from rounded to sharp, and then sets it to white. This now makes the preview border visible on the Linux console, instead of being invisible / lost space.

I believe I've accounted for every usage of fzf and preview windows. The colors used for the page text could possibly be tweaked - both the selected page and the unselected page colors. The start/end of the page string could maybe use some gussying up.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-lib.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-lib.sh
@@ -67,6 +67,53 @@ column_wrap() {
   echo -e "${footer//^/${pad}}"
 }
 
+# arg1: optional page to mark as active
+# prints: a header string with the active page highlighted yellow
+# returns: nothing
+
+# shellcheck disable=SC2120
+global_header() {
+  local header page tab
+
+  # Accept a parameter to explicitly set a page to be shown
+  if [ -n "${1}" ]; then
+    page="${1}"
+  else
+    # Name of the calling function
+    page="${FUNCNAME[1]}"
+  fi
+
+  # Set the entire string to one color
+  header="\\033[0;37m Boot Environments | Snapshots | Kernels | Pool Status | Logs | Help \\033[0m"
+
+  case "${page}" in
+    draw_be)
+      tab="Boot Environments"
+      ;;
+    draw_kernel)
+      tab="Kernels"
+      ;;
+    draw_snapshots)
+      tab="Snapshots"
+      ;;
+    draw_pool_status)
+      tab="Pool Status"
+      ;;
+    help_pager)
+      tab="Help"
+      ;;
+    zlogtail)
+      tab="Logs"
+      ;;
+    *)
+      zdebug "Called from unknown function: ${page}"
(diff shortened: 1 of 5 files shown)


zdykstra:
Updated to use preview window labels on the snapshot screen.
image

zfsbootmenu/lib/zfsbootmenu-lib.sh
```diff
@@ -206,14 +253,24 @@ draw_snapshots() {
 
   zdebug "snapshots: ${snapshots[*]}"
 
-  if ! selected="$( HELP_SECTION=snapshot-management ${FUZZYSEL} \
+  local LEGACY_CONTEXT
+  if [ -z "${HAS_BORDER_LABEL}" ]; then
+    LEGACY_CONTEXT=1
+  fi
```
ahesford, line 259: Maybe call this `PREVIEW_PAD` or something more obvious, and set it to 0 or 1 explicitly to avoid the `${LEGACY_CONTEXT:-0}` substitution below.

zfsbootmenu/lib/zfsbootmenu-lib.sh
```diff
@@ -67,6 +67,51 @@ column_wrap() {
   echo -e "${footer//^/${pad}}"
 }
 
+# args: optional page to mark as active
+# prints: writes a string directly to the controlling terminal
+# returns: nothing
+
+# shellcheck disable=SC2120
+draw_page() {
```
ahesford, line 75: Would `tab_header` or `page_header` be more descriptive than `draw_page`?

ahesford (approved):

zfsbootmenu/lib/zfsbootmenu-lib.sh
zdykstra: I changed it to `global_header` since it no longer actually draws directly to the screen.

zfsbootmenu/lib/zfsbootmenu-lib.sh
zdykstra: `LEGACY_CONTEXT` does double duty here. It's a flag to indicate that an optional parameter should be passed to `zfsbootmenu-preview`, and it's abused to expand the preview window height by one.  It's not a very pretty construct, so maybe a comment indicating what it's doing is in order?

PR #360, opened 2022-11-15
Fix draw_diff not exiting correctly

zdykstra:
`draw_diff` needs to be able to trap SIGINT to gracefully handle exiting when `zfs diff` blocks. It was mildly cleaned up to reflect that it's now a full-fledged script instead of a function call.

This highlighted that `global_header` was a bit sloppy in how it needed to detect the caller. I've switched it to use the basename of the calling script if the function is `main`.

### DIFF
--- a/zfsbootmenu/libexec/zfsbootmenu-diff
+++ b/zfsbootmenu/libexec/zfsbootmenu-diff
@@ -1,0 +1,76 @@
+#!/bin/bash
+# vim: softtabstop=2 shiftwidth=2 expandtab
+
+# shellcheck disable=SC1091
+source /lib/profiling-lib.sh >/dev/null 2>&1 || true
+source /etc/zfsbootmenu.conf >/dev/null 2>&1 || exit 1
+source /lib/kmsg-log-lib.sh >/dev/null 2>&1 || exit 1
+source /lib/zfsbootmenu-core.sh >/dev/null 2>&1 || exit 1
+source /lib/zfsbootmenu-lib.sh >/dev/null 2>&1 || exit 1
+
+# prevent ctrl-c from killing us, so that zfs diff can exit cleanly
+trap '' SIGINT
+
+snapshot="${1}"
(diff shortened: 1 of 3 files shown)


zfsbootmenu/libexec/zfsbootmenu-diff
```diff
@@ -0,0 +1,75 @@
...
+# Bash won't use an FD referenced in a variable on the left side of a pipe
+exec 3>&"${zfs_diff[0]}"
+
+# shellcheck disable=SC2154
+line_one="$( center_string "---${snapshot}" )"
+left_pad="${line_one//---${snapshot}/}"
+line_one="$( colorize red "${line_one}" )"
+line_two="${left_pad}$( colorize green "+++${diff_target}" )"
+
+sed "s,${mnt},," <&3 | HELP_SECTION=diff-viewer ${FUZZYSEL} --prompt "> " \
```
ahesford, line 68:
```suggestion
sed "s,${mnt},," <&3 | HELP_SECTION=diff-viewer ${FUZZYSEL:-fzf} --prompt "> " \
```

ahesford (approved):

zfsbootmenu/libexec/zfsbootmenu-diff
zdykstra: done!

zfsbootmenu/libexec/zfsbootmenu-diff
```diff
@@ -0,0 +1,75 @@
...
+  ${HAS_BORDER_LABEL:+--border-label="$( global_header )"} \
+  --preview="echo -e '${line_one}\n${line_two}'" --no-sort \
+  --preview-window="up:${PREVIEW_HEIGHT},border-sharp"
+
+[ -n "${zfs_diff_PID}" ] && kill "${zfs_diff_PID}"
```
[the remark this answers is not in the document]
zdykstra: done!

PR #361, opened 2022-11-23
Improve containerized builds

ahesford:
- The in-container builder, `zbm-build.sh`, links the in-repo default `mkinitcpio.conf` into `/etc/zfsbootmenu` so mkinitcpio builds work

- If the in-container builder finds a custom `mkinitcpio.conf` in the build directory, that overrides the default configuration

- The in-container builder now looks for an `rc.d` subdirectory of the build directory and will invoke every executable file therein before calling `generate-zbm` to "terraform" the build container.

- The `zbm-builder.sh` convenience wrapper now supports a configuration file to allow defaults to be specified. This requires a two-pass getopts to find and load the configuration file before parsing remaining options.

- A new option, `-R`, will remove any existing `hostid` and `zpool.cache` files from the build directory to make sure they are always up to date with the host versions.

The changes herein allow me to build a custom ZFSBootMenu image using mkinitcpio with dropbear and a self-contained configuration/build directory. I will need to document this by adding a wiki page or doc file and possibly creating a skeleton that can be installed with make.


zdykstra (approved):
Nothing jumps out at me here. I'll look forward to seeing the documentation so that the `rc.d` changes make more sense to me.

ahesford:
7a83ae13eba65fd2d02f0c8f3d325e71308224e3

PR #375, opened 2022-12-05
testing: Allow environment customization, better choose initramfs generators

ahesford:
Adding a `-E` flag to `testing/setup.sh` allows us to pass environment variables to the image build process, which means we can parameterize releases and do all sorts of other good stuff. I tested a setup with

```sh
./setup.sh -a -o debian -E RELEASE=buster -D /tmp/debs
./setup.sh -i -x -o ubuntu -E RELEASE=kinetic -D /tmp/debs
```

and was greeted with a shockingly old Debian installation followed by a newer Ubuntu than we install by default.

While I was at it, I rethought how we choose initramfs generators, meaning things work as expected if `dracut` isn't locally installed.

### DIFF
--- a/testing/helpers/chroot-debian.sh
+++ b/testing/helpers/chroot-debian.sh
@@ -1,8 +1,11 @@
 #!/bin/bash
 
+: "${RELEASE:=bullseye}"
+: "${APT_REPOS:=main contrib}"
+
 cat << EOF > /etc/apt/sources.list
-deb http://deb.debian.org/debian bullseye main contrib
-deb-src http://deb.debian.org/debian bullseye main contrib
+deb http://deb.debian.org/debian ${RELEASE} ${APT_REPOS}
+deb-src http://deb.debian.org/debian ${RELEASE} ${APT_REPOS}
 EOF
 
 apt-get update
--- a/testing/run.sh
+++ b/testing/run.sh
@@ -71,10 +71,6 @@ else
   done
 fi
 
-if ! [ -d "${TESTDIR}/dracut.conf.d" ] ; then
-  error "test directory '${TESTDIR}' is incomplete"
-fi
-
 # Support x86_64 and ppc64(le)
 case "$(uname -m)" in
   ppc64*)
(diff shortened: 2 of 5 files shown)


testing/run.sh
```diff
@@ -197,13 +192,36 @@ while getopts "${CMDOPTS}" opt; do
...
-: > "${TESTDIR}/dracut.conf.d/testing.conf"
-: > "${TESTDIR}/mkinitcpio.d/testing.conf"
+# Enable logging for dracut and verify the configuration exists
+if ((DRACUT)); then
+  if [ ! -d "${TESTDIR}/dracut.conf.d" ]; then
+    echo "ERROR: cannot use dracut without dracut.conf.d"
+    exit 1
+  fi
+
+  AAPPEND+=( "rd.log=kmsg" )
```
zdykstra, line 214: Despite the name of `rd.log`, this is actually a `initcpio` [framework parameter.](https://github.com/archlinux/mkinitcpio/blob/master/init_functions#L93-L115)

testing/helpers/chroot-debian.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 
+: "${RELEASE:=bullseye}"
```
zdykstra, line 3: Can the `RELEASE` and possibly the other environment variables be added to the help output of `setup.sh`? We won't be able to enumerate through the various values / show default values, but at least the variable names + description of what they do can be documented.

zdykstra (changes requested):
This is a nice set of refinements to tooling that was largely developed ad-hoc over the last few years. It sets us up nicely for `initramfs-tools` support.

testing/run.sh
ahesford: Not sure how I misplaced this, but I relocated it.

testing/helpers/chroot-debian.sh
ahesford: Done.

zdykstra (approved):

PR #378, opened 2022-12-14
Pool compatibility support in testing / built images

zdykstra:
Expand pool compatibility support in the testing framework. By default, `Debian` and `Ubuntu` test environments will use a pool compatibility of `openzfs-2.0-linux`, which lines up with the version available in the latest releases. The `POOL_COMPAT` environment variable can be manually set (or overridden in the case of the previous distributions) to any specific value.

Dracut and initcpio now also copy these files in to the ZBM image so that they can be read/used in the recovery shell.

### DIFF
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -95,6 +95,11 @@ install() {
     inst_simple "${doc}" "/usr/share/docs/${relative}"
   done <<<"$( find "${zfsbootmenu_module_root}/help-files" -type f )"
 
+  compat_dirs=( "/etc/zfs/compatibility.d" "/usr/share/zfs/compatibility.d/" )
+  for compat_dir in "${compat_dirs[@]}"; do
+    # shellcheck disable=2164
+    [ -d "${compat_dir}" ] && tar -cf - "${compat_dir}" | ( cd "${initdir}" ; tar xfp - )
+  done
   _ret=0
 
   # Core ZFSBootMenu functionality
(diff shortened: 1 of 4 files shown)


zdykstra:
Damn, this isn't copying symlinks - and quite a few compat files are symlinked to others.

dracut/module-setup.sh
```diff
@@ -95,6 +95,11 @@ install() {
     inst_simple "${doc}" "/usr/share/docs/${relative}"
   done <<<"$( find "${zfsbootmenu_module_root}/help-files" -type f )"
 
+  compat_dirs=( "/etc/zfs/compatibility.d" "/usr/share/zfs/compatibility.d/" )
+  for compat_dir in "${compat_dirs[@]}"; do
+    # shellcheck disable=2164
+    [ -d "${compat_dir}" ] && tar -cf - "${compat_dir}" | ( cd "${initdir}" ; tar xfp - )
```
zdykstra, line 101: I  couldn't find a `dracut` function that did this properly. Since we're just copying a few text files, a tar pipeline is sufficient for this.

testing/setup.sh
```diff
@@ -322,6 +326,26 @@ if ((IMAGE)); then
       IMAGE_SCRIPT="./helpers/image.sh"
     fi
 
+    AUTO_POOL_COMPAT=1
+    for environ in "${ENVIRONS[@]}"; do
+      if [[ "${environ}" =~ "POOL_COMPAT" ]]; then
```
ahesford, line 331: I'd either anchor this as `"^POOL_COMPAT="` or use a case block matching `POOL_COMPAT=*`

ahesford (approved):

PR #379, opened 2022-12-20
Key management: properly normalize Booleans, make cache mutex per-PID

ahesford:
The cache mutex exists to avoid infinite recursion when attempting to load a key for an encryption root from its key source. However, the recursion is only a concern when the load is done in the same PID that is trying to do the caching. In other cases, such as when a remote user has terminated a local instance and is not attempting to unlock a pool, a generic mutex will prevent all key caching. Making the mutex per-PID avoids this conflict.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -1547,7 +1547,7 @@ cache_key() {
   fi
 
   keycache="${BASE}/.keys/${keysrc}"
-  mutex="${keycache}/.cachemutex"
+  mutex="${keycache}/.cachemutex.$$"
 
   if [ -e "${mutex}" ]; then
     # Attempting to load key source could lead to infinite recursion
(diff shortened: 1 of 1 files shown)


zdykstra:
I spun up a simple encrypted Void test environment. Using this PR, `cache_key` is never called. Boot the VM from `master` works as expected; the keyfile is cached.

ahesford:
I believe that was caused by nested calls to `load_key`, where the first one set `NO_CACHE=0` and then later checked `[ -n "${NO_CACHE}" ]` and forced it high. I reworked the variable noramlization a bit, although I suppose the original form would work just fine.

zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1667,9 +1667,16 @@ load_key() {
     return 0
   fi
 
-  # Default to 0 when unset
-  [ -n "${CLEAR_SCREEN}" ] || CLEAR_SCREEN=0
```
zdykstra, line 1671: What wasn't working with this logic? It seemed to be fine as-is - there's more complexity with the new checks - is that solving a specific problem?

ahesford: No, this is just conflating a couple of things. The only thing that matters for the key caching problem is limiting the scope of the mutex to the current PID.

The changes to CLEAR_SCREEN and NO_CACHE were just an attempt to normalize the value and also avoid potential (but nonexistent) issues with numeric comparisons.

ahesford:
I pared the boolean handling changes way back. I still want to go from `$NO_CACHE -eq 0` to `$NO_CACHE -ne 1` beacuse `NO_CACHE` is an explit opt-in and it should only apply if the variable is exactly right. However, that still fails to default to caching if `$NO_CACHE` is not numeric, so `! [ $NO_CACHE -ne 0 ]` gets me where I want to go.

zdykstra (approved):

ahesford:
c578ef530df07f61c2aed341b2e1573b3b903fd4

PR #386, opened 2022-12-22
Drop the beginner's guide from the official docs; it belongs on the wiki

ahesford:
If the guide receives official review, we can consider moving it back to Read the Docs, but it does not belong there when the content still warns that the team has not reviewed the content.


zdykstra:
:+1:

PR #390, opened 2022-12-29
Use ::include in distro specific/agnostic ways

zdykstra:
Sweeping changes here, that accomplish the following broad strokes:

* Create `docs/guides/_includes` with files that are distribution-agnostic
* Create docs/guides/(debian|void-linux)/_includes` with files that are specific to that distribution
* Set common `replace::` entries in the guides as a rudimentary template system
* Rework the existing Void Linux UEFI guides to be almost exclusively composed of `.. include::` entries
* Further pattern the Debian guides after the Void UEFI guides, introducing a new detached UEFI guide

The Debian basic UEFI guide has been extensively tested. The guides could use a spot check against the copy in `master`. I'll be testing the Void UEFI guide(s) for accuracy.


zdykstra:
If the Debian UEFI guide location changes, I'll have to add a redirect on the RTD project page to fix up old links ... though there are probably very very few given how new this documentation is.

docs/guides/_include/zbm-setup.rst
```diff
@@ -2,11 +2,30 @@ Install and configure ZFSBootMenu
...
+
+      .. code-block::
+
+        zfs set org.zfsbootmenu:commandline="quiet" zroot/ROOT
+
+      Setup key caching in ZFSBootMenu.
+
+      .. parsed-literal::
+
+        zfs set org.zfsbootmenu:keysource="zroot/ROOT/\ |distribution|"
```
ahesford, line 22: This is missing the target encroot where the property will be set.

zdykstra: Whoops. I now have it set on `zroot`, which is where the encroot actually is configured for these guides.

PR #391, opened 2023-01-03
docs/guides/: change early EFI boot check

zdykstra:
Neither `hrmpf` or the Debian live CDs mount `efivarfs` by default, so `efibootmgr` fails as an early check. Checking `dmesg` for that string seems to be a reliable check to differentiate BIOS vs EFI booting.

### DIFF
--- a/docs/guides/_include/efi-boot-check.rst
+++ b/docs/guides/_include/efi-boot-check.rst
@@ -1,0 +1,6 @@
+Confirm EFI support:
+
+.. parsed-literal::
+
+  # **dmesg | grep -i efivars**
+  [    0.301784] Registered efivars operations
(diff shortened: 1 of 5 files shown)


ahesford (changes requested):
Can you make these `parsed-literal` blocks and bold the user input to be consistent with the other sections where user input and program output are mixed into a common block?

zdykstra:
> Can you make these `parsed-literal` blocks and bold the user input to be consistent with the other sections where user input and program output are mixed into a common block?

Moved to a distro-agnostic include and switched to `parsed-literal`. I still need to vet that `dmesg` works without sudo in `hrmpf`.

ahesford (approved):

PR #397, opened 2023-01-09
zfsbootmenu-core: add function to search for and mount /boot

ahesford:
In very rare circumstances, it may be desirable to separate `/boot` from its boot environment. For example, one may wish to expose an encrypted (but unlocked) root filesystem via NFS without also exposing `/boot/initramfs*` that might contain encryption keys for the pool. Moving `/boot` to another file system is the best way to prevent accidental network leaks in this case.

This can already be hacked around with the current configuration by, for example, setting

    canmount=legacy
    org.zfsbootmenu:active=on

on the `/boot` filesystem and then making a symlink

```bash
ln -s . /boot/boot
```

However, this is undesirable because ZFSBootMenu will display the boot filesystem rather than the true root in the menu, and a lot will be inconsistent or break entirely. ZFS properties that we read to configure boot environments would have to be set on the boot filesystem instead of the root, the `zbm-kcl` helper may not choose the right destination for edits, the ZBM snapshot menu becomes almost useless, and jump-to-chroot is impossible because what ZBM thinks is the chroot is missing all the root stuff.

With this (untested) change, we can look for a `/boot` mount point in the two places that need to identify kernels in `/boot`: when enumerating kernels in `find_be_kernels` and when loading a kernel to boot in `kexec_kernel`. To make this as minimally invasive as possible, the behavior is:

1. If the root filesystem has *any* files in `/boot` (whether or not they look like kernels, but ignoring dot files), do nothing. We will only attempt to find and mount `/boot` when the directory appears completely empty on the root.
2. If the root contains `/etc/fstab`, walk it to try to find the *first* record for `/boot`. (We should not try to get sophisticated and allow multiple over-mounts here.) If possible, mount the identified device `ro`. If the mount fails, do not attempt further mounts---just fall back to the original behavior. (**NOTE** This allows `/boot` to reside on something other than ZFS, provided the appropriate mount helpers are found in the ZFSBootMenu image.)
3. If `/etc/fstab` does not mention `/boot`, look for *children* of the root filesystem that have `mountpoint=/boot`. If one is found, mount it if possible.

A separate `/boot` breaks atomicity of the boot environment. Attempting to duplicate and boot a snapshot could fail because the snapshot may not be in sync with the current state of the `/boot` filesystem. I do not think we should get too smart about trying to fix this problem. It is a caveat of adopting a non-sanctioned layout that features like snapshot rollbacks may require manual intervention. Despite the possible thorns, this still provides a better ZBM experience than the hacky workaround noted above.

I would have preferred not to track `boot_mounted` and instead just do `umount -R "${mnt}"` where appropriate, but busybox `umount` does not support `-R`. The current form appears to be the least intrusive option.

I have not done any testing of this logic because I want to know that this proposal isn't objectionable. If not, I'll migrate my NFS configuration to work as this change expects and debug as approriate.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -272,7 +272,7 @@ mount_zfs() {
 # returns: 1 on error, otherwise does not return
 
 kexec_kernel() {
-  local selected fs kernel initramfs tdhook output
+  local selected fs kernel initramfs tdhook output boot_mounted
 
   selected="${1}"
   if [ -z "${selected}" ]; then
@@ -297,6 +297,12 @@ kexec_kernel() {
     return 1
   fi
 
+  if find_be_bootdir "${fs}" "${mnt}"; then
+    boot_mounted=yes
+  else
+    boot_mounted=""
+  fi
+
   cli_args="$( load_be_cmdline "${fs}" )"
   root_prefix="$( find_root_prefix "${fs}" "${mnt}" )"
 
@@ -306,6 +312,7 @@ kexec_kernel() {
   then
     zerror "unable to load ${mnt}${kernel} and ${mnt}${initramfs} into memory"
     zerror "${output}"
+    [ -n "${boot_mounted}" ] && umount "${mnt}/boot"
     umount "${mnt}"
     timed_prompt -d 10 \
       -m "$( colorize red 'Unable to load kernel or initramfs into memory' )" \
@@ -321,6 +328,7 @@ kexec_kernel() {
     fi
   fi
 
+  [ -n "${boot_mounted}" ] && umount "${mnt}/boot"
   umount "${mnt}"
 
   while read -r _pool; do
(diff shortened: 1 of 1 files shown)


ahesford:
Did some trivial testing on Alpine. Moving `/boot` to `ztest/ROOT/alpine/boot` with `mountpoint=/boot` with `mountpoint=legacy`, as well as `ztest/ROOT/boot` with `mountpoint=legacy` all work as expected, with `/etc/fstab` on `ztest/ROOT/alpine` configured as appropriate for `legacy` mounts.

zdykstra:
I don't see there being a strong enough justification for this feature to accept the documentation and support headache that it will bring.

While it might be considered a niche and advanced feature, people will attempt to wrangle all types of system configurations into working with it. By demanding that `/boot` is always part of the root dataset, we reduce environments to one happy path and one set of assumptions. This feature will open up a huge array of possible integration/setup issues, all of which will serve to frustrate users and the development team.

Serving an NFS root filesystem with a 'clean' boot directory seems like the simplest solution to the original problem. The node acting as the NFS server can't then boot from the same exact copy of the boot environment as the client nodes, but that's a synchronization problem handled by Ansible/Puppet/Chef/chezmoi, or a 'split apart storage and compute' problem.

I appreciate that you're solving a real problem - I don't believe that it should be solved inside ZBM, though.

ahesford:
I can't defend against this criticism. In fact, I would have avoided documenting this change because I share your view that this has the potential for chaos.

PR #405, opened 2023-01-16
zfsbootmenu: manage efivarfs when needed

zdykstra:
`initcpio` mounts `efivarfs` as read-write, `dracut` doesn't mount it at all. Inside zfsbootmenu-init, we now check if the system is in EFI mode and then mount (or remount) `efivarfs` as read-only.

When chrooting into a boot environment, `efivarfs` is mounted in the chroot as read-only if the zpool is read-only, and read-write if the zpool is read-write.

When entering a recovery shell, `efivarfs` is re-mounted as read-write if `efibootmgr` is present in the initramfs. Exiting the shell remounts it as read-only.

### DIFF
--- a/zfsbootmenu/bin/zfs-chroot
+++ b/zfsbootmenu/bin/zfs-chroot
@@ -8,6 +8,8 @@ cleanup() {
     umount "${_fs}" || zerror "unable to unmount ${_fs}"
   done
 
+  mount_efivarfs
+
   _mnt=()
   trap - HUP INT QUIT ABRT EXIT
 }
@@ -37,6 +39,19 @@ if ! mountpoint="$( allow_rw=yes mount_zfs "${selected}" )"; then
   exit 1
 fi
 
+pool="${selected%%/*}"
+
+# Snapshots and read-only pools always produce read-only mounts
+if is_snapshot "${selected}" || ! is_writable "${pool}"; then
+  writemode="$( colorize green "read-only")"
+  efivarmode="ro"
+else
+  writemode="$( colorize red "read/write")"
+  efivarmode="rw"
+fi
+
+notices+=( "* $( colorize orange "${selected}" ) is mounted ${writemode}" )
+
 # Track submounts so we know how to clean up on exit
 trap cleanup HUP INT QUIT ABRT EXIT
 _mnt=( "${mountpoint}" )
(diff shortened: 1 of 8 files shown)


zdykstra:
Once the code / logic is settled on, I'll add documentation and shell completions as necessary.

ahesford:
If efivarfs is mounted `ro` in -init to normalize everything between dracut and mkinitcpio, it ought to be sufficient to just check that `/sys/firmware/efi/efivars` is a mount point to determine whether it needs to be remounted every else.

zdykstra:
> If efivarfs is mounted `ro` in -init to normalize everything between dracut and mkinitcpio, it ought to be sufficient to just check that `/sys/firmware/efi/efivars` is a mount point to determine whether it needs to be remounted every else.

That's basically what's happening now. `mount_efivarfs` still has an early check to see if the system is even booted in EFI mode, though.

zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1805,6 +1810,10 @@ emergency_shell() {
       zdebug "unmounting: ${mp}"
     fi
   done < /proc/self/mounts
+
+  if command -v efibootmgr && is_efi_system ; then
```
ahesford, line 1814:
```suggestion
  command -v efibootmgr >/dev/null 2>&1 && mount_efivarfs
```

zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1795,6 +1795,11 @@ emergency_shell() {
 	type '$( colorize red "exit")' to return to ZFSBootMenu
 
 	EOF
+
+  if command -v efibootmgr && is_efi_system ; then
```
ahesford, line 1799:
```suggestion
command -v efibootmgr >/dev/null 2>&1 && mount_efivarfs "rw"
```

zfsbootmenu/bin/zfs-chroot
```diff
@@ -55,20 +69,19 @@ mount -t sysfs sys "${mountpoint}/sys" \
 mount -B /dev "${mountpoint}/dev" \
   && _mnt=( "${mountpoint}/dev" "${_mnt[@]}" )
 
+if is_efi_system ; then
```
ahesford, line 72:
```suggestion
if mount_efivarfs "${efivarmode}" ; then
```

ahesford:
I overlooked that. This seems reasonable, although if it bails out early on non-EFI systems, we can skip a few other tests and just jump right into the mount function.

zdykstra:
This seems generally ready to go. The only thing that might be warranted is to note the state of efivarfs when chrooting.

ahesford (approved):
Looks good to me

PR #409, opened 2023-01-28
contrib/: expand font setting capabilities

zdykstra:
ZFSBootMenu binary releases use a default font size that's barely suitable for 2k screens and entirely unsuitable for 4k screens. In an attempt to fix this, a new script has been added that introduces automatic console font resizing.

Utilizing the Terminus font, we can pick a range of bold font sizes to better match screen resolutions. Working backwards from the largest to the smallest, the hook simply runs `setfont` and then checks if COLUMNS is at least 100 characters. If it is not, the next font is selected and it tries again.

Based on documentation, Terminus fonts prefixed with a 'v' have all mappings/codepages in them; so they should support the maximal number of languages.

contrib/20-console-autosize.sh is included in all release and recovery images. The automatic font setting behavior can be disabled via zbm.autosize=(0|off).

contrib/console-init.sh has been renamed to ensure it precedes this hook, since 10-console-init.sh calls console_init from Dracut, which is hard-coded to set a console font. Since console_init also performs setting a keymap, it should always be executed.

The legacy behavior of setting a font via rd.vconsole.font overrides/disables zbm.autosize.

TODO:

* [ ] Document `zbm.autosize` in man pages


ahesford:
Maybe it's time to split `contrib` in two, moving scripts we actually use to a top-level `hooks` tree that can have some official documentation and claim official sanction.

zdykstra:
The `contrib/` directory has multiple different classifications of script right now:

`"Official"`:
* ZFSBootMenu early hook scripts (console related)
* ZFSBootMenu late hook scripts (xhci related)
* generate-zbm syslinux.conf support

`"Unofficial"`:
* generate-zbm esp-sync
* ZFSBootMenu early hook luks-unlock
* ZFSBootMenu late hook snapshot before kexec
* Helper script to drive the build process, to enable dropbear in ZBM

How much sorting should be done here? `contrib/` can be a catch-all for anything unofficial. Officially supported hooks should be placed into a proper `hooks/{zfsbootmenu/...,generate-zbm/...}` directory structure and then copied *somewhere* on the filesystem in the Makefile.

classabbyamp:
+1 on using contrib only for unofficial scripts

ahesford:
I agree we shouldn't be organizing the unofficial stuff in `contrib`. Just dump it there and let people wade around. As for hooks, what about

```
hooks/early-setup.d
hooks/setup.d
hooks/teardown.d
hooks/generate-zbm.pre.d
hooks/generate-zbm.post.d
```

The Makefile can install the whole tree to `/usr/share/zfsbootmenu/hooks`. Then, we can ship a default configuraton that looks in `/etc/zfsbootmenu/hooks/*` and either:
1. Just tell people to symlink what they want in these directories, or
2. Add logic that looks in *both* `/etc/zfsbootmenu/hooks` (or another, configurable path) *plus* `$zfsbootmenu_module_root/hooks` and lets the `/etc/zfsbootmenu` variant shadow the system location (*e.g.*, by symlinking to /dev/null to disable system hooks).

I favor option 1 because it's the simpler of the two.

zdykstra:
I'd prefer to have users symlink in scripts that they want to enable. There's no masking logic then that we have to figure out; if a script or symlink is present in the right directory, it's added (or executed) as needed.

I'd like to add a helper function to `zfsbootmenu/install-helpers.sh` that knows how to find the ZBM run-time hooks directories that copies in files to the ZBM image.  That would replace the need for explicitly defining files in `zfsbootmenu_early_setup`, `zfsbootmenu_setup` and `zfsbootmenu_teardown` variables in `dracut.conf.d/zfsbootmenu.conf` (and the equivalent for initcpio).

PR #417, opened 2023-03-26
docs: remove zpool.cache usage

zdykstra:
There's little benefit to setting a zpool cache file for common root-on-ZFS systems. There are benefits on systems with many disks, but those that need it can set it up. For the rest of us, it's just something that can cause headaches.


zdykstra:
This branch should have been called `cache-me-out`. What a wasted opportunity.

ahesford (approved):
Cashiering the cache is fine with me.

PR #421, opened 2023-03-27
zsnapshot: recovery environment helper tool

### LINKED ISSUE #418: ZBM does not allow snapshot revert if current FS has no kernel
If your only filesystem has no available kernel (for example, zfs failing to build on 6.2.8), ZBM dumps you into recovery shell only,

Even if you have snapshots available to revert to, you can only go to shell, making recovery more difficult than just promoting a snapshot from the menu

zdykstra:
In the event that no boot environments with a valid /boot are found, recovery might be as simple as interacting with an older snapshot. The recovery shell tool `zsnapshot` attempts to make that easier by allowing access to the main snapshot interface for a given dataset.

fzf defaults suitable for the main interfaces were moved to a lib file that can be sourced. These defaults allow smaller tools to match the same behavior that fzf has when invoked from /bin/zfsbootmenu .

select_kernel now returns 1 when the 'kernels' file for a boot environment can't be found, with nothing printed. The caller is expected to handle that condition accordingly.

Fixes #418

### DIFF
--- a/docs/online/recovery-shell.rst
+++ b/docs/online/recovery-shell.rst
@@ -16,6 +16,10 @@ Common Commands
 
   Directly *kexec* a kernel and initramfs from a boot environment, allowing any kernel and initramfs to be loaded into memory and immediately booted.
 
+**zsnapshots** *dataset*
+
+  Access the snapshot browser for a dataset, allowing cloning and rollback operations to be initiated.
+
 **zreport**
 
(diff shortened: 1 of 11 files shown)


Biyoni:
Very quick! I was just about to write it this morning!

Tested, working great!

However the binary is called `zsnapshots` but the docs refer to `zsnapshot`

PR #424, opened 2023-05-05
zfsbootmenu: warn when canmount=on is set for root dataset

zdykstra:
Root datasets should never be automatically mounted. Anything that is considered a root dataset (regardless of a valid kernel/initramfs pair being present) should be checked for `canmount=on` being set. If it is, throw up a zwarn() . This possibly needs to be more aggressively warned via a timed_prompt - forcing user interaction.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-lib.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-lib.sh
@@ -546,7 +546,7 @@ get_sort_key() {
 # returns: 0 iff at least one valid BE was found
 
 populate_be_list() {
-  local be_list fs mnt active candidates ret sort_key
+  local be_list fs canmount mnt active candidates ret sort_key
 
   be_list="${1}"
   if [ -z "${be_list}" ]; then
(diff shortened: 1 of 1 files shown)


ahesford (approved):
I think a subtle warning like this or a more obvious, passive indicator is fine. I don't know that we should interrupt regular flow with a timed prompt.

ahesford (approved):
I think a subtle warning like this or a more obvious, passive indicator is fine. I don't know that we should interrupt regular flow with a timed prompt.

zdykstra:
Yeah. If we continue to see people impacted by this, the warning can be re-evaluated. For now, a normal warning delivered via the common mechanism is a good start.

ahesford:
How about this: count the number of root filesystems; if there is exactly one and has `canmount=auto`, use the unobtrusive logged warning. If there are more than one and any of them have `canmount=auto`, throw up a timed prompt warning the user that these may fail to boot.

zdykstra:
I'm checking for the wrong condition - whoops. I'll rework this to count and set a flag file that zfsbootmenu-init can check for.

PR #426, opened 2023-05-17
docs/guides/debian/: add initial bookworm guide

### LINKED ISSUE #425: Update Documentation to mention Debian 12 Bookworm (successful installation)
I recently followed the Debian Bullseye UEFI setup instructions, but installed Debian 12 Bookworm instead of Bullseye. It worked flawlessly with no alterations (besides `s/bullseye/bookworm/g`) needed to the docs. I used a Single Disk, so I can't speak for the other installation options, but those should work fine as well.

zdykstra:
Prepare for the upcoming Debian 12 (Bookworm) release by adding a guide for it. Links for CD Images in both the Bullseye and Bookworm guides need to be fixed so that they point to the right resources after Debian 12 is officially released.

We can possibly create links for Bullseye images to some persistent directory at https://cdimage.debian.org/mirror/cdimage/archive/ . I'll need to poke around the cdimage.debian.org in more detail to find a stable URL, though.

Closes #425

### DIFF
--- a/docs/guides/debian.rst
+++ b/docs/guides/debian.rst
@@ -4,4 +4,5 @@ Debian
 .. toctree::
   :titlesonly:
 
-  debian/uefi
+  debian/bullseye-uefi
+  debian/bookworm-uefi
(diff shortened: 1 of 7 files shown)


zdykstra:
https://lists.debian.org/debian-devel-announce/2023/04/msg00007.html

ahesford:
I most certainly will **not** be hosting a Debian release party.

PR #429, opened 2023-05-31
zfs-chroot: add read-write prompt

zdykstra:
When entering a chroot for a regular dataset and the pool is read-only, provide a timed_prompt that lets the user choose between entering a read-only chroot (return key/enter key) or a read-write chroot (escape key). If the dataset is a snapshot or the pool is already read-write, do not show the prompt.

Fixes #428


zdykstra:
https://asciinema.org/a/YJNpLPvBRDtUoqk1qlCINNcEK

ahesford (approved):
Looks good to me

PR #433, opened 2023-06-14
zbm-builder.sh: Add support for running on hosts with SELinux enabled

chenxiaolong:
This commit updates `zbm-builder.sh` to include a new `-M` option for specifying podman/docker volume mount flags. On systems with SELinux enabled, `-M z` must be specified so that the volume will be relabeled with the `container_file_t` SELinux type. Otherwise, the directory contents are not accessible within the container.

In addition, when using `-l <source dir>`, `-O -e=DRACUT_NO_XATTR=1` must be specified. By default, when dracut copies files to its temporary directory, it runs:

    cp --preserve=mode,xattr,timestamps,ownership <...>

This environment variable makes dracut disable the `xattr` flag. This is necessary because cp would try to preserve the `security.selinux` xattr, which is never allowed inside a container. This should have no impact on the generated initramfs image because ZFSBootMenu does not use any custom xattrs and cpio is incapable of storing them anyway. mkinitcpio does not try to set xattrs, so no workaround is needed for that.

### DIFF
--- a/docs/guides/general/container-building.rst
+++ b/docs/guides/general/container-building.rst
@@ -108,6 +108,11 @@ Building a ZFSBootMenu Image
 To build a default image, invoke ``zbm-builder.sh`` with no arguments. For example, from the directory that contains the
 script, run ``./zbm-builder.sh`` to produce a default kernel/initramfs pair in the ``./build`` subdirectory.
 
+To build on a host with SELinux enabled, the ``-M z -O -e=DRACUT_NO_XATTR=1`` options are needed. The `z` mount flag will
+relabel both the build (``-b``) and source (``-l``) directories with the ``container_file_t`` type so that the container
+is able to access the files. The ``DRACUT_NO_XATTR`` will prevent dracut from trying to set xattrs on temporary files
+within the container. Specifically, a container is never allowed to set the ``security.selinux`` xattr.
+
 The default behavior of ``zbm-builder.sh`` will:
 
 1. Pull the default builder image, ``ghcr.io/zbm-dev/zbm-builder:latest``.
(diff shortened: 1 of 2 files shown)


chenxiaolong:
I did tested with mkinitcpio. It doesn't need anything like `DRACUT_NO_XATTR` because it never tries to set xattrs in the first place.

ahesford:
I think this is a bit too specific. We're already half-way there with the `-O` flag, so I'd propose adding an `-M` flag that can be used to specify volume mount options for the `/build` and `/zbm` mountpoints.  If not empty, it can be added to the volume flags the way you are doing `SELINUX_FLAG`, but also lets other users to other volume weirdness with their containers. (The `/zbm` mount should still have `ro` forcefully inserted before any `MOUNT_FLAGS` that might be appended.)

As for the `DRACUT_NO_XATTR`, this can already be accomplished with `-O --env=DRACUT_NO_XATTR=1`.

Yes, this requires two flags instead of your one, but it is totally generic and can be simplified by either
- Just aliasing the `zbm-builder.sh` command to add the `-M z -O --env=DRACUT_NO_XATTR=1` option, or
- Creating a `zbm-builder.conf` file with the contents
```sh
MOUNT_FLAGS="z"
RUNTIME_ARGS+=( --env=DRACUT_NO_XATTR=1 )
```
and making sure `zbm-builder.sh` sees the config.

chenxiaolong:
Thanks for the feedback. I've updated the PR with your `-M` suggestion and updated the documentation accordingly.

ahesford:
Looks good, although I propose a few changes to the documentation in 00e500f33ce09cc9e843fccbeaede32f3bcbb4e5. Please edit the changes as appropriate or, if you approve, I'll squash and merge.

chenxiaolong:
Thanks, @ahesford! I like the new wording you have in 00e500f33ce09cc9e843fccbeaede32f3bcbb4e5 as-is.

ahesford:
Thanks again!

PR #434, opened 2023-06-16
zfsbootmenu-core: try to load keys in mount_zfs

ahesford:
Attempting to load encryption keys when mounting a ZFS filesystem avoids the need to manually load keys in a few places and ensure that mount_zfs tries hard to do the right thing even when run interactively.

Because mount_zfs is also called from cache_key and it might be problematic to enter some indirect recursion when manipulating ZFS keys, setting NO_LOAD_KEY=1 will tell mount_zfs to skip this attempt and just fail when the BE is locked.

This isn't really tested, but it should also work.

(While I was at it, I replaced `reset` and `clear` in `zfs-chroot` with equivalent `tput` commands in case the wrappers aren't installed or the aliases set in the recovery `bashrc` don't make their way into the invocation.

### DIFF
--- a/zfsbootmenu/bin/zfs-chroot
+++ b/zfsbootmenu/bin/zfs-chroot
@@ -41,12 +41,11 @@ if ! is_snapshot "${selected}" && ! is_writable "${pool}" ; then
     -r "Enter r/o chroot" \
     -p "Entering chroot in $( colorize yellow "%0.2d" ) seconds" ; then
       set_rw_pool "${pool}"
-      CLEAR_SCREEN=1 load_key "${selected}"
   fi
 fi
 
-reset
-clear
+tput reset
+tput clear
 
 if ! mountpoint="$( allow_rw=yes mount_zfs "${selected}" )"; then
   zerror "failed to mount ${selected}"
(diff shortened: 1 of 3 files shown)


midzelis:
Thanks!

zdykstra:
I don't see this as being functional with the current design of `mount_zfs` - It prints the mount point of the dataset to stdout. 

`load_key` can't be called from a subshell, since the user needs to interact with it.

Every usage of `mount_zfs` is in a subshell:

```
$ rg mount_zfs *
libexec/zfsbootmenu-diff
49:if ! mnt="$( mount_zfs "${base_fs}" )" ; then

lib/zfsbootmenu-core.sh
219:mount_zfs() {
298:  if ! mnt=$( mount_zfs "${fs}" ); then
698:  if ! mnt="$( mount_zfs "${fs}" )"; then
1588:  if ! mnt="$( NO_LOAD_KEY=1 mount_zfs "${keysrc}" )"; then

bin/zfs-chroot
50:if ! mountpoint="$( allow_rw=yes mount_zfs "${selected}" )"; then

```

The only way I can see this working is if we give up printing the mount point to stdout in `mount_zfs`, and instead do something like:

```
if ! mount_zfs "${fs}"; then
  # do something
  return 1
else
  mnt="$( be_location "${fs}" )/mnt"
fi
```
This seems like it'd be trading one gotcha for another - would it not make as much sense to precede all uses of `mount_zfs` with a `load_key` ?

ahesford:
I completely overlooked the subshell problem! This is what happens when I propose changes without actually testing them...

PR #449, opened 2023-07-17
Add clevis-based automated unlock functionality

rdmitry0911:


ahesford:
This kind of highly specialized use is not appropriate for inclusion in the general flow of ZFSBootMenu. If you can make this robust as an early-setup hook, we would consider adding the functionality as a `contrib` script.

PR #452, opened 2023-07-18
Introducing load_key hook in zfsbootmenu/lib/zfsbootmenu-core.sh

rdmitry0911:
This is a hook placeholder for scripts to run just before loading key for a dataset. It can be used for example for automatic dataset unlocking using clevis or for some other purposes. The only argument provided to the hook is dataset name
As a proof of concept a load_key_clevis_hook.sh script is included in contrib directory

### DIFF
--- a/docs/guides/general/container-building.rst
+++ b/docs/guides/general/container-building.rst
@@ -139,14 +139,16 @@ ZFSBootMenu supports :ref:`custom hooks <zbm-dracut-options>` in three stages:
    any pools.
 2. ``setup`` hooks run after pools are imported, right before ZFSBootMenu will either boot a default environment or
    present a menu.
-3. ``teardown`` hooks run immediately before ZFSBootMenu will ``kexec`` the kernel for the selected environment.
+3. ``load_key`` hooks run immediately before attempting to load a key for a locked filesystem. Each hook is called with the name of the encryption root to be unlocked as its sole argument.
+4. ``teardown`` hooks run immediately before ZFSBootMenu will ``kexec`` the kernel for the selected environment.
 
 When ``zbm-builder.sh`` runs, it will identify custom hooks as executable files in the respective subdirectories of its
 build directory:
 
 1. ``hooks.early_setup.d``
 2. ``hooks.setup.d``
-3. ``hooks.teardown.d``
+3. ``hooks.load_key.d``
+4. ``hooks.teardown.d``
 
 For each hook directory that contains at least one executable file, ``zbm-builder.sh`` will write custom configuration
 snippets for ``dracut`` and ``mkinitcpio`` that will include these files in the output images.
--- a/docs/man/zfsbootmenu.7.rst
+++ b/docs/man/zfsbootmenu.7.rst
@@ -233,6 +233,15 @@ In addition to standard dracut configuration options, the ZFSBootMenu dracut mod
 
   Any installed hooks are run right before the ZFSBootMenu menu will be presented; ZFS pools will generally have been imported and the default boot environment will be available in the *BOOTFS* environment variable. Hooks will not be run if the countdown timer expires (or was set to zero) and the default boot environment is automatically selected. **Note:** The hooks may be run multiple times if the menu is invoked multiple times, e.g., by dropping to an emergency shell and then returning to the menu. If a script should only run once, the script is responsible for keeping track of this.
 
+<<<<<<< HEAD
+=======
+**zfsbootmenu_load_key=<executable-list>**
+
+  An optional variable specifying a space-separated list of paths to load-key hooks that will be installed in the ZFSBootMenu initramfs. Any path in the list **<executable-list>** that exists and is executable will be installed.
+
+  Any installed hooks are run immediately before ZFSBootMenu menu attempts to load a key for the encryption root of a locked filesystem. Each hook will be run with the name of the encryption root as its sole argument. If any hook exits with a non-zero return code, the load attempt will be aborted and the encryption root will not be unlocked. Hooks can be used to prepare keys needed to unlock the encryption root or may even attempt to unlock the root directly; a filesystem that is unlocked after the hooks have run will be considered to have had its keys successfully loaded. **Note:** The hooks may be run multiple times if keys for a given encryption root need to be loaded multiple times, e.g., by triggering a re-import of the pool. If a script should only run once, the script is responsible for keeping track of this.
+
+>>>>>>> f2ddbec (Update docs/man/zfsbootmenu.7.rst)
 **zfsbootmenu_teardown=<executable-list>**
 
   An optional variable specifying a space-separated list of paths to teardown hooks that will be installed in the ZFSBootMenu initramfs. Any path in the list **<executable-list>** that exists and is executable will be installed.
@@ -260,6 +269,10 @@ In addition to standard dracut configuration options, the ZFSBootMenu dracut mod
     kexec --unload
 
   should be sufficient to return to the main menu. Likewise, the hook may construct and execute its own *kexec* command to alter boot-time parameters. This may be useful, for example, to allow ZFSBootMenu to select a boot environment and then restructure the boot process to launch a Xen kernel with the selected environment configured as dom0.
+
+**zfsbootmenu_skip_gcc_s=yes**
+
+  The ZFSBootMenu module attempts to detect and install a copy of the library **libgcc_s.so** in its initramfs image on glibc systems. Because several executables may have latent dependencies on this library via a **dlopen** call in glibc itself, a failure to detect and install the library will cause initramfs generation to fail. If the host system has no dependencies on **libgcc_s.so**, set **zfsbootmenu_skip_gcc_s=yes** to avoid this failure. Alternatively, if **libgcc_s.so** is present in an undetected location, set this option and configure Dracut to explicitly install the library.
 
 .. _zbm-mkinitcpio-options:
 
--- a/docs/man/zfsbootmenu.7.rst
+++ b/docs/man/zfsbootmenu.7.rst
@@ -233,15 +233,12 @@ In addition to standard dracut configuration options, the ZFSBootMenu dracut mod
 
   Any installed hooks are run right before the ZFSBootMenu menu will be presented; ZFS pools will generally have been imported and the default boot environment will be available in the *BOOTFS* environment variable. Hooks will not be run if the countdown timer expires (or was set to zero) and the default boot environment is automatically selected. **Note:** The hooks may be run multiple times if the menu is invoked multiple times, e.g., by dropping to an emergency shell and then returning to the menu. If a script should only run once, the script is responsible for keeping track of this.
 
-<<<<<<< HEAD
-=======
 **zfsbootmenu_load_key=<executable-list>**
 
   An optional variable specifying a space-separated list of paths to load-key hooks that will be installed in the ZFSBootMenu initramfs. Any path in the list **<executable-list>** that exists and is executable will be installed.
 
   Any installed hooks are run immediately before ZFSBootMenu menu attempts to load a key for the encryption root of a locked filesystem. Each hook will be run with the name of the encryption root as its sole argument. If any hook exits with a non-zero return code, the load attempt will be aborted and the encryption root will not be unlocked. Hooks can be used to prepare keys needed to unlock the encryption root or may even attempt to unlock the root directly; a filesystem that is unlocked after the hooks have run will be considered to have had its keys successfully loaded. **Note:** The hooks may be run multiple times if keys for a given encryption root need to be loaded multiple times, e.g., by triggering a re-import of the pool. If a script should only run once, the script is responsible for keeping track of this.
 
->>>>>>> f2ddbec (Update docs/man/zfsbootmenu.7.rst)
 **zfsbootmenu_teardown=<executable-list>**
 
   An optional variable specifying a space-separated list of paths to teardown hooks that will be installed in the ZFSBootMenu initramfs. Any path in the list **<executable-list>** that exists and is executable will be installed.
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -68,26 +68,21 @@ install() {
     fi
   done
 
-  # Workaround for zfsonlinux/zfs#4749 by ensuring libgcc_s.so(.1) is included
-  _ret=0
-  # If zpool requires libgcc_s.so*, dracut will track and include it
-  if ! ldd "$( command -v zpool )" | grep -qF 'libgcc_s.so'; then
-    # On systems with gcc-config (Gentoo, Funtoo, etc.), use it to find libgcc_s
-    if command -v gcc-config >/dev/null 2>&1; then
-      dracut_install "/usr/lib/gcc/$(s=$(gcc-config -c); echo "${s%-*}/${s##*-}")/libgcc_s.so.1"
-      _ret=$?
-    # Otherwise, use dracut's library installation function to find the right one
-    elif ! inst_libdir_file "libgcc_s.so*"; then
-      # If all else fails, just try looking for some gcc arch directory
-      dracut_install /usr/lib/gcc/*/*/libgcc_s.so*
-      _ret=$?
-    fi
-  fi
-
-  if [ ${_ret} -ne 0 ]; then
-    dfatal "Unable to install libgcc_s.so"
+  # Add libgcc_s as appropriate
+  local _libgcc_s
+  if ! _libgcc_s="$( find_libgcc_s )"; then
+    dfatal "Unable to locate libgcc_s.so"
     exit 1
   fi
+
+  local _lib
+  while read -r _lib ; do
+    [ -n "${_lib}" ] || continue
+    if ! dracut_install "${_lib}"; then
+      dfatal "Failed to install '${_lib}'"
+      exit 1
+    fi
+  done <<< "${_libgcc_s}"
 
   # shellcheck disable=SC2154
   while read -r doc ; do
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -156,6 +156,18 @@ install() {
     done
   fi
 
+  # Install "load_key" hooks
+  # shellcheck disable=SC2154
+  if [ -n "${zfsbootmenu_load_key}" ]; then
+    for _exec in ${zfsbootmenu_load_key}; do
+      if [ -x "${_exec}" ]; then
+        inst_simple "${_exec}" "/libexec/load_key.d/$(basename "${_exec}")" || _ret=$?
+      else
+        dwarning "load_key script (${_exec}) missing or not executable; cannot install"
+      fi
+    done
+  fi
+
   # Install "teardown" hooks
   # shellcheck disable=SC2154
   if [ -n "${zfsbootmenu_teardown}" ]; then
--- a/dracut/module-setup.sh
+++ b/dracut/module-setup.sh
@@ -113,7 +113,7 @@ install() {
     inst_simple "${_libexec}" "/libexec/$( basename "${_libexec}" )" || _ret=$?
   done
 
-  # User-facing utilities, useful for running in a recover shell
+  # User-facing utilities, useful for running in a recovery shell
   for _bin in "${zfsbootmenu_module_root}"/bin/*; do
     inst_simple "${_bin}" "/bin/$( basename "${_bin}" )" || _ret=$?
   done
--- a/etc/zbm-builder/mkinitcpio.conf
+++ b/etc/zbm-builder/mkinitcpio.conf
@@ -28,6 +28,7 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard)
 #
 #     hooks.early_setup.d
 #     hooks.setup.d
+#     hooks.load_key.d
 #     hooks.teardown.d
 #
 # of the same build directory. Support for mkinitcpio.conf.d mimics similar
--- a/etc/zfsbootmenu/mkinitcpio.conf
+++ b/etc/zfsbootmenu/mkinitcpio.conf
@@ -75,7 +75,7 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard zfsbootmenu)
 # tested and may break some (non-essential) ZFSBootMenu features.
 #zfsbootmenu_miser="no"
 
-# zfsbootmenu_early_setup, zfsbootmenu_setup, zfsbootmenu_teardown
+# zfsbootmenu_early_setup, zfsbootmenu_setup, zfsbootmenu_load_key, zfsbootmenu_teardown
 # ZFSBootMenu supports user-provided hooks that may be run at various points in
 # the boot process. Set each of these variables to an array of executable files
 # that should be installed as user hooks in the provided image. Files that are
@@ -94,6 +94,11 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard zfsbootmenu)
 # environment is about to be launched. Any writable ZFS pools will have been
 # exported, but read-only pools will generally still be available.
 #
+# 'zfsbootmenu_load_key' hooks are run immediately before attempting to load an
+# encryption key and unlocked a locked filesystem. Each hook is called with a
+# single argument: the name of an encryption root that should be unlocked.
+#
 #zfsbootmenu_early_setup=()
 #zfsbootmenu_setup=()
+#zfsbootmenu_load_key=()
 #zfsbootmenu_teardown=()
--- a/etc/zfsbootmenu/mkinitcpio.conf
+++ b/etc/zfsbootmenu/mkinitcpio.conf
@@ -90,13 +90,13 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard zfsbootmenu)
 # 'zfsbootmenu_setup' hooks are run after pools are imported and right before
 # the menu is presented.
 #
+# 'zfsbootmenu_load_key' hooks are run immediately before attempting to load an
+# encryption key and unlock a locked filesystem. Each hook is called with a
+# single argument: the name of an encryption root that should be unlocked.
+#
 # 'zfsbootmenu_teardown' hooks are run immediately before a chosen boot
 # environment is about to be launched. Any writable ZFS pools will have been
 # exported, but read-only pools will generally still be available.
-#
-# 'zfsbootmenu_load_key' hooks are run immediately before attempting to load an
-# encryption key and unlocked a locked filesystem. Each hook is called with a
-# single argument: the name of an encryption root that should be unlocked.
 #
 #zfsbootmenu_early_setup=()
 #zfsbootmenu_setup=()
--- a/initcpio/install/zfsbootmenu
+++ b/initcpio/install/zfsbootmenu
@@ -179,6 +179,12 @@ build() {
     done
 
     # shellcheck disable=SC2154
+    for _file in "${zfsbootmenu_load_key[@]}"; do
+        [ -x "${_file}" ] || continue
+        add_file "${_file}" "/libexec/load_key.d/${_file##*/}"
+    done
+
+    # shellcheck disable=SC2154
     for _file in "${zfsbootmenu_teardown[@]}"; do
         [ -x "${_file}" ] || continue
         add_file "${_file}" "/libexec/teardown.d/${_file##*/}"
--- a/initcpio/install/zfsbootmenu
+++ b/initcpio/install/zfsbootmenu
@@ -138,6 +138,22 @@ build() {
     # Binaries required for ZBM operation
     add_zbm_binaries
 
+    # Add libgcc_s as appropriate
+    local _libgcc_s
+    if ! _libgcc_s="$( find_libgcc_s )"; then
+        error "unable to locate libgcc_s.so"
+        exit 1
+    fi
+
+    local _lib
+    while read -r _lib ; do
+        [ -n "${_lib}" ] || continue
+        if ! add_binary "${_lib}"; then
+            error "Failed to install '${_lib}'"
+            exit 1
+        fi
+    done <<< "${_libgcc_s}"
+
     # On-line documentation
     while read -r doc; do
         relative="${doc#"${zfsbootmenu_module_root}/"}"
--- a/zbm-builder.sh
+++ b/zbm-builder.sh
@@ -239,7 +239,7 @@ if ! [ -r "${BUILD_DIRECTORY}"/config.yaml ]; then
 fi
 
 # Try to include ZBM hooks in the images by default
-for stage in early_setup setup teardown; do
+for stage in early_setup setup load_key teardown; do
   [ -d "${BUILD_DIRECTORY}/hooks.${stage}.d" ] || continue
 
   # Only executable hooks are added to the image
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -1659,7 +1659,7 @@ cache_key() {
 # NOTE: this function should *not* be called from a subshell
 
 load_key() {
-  local fs encroot key keypath keyformat keylocation keysource
+  local fs encroot key keypath keyformat keylocation keysource _hook
 
   fs="${1}"
   if [ -z "${fs}" ]; then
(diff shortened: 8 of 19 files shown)


zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1673,6 +1673,16 @@ load_key() {
     return 0
   fi
 
+  # Run early load_key hooks, if they exist
+  if [ -d /libexec/load_key.d ]; then
+    tput clear
```
ahesford, line 1678:
```suggestion
```
The screen should not be cleared before hooks run; if a hook needs a clear screen, do it there.

zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1673,6 +1673,16 @@ load_key() {
...
+    for _hook in /libexec/load_key.d/*; do
```
ahesford, line 1679: `_hook` should be declared local at the top of this function.

zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1673,6 +1673,16 @@ load_key() {
...
+      zinfo "Processing hook: ${_hook}"
+      [ -x "${_hook}" ] && "${_hook}" "${encroot}"
```
ahesford, line 1681:
```suggestion
      [ -x "${_hook}" ] || continue
      if ! "${_hook}" "${encroot}"; then
        zwarn "load-key hook failed for '${encroot}', aborting load attempt"
        return 1
      fi
```
Let's use the return code of hooks to signal a failure condition that aborts key loading altogether.

zfsbootmenu/help-files/92/zfsbootmenu.7.ansi
```diff
@@ -136,7 +136,7 @@
 
     [1mzbm.hookroot=<hookspec>[0m
 
-      Tell ZFSBootMenu to attempt to read any early-setup, setup or teardown hooks from the
+      Tell ZFSBootMenu to attempt to read any early-setup, setup load_key or teardown hooks from the
```
ahesford, line 139:
```suggestion
      Tell ZFSBootMenu to attempt to read any early-setup, setup, load_key or teardown hooks from the
```
You should just be running `releng/rst2help.sh` to render these help files based on `docs/man` sources.

etc/zfsbootmenu/mkinitcpio.conf
```diff
@@ -94,6 +94,12 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard zfsbootmenu)
 # environment is about to be launched. Any writable ZFS pools will have been
 # exported, but read-only pools will generally still be available.
 #
+# 'zfsbootmenu_load_key' hooks are run immediately before loading a key
+# for an encrypted dataset The only argument provided is dataset name.  
+# They can be used for alternative ways for key loading, like clevis, for example.
```
ahesford, line 99:
```suggestion
# 'zfsbootmenu_load_key' hooks are run immediately before attempting to load an
# encryption key and unlocked a locked filesystem. Each hook is called with a
# single argument: the name of an encryption root that should be unlocked.
```
Also, move this after the `zfsbootmenu_setup` description but before `zfsbootmenu_teardown` to keep the order in sync with the general order in which hooks will be executed.

etc/zfsbootmenu/mkinitcpio.conf
```diff
@@ -75,7 +75,7 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard zfsbootmenu)
 # tested and may break some (non-essential) ZFSBootMenu features.
 #zfsbootmenu_miser="no"
 
-# zfsbootmenu_early_setup, zfsbootmenu_setup, zfsbootmenu_teardown
+# zfsbootmenu_early_setup, zfsbootmenu_setup, zfsbootmenu_teardown zfsbootmenu_load_key
```
ahesford, line 78:
```suggestion
# zfsbootmenu_early_setup, zfsbootmenu_setup, zfsbootmenu_load_key, zfsbootmenu_teardown
```

etc/zfsbootmenu/mkinitcpio.conf
```diff
@@ -94,6 +94,12 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard zfsbootmenu)
...
 #
+# 'zfsbootmenu_load_key' hooks are run immediately before loading a key
+# for an encrypted dataset The only argument provided is dataset name.  
+# They can be used for alternative ways for key loading, like clevis, for example.
+#
 #zfsbootmenu_early_setup=()
 #zfsbootmenu_setup=()
 #zfsbootmenu_teardown=()
+#zfsbootmenu_load_key=()
+
```
ahesford, line 105:
```suggestion
#zfsbootmenu_load_key=()
#zfsbootmenu_teardown=()
```

initcpio/install/zfsbootmenu
```diff
@@ -184,6 +184,12 @@ build() {
         add_file "${_file}" "/libexec/teardown.d/${_file##*/}"
     done
 
+# shellcheck disable=SC2154
+    for _file in "${zfsbootmenu_load_key[@]}"; do
+        [ -x "${_file}" ] || continue
+        add_file "${_file}" "/libexec/load_key.d/${_file##*/}"
+    done
```
ahesford, line 191: Put this before the `zfsbootmenu_teardown` block just to keep the order logically consistent.

zbm-builder.sh
```diff
@@ -239,7 +239,7 @@ if ! [ -r "${BUILD_DIRECTORY}"/config.yaml ]; then
 fi
 
 # Try to include ZBM hooks in the images by default
-for stage in early_setup setup teardown; do
+for stage in early_setup setup teardown load_key; do
```
ahesford, line 242:
```suggestion
for stage in early_setup setup load_key teardown; do
```

etc/zbm-builder/mkinitcpio.conf
```diff
@@ -29,6 +29,7 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard)
 #     hooks.early_setup.d
 #     hooks.setup.d
 #     hooks.teardown.d
+#     hooks.load_key.d
```
ahesford, line 32:
```suggestion
#     hooks.load_key.d
#     hooks.teardown.d
```

dracut/module-setup.sh
```diff
@@ -168,6 +168,18 @@ install() {
...
+  # shellcheck disable=SC2154
+  if [ -n "${zfsbootmenu_load_key}" ]; then
+    for _exec in ${zfsbootmenu_load_key}; do
+      if [ -x "${_exec}" ]; then
+        inst_simple "${_exec}" "/libexec/load_key.d/$(basename "${_exec}")" || _ret=$?
+      else
+        dwarning "load_key script (${_exec}) missing or not executable; cannot install"
+      fi
+    done
+  fi
```
ahesford, line 181: Move this above the `teardown` block to keep the order consistent.

docs/guides/general/container-building.rst
```diff
@@ -139,14 +139,16 @@ ZFSBootMenu supports :ref:`custom hooks <zbm-dracut-options>` in three stages:
    any pools.
 2. ``setup`` hooks run after pools are imported, right before ZFSBootMenu will either boot a default environment or
    present a menu.
-3. ``teardown`` hooks run immediately before ZFSBootMenu will ``kexec`` the kernel for the selected environment.
+3. ``load_key`` hooks run immediately before key loading for a dataset. The only argument provided to a hook is ``dataset`` name.
```
ahesford, line 142:
```suggestion
3. ``load_key`` hooks run immediately before attempting to load a key for a locked filesystem. Each hook is called with the name of the encryption root to be unlocked as its sole argument.
```

README.md
```diff
@@ -4,7 +4,9 @@
...
 
-In essence, ZFSBootMenu is a small, self-contained Linux system that knows how to find other Linux kernels and initramfs images within ZFS filesystems. When a suitable kernel and initramfs are identified (either through an automatic process or direct user selection), ZFSBootMenu launches that kernel using the `kexec` command.
+In essence, ZFSBootMenu is a small, self-contained Linux system that knows how to find other Linux kernels and initramfs images within ZFS filesystems. When a suitable kernel and initramfs are identified (either through an automatic process or direct user selection), ZFSBootMenu launches that kernel using the `kexec` command. 
+
+### This fork introduces `load_key hook` placeholder inside zfsbootmenu-core.sh for user scripts to run just before loading a key for the dataset. It can be used for example for automatic dataset unlocking using clevis or for some other purposes. The only argument provided to the hook is dataset name As a proof of concept a load_key_clevis_hook.sh script is included in contrib directory that adds automated anlocking of the pool based on clevis package and tpm2 crypto module
```
ahesford, line 9: None of these changes belong here.

contrib/README.md
```diff
@@ -59,3 +59,8 @@ review the scripts themselves for more thorough descriptions of their use.
 
 - `zbm-sign.pl` - A Perl script, suitable for use as a generate-zbm post-run
   hook, that will sign ZFSBootMenu EFI images for use with Secure Boot.
+
+- `load_key_clevis_hook.sh` - This is a proof of concept for automated dataset unlocking
+  in trusted environment using clevis.
```
ahesford, line 64:
```suggestion
- `load_key_clevis_hook.sh` - A proof-of-concept hook to decrypt ZFS filesystem keys using `clevis`.
```

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+      dr="${netconf[2]}"
+      dns="${netconf[3]}"
+      zdebug "Reconfigure network: dev: $dev ip: $ip default route: $dr dns: $dns"
+      ip addr add "$ip" brd + dev "$dev"
+      ip route add default via "$dr"
+      echo "nameserver $dns" > /etc/resolv.conf
+    fi
+  fi
+
+
```
ahesford, line 92: This shouldn't be here; to use this hook, one of the dracut or mkinitcpio network packages should be configured properly to ensure network access.

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+#
+
+get_fs_value()
+{
+        fs="$1"
+        value=$2
+
+        zfs get -H -ovalue "$value" "$fs" 2> /dev/null
+}
+
```
ahesford, line 40: This is only used once (after striking the network configuration block) and doesn't need to be wrapped in a function.

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+      fi
+    done
+    echo "$PASS"
+    return 0
+  else
+    # In fact we don't want to reseal
+    echo ""
+    return 1
+  fi
+}
```
ahesford, line 65: This kind of error handling shouldn't be done in a hook like this. If the unlock fails, let the user drop to a recovery shell and re-encrypt the passphrase.

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+      dns="${netconf[3]}"
+      zdebug "Reconfigure network: dev: $dev ip: $ip default route: $dr dns: $dns"
+      ip addr add "$ip" brd + dev "$dev"
+      ip route add default via "$dr"
+      echo "nameserver $dns" > /etc/resolv.conf
+    fi
+  fi
+
+
+  CLEVIS_CHECK="$(zfs get -H -p -o value latchset.clevis:decrypt -s local $dataset_for_clevis_unlock)"
```
ahesford, line 93:
```suggestion
  CLEVIS_CHECK="$(zfs get -H -o value latchset.clevis:decrypt -s local "$dataset_for_clevis_unlock")" || CLEVIS_CHECK=
```
You need to properly quote the argument to avoid splitting filesystem names that might have spaces, and you should probably capture the return code here to sanitize any output in the event of a failure.

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+  if [[ "$CLEVIS_CHECK" == "yes" ]]; then
+    zdebug "Found dataset for clevis unlocking: $dataset_for_clevis_unlock"
+    KEYLOCATION="$(get_fs_value "${dataset_for_clevis_unlock}" keylocation)"
+    if [[ -z "$(echo "$KEYLOCATION"|grep file)" ]]; then
+        # That's not us
+        zdebug "Warning: keylocation is not file while clevis unlock is set for dataset $dataset_for_clevis_unlock"
+        return 0
+    fi
```
ahesford, line 101: Yikes, this is not the way to check if this is a file URI.

```suggestion
    KEYLOCATION="$(zfs get -H -o value keylocation "${dataset_for_clevis_unlock}")" || KEYLOCATION=
    KEYFILE="${KEYLOCATION#file://}"
    if [ "${KEYLOCATION}" = "${KEYFILE}" ] || [ -z "${KEYFILE}" ]; then
      return 0
    fi
```

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+              zdebug "Warning: We failed. The jwe is not correctly stored in dataset $dataset_for_clevis_unlock"
+              return 1
+            fi
+            echo "$RESEAL" >"$KEYLOCATION"
+            return 0
+          else
+            # We don't want to reseal
+            return 1
+          fi
+        fi
```
ahesford, line 149: I don't think any of this is a good idea. Handling these error conditions is best done interactively by a user who understands what's going on, not in an automatic hook.

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+  if [[ "$CLEVIS_CHECK" == "yes" ]]; then
+    zdebug "Found dataset for clevis unlocking: $dataset_for_clevis_unlock"
+    KEYLOCATION="$(get_fs_value "${dataset_for_clevis_unlock}" keylocation)"
+    if [[ -z "$(echo "$KEYLOCATION"|grep file)" ]]; then
+        # That's not us
+        zdebug "Warning: keylocation is not file while clevis unlock is set for dataset $dataset_for_clevis_unlock"
+        return 0
+    fi
+    # We suppose the keylocation has value in format file:///something/key
+    KEYLOCATION="${KEYLOCATION:7}"
```
ahesford, line 103:
```suggestion
```
You should just use `KEYFILE`, as I set above, for all subsequent uses of `KEYLOCATION` in this script.

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+    KEYSTATUS="$(zfs get -H -p -o value keystatus -s none $dataset_for_clevis_unlock)"
+    if [[ "$KEYSTATUS" == "unavailable" ]]; then
+        # Prepare the key with password for unlocking in right place
+        mkdir -p "$(dirname "$KEYLOCATION")"
+        JWE="$(zfs get -H -p -o value latchset.clevis:jwe -s local $dataset_for_clevis_unlock)"
```
ahesford, line 108:
```suggestion
        JWE="$(zfs get -H -p -o value latchset.clevis:jwe -s local "$dataset_for_clevis_unlock")"
```
You should also monitor the return value of this command and take appropriate action when it fails, instead of passing garbage to `clevis decrypt`.

contrib/load_key_clevis_hook.sh
```diff
@@ -0,0 +1,189 @@
...
+        echo "$JWE" | clevis decrypt >"$KEYLOCATION"
+        cat "$KEYLOCATION" | zfs load-key -n $dataset_for_clevis_unlock
+        if [[ "$?" == "0" ]]; then
+          # We are fine
+          return 0
```
ahesford, line 113:
```suggestion
        echo "$JWE" | clevis decrypt >"${KEYFILE}"
       zfs load-key -L "${KEYFILE}" -n $dataset_for_clevis_unlock && return
```
I'm not thrilled about indiscriminately writing content to `$KEYFILE`. What if this conflicts with another key file or points to an important file in the ZFSBootMenu image itself? This could break things. You should probably work some magic like adding a unique prefix (as we do in `load-key`) and then unlock the pool with an explicit path override. Frankly, abusing the `keylocation` property here seems like a bad idea altogether.

ahesford (changes requested):
In principle, I'm not opposed to adding load-key hooks to allow some early setup, but there is a lot to think about when trying to generalize your single-purpose Clevis hook for a broader user base.

zfsbootmenu/lib/zfsbootmenu-core.sh
rdmitry0911: Got it

zfsbootmenu/lib/zfsbootmenu-core.sh
rdmitry0911: Got it

zfsbootmenu/lib/zfsbootmenu-core.sh
rdmitry0911: Agree

contrib/load_key_clevis_hook.sh
rdmitry0911: I would leave zdebug message before return as it provides useful information.

zdykstra: Warnings and errors should be logged via `zwarn` and `zerror` respectively, without a `Warning:` or `Error:` prefix.  `zdebug` logs are only logged if the user boots with `loglevel=7`, and are not really user-visible.

contrib/load_key_clevis_hook.sh
rdmitry0911: These are not error handlings. Those conditions are all valid. They just might happen. There is no need for the user to check where he/she is and what to do in this case. Resealing occures just once when ZBM, tpm2 and encrypted dataset are introduced to each other. After this introduction no resealing is required. The passphrase can be extracted from the hash stored in the special dataset property

ahesford: In that case, this definitely doesn't belong here. A one-shot initialization of the encryption scheme should happen long before ZFSBootMenu ever touches the pool.

contrib/load_key_clevis_hook.sh
rdmitry0911: Ok, got it

contrib/load_key_clevis_hook.sh
rdmitry0911: I put a check if the file exists to avoid damaging of an existing file and put a recomendation of naming the keyfiles in a description.

rdmitry0911: Without keylocation=file there is no way to boot a target system without asking a password. Just avoid conflict in filenames and we are fine

rdmitry0911: Done

ahesford: Sure there is; just override a non-existent location with `-L`. Also, there was a bug in my suggestion: it should be `-L "${KEYLOCATION}"` instead of `-L "${KEYFILE}"`.

rdmitry0911: Done

contrib/load_key_clevis_hook.sh
rdmitry0911: This is wrong. When zbm booter starts, it doesn't know how many times (if any) it started before on this host with this volume. There is no counter for that. So, we have to suppose, that this is not the first time as this is more likly. So we try to automatically unlock the dataset using this hook. And if during this unlocking process it turned out this is the first time we are running here, than only in this exclusive case we run resealing function. May be the name of this function is not completely relevant as it referes to something that happens quite often, while it is not true. It happens quite rare. Only when anyone of host hardware, ZBM or dataset has been changed since last boot. Suggest another name for the function, but it is sitting in a right place.

ahesford: I simply will not accept a script that operates as a load-key hook that does things like set properties on filesystems and, even worse, exports and re-imports the pool out from underneath the bootloader. It is potentially destructive, might unmount things that ZFSBootMenu has mounted and expected to be mounted, and doesn't even manage the import properly (all imports should use `-N` to suppress automount behavior).

Importing, mounting and exporting things in an `early_setup` hook are appropriate because ZFSBootMenu has not yet taken responsibility for any pools. At this stage in the execution, you should not be doing these manipulations. Whenever ZFSBootMenu is running, it should be considered the exclusive "owner" of the pool.

Further, as I noted before, these kinds of metadata setup operations are better suited to manual interactions with the user.

Your options at this point are to strip this logic from the script, or remove the script from this PR entirely so we can consider just the inclusion of hook support.

rdmitry0911: Ok, I'll remove the script if it contradicts with the policy. I first tried early_setup hook, but it is inconvenient. I wrote about that in discussions section. BTW, may be you know a safe way to change user options in load_key hook? May be importing a pool with -N would be enough for that?

README.md
rdmitry0911: Done

contrib/load_key_clevis_hook.sh
rdmitry0911: This function does not unlock the dataset it just seals the passphrase with the tpm2 device. If it fails, the script will continue as usual. Failure of this function means that next boot will not automatically decrypt the dataset. If user wants to manually store passphrase hash in user property, than it is better to do in standard recovery shell of ZBM and not form here.

contrib/load_key_clevis_hook.sh
rdmitry0911: If net config from OTB dracut or mkinitcpio network packages is used, then rebuilding of ZBM will be required each time It is installed on another host or in case of network changes. The way I'm using does not require rebuilding of ZBM. It makes it more universal.

contrib/load_key_clevis_hook.sh
rdmitry0911: I checked out several cases for importing/exporting pools using this hook with -N modifier for import. Everything works just fine. If you still insist on removing this hook from contrib, no problem, I'll do that. Just let me know your opinion.

rdmitry0911: I decided to delete file for pr to concentrate on load_key hook. We can negotiate this hook inclusion in contrib as a separate pr later

zfsbootmenu/lib/zfsbootmenu-core.sh
rdmitry0911: I can't commit your suggestion as I did it by myself

ahesford (changes requested):
You really shouldn't be doing these edits from the GitHub UI. You need to:
- [ ] Drop the edits to `contrib/README.md`
- [ ] Restore `README.md` to its original state. Your latest commit deleted the entire file.
- [ ] Add a section to `docs/man/zfsbootmenu.7.rst`, between the `zfsbootmenu_setup` and `zfsbootmenu_teardown` sections of the "Options for dracut" block, fully describing the hooks.
- [ ] Squash all of these commits down to one.

zfsbootmenu/help-files/92/zfsbootmenu.7.ansi
```diff
@@ -136,7 +136,7 @@
 
     [1mzbm.hookroot=<hookspec>[0m
 
-      Tell ZFSBootMenu to attempt to read any early-setup, setup or teardown hooks from the
+      Tell ZFSBootMenu to attempt to read any early-setup, setup, load_key or teardown hooks from the
```
zdykstra, line 139: You can't build this file by hand. This line is too long. Changes to `zfsbootmenu/help-files` need to be made by `releng/rst2help.sh`.

zfsbootmenu/help-files/132/zfsbootmenu.7.ansi
```diff
@@ -117,7 +117,7 @@
 
     [1mzbm.hookroot=<hookspec>[0m
 
-      Tell ZFSBootMenu to attempt to read any early-setup, setup or teardown hooks from the path specified by [33mhookspec[0m in addition
+      Tell ZFSBootMenu to attempt to read any early-setup, setup, load_key or teardown hooks from the path specified by [33mhookspec[0m in addition
```
zdykstra, line 120: This line is too long - don't edit this file by hand.

etc/zfsbootmenu/mkinitcpio.conf
```diff
@@ -94,6 +94,11 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard zfsbootmenu)
 # environment is about to be launched. Any writable ZFS pools will have been
 # exported, but read-only pools will generally still be available.
 #
+# 'zfsbootmenu_load_key' hooks are run immediately before attempting to load an
```
zdykstra, line 97: This text needs to be before the text for `zfsbootmenu_teardown`.

etc/zfsbootmenu/mkinitcpio.conf
```diff
@@ -94,6 +94,11 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard zfsbootmenu)
...
+# encryption key and unlocked a locked filesystem. Each hook is called with a
```
zdykstra, line 98: `s/unlocked/unlock`

zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1673,6 +1673,19 @@ load_key() {
     return 0
   fi
 
+  # Run early load_key hooks, if they exist
+  if [ -d /libexec/load_key.d ]; then
+    for _hook in /libexec/load_key.d/*; do
+      zinfo "Processing hook: ${_hook}"
+      [ -x "${_hook}" ] || continue
+      if ! "${_hook}" "${encroot}"; then
+        zwarn "load-key hook failed for '${encroot}', aborting load attempt"
```
zdykstra, line 1682: This warning should indicate the name of the hook that failed, otherwise the user is left guessing if multiple `load_key.d` hooks exist.

zdykstra (changes requested):

zfsbootmenu/lib/zfsbootmenu-core.sh
rdmitry0911: Agree. Put "${_hook}" into zwarn

zdykstra: This change has put a double-quoted argument inside another double-quoted argument, breaking it. 

Are any of these changes being tested prior to committing them? I'd like to see this PR merged, but the current process you're using is unacceptable. Look through all of the open conversations/review notes, fix everything addressed in them, and squash everything down to a single commit. Only when that's all done and you're confident that everything has been fully address, push to the PR so that there's a single commit ready for final review.

zfsbootmenu/help-files/92/zfsbootmenu.7.ansi
rdmitry0911: `releng/rst2help.sh` requires `rst2ansi` program. Where to get it?

etc/zfsbootmenu/mkinitcpio.conf
rdmitry0911: Done

etc/zfsbootmenu/mkinitcpio.conf
rdmitry0911: done

zfsbootmenu/lib/zfsbootmenu-core.sh
rdmitry0911: You right. It was my mistake with double-quoted argument and I fixed that. I've just tested the final version. It works. I can delete this pr and open a new one with all the commits squashed  to one. Is that what you have in mind?

ahesford: No, don't close a PR and open the same thing. Use `git rebase` or other git history rewriting tricks to collapse all of these edits and force-push back to this branch.

rdmitry0911: Sorry I'm not an experienced user of a github. This is my first pr. From my point of view all the issues and notes you addressed are implemented and there are no open discussions other then the one with the releng/rst2help.sh doc generator which is going to produce a bunch of a new files. As I've read in some articles, there is no way to rewrite the history in github. There is a possibility to create a new commits that will restore the previous state of the data by changing the current data to the old one. If this is true, then I have no idea how to collapse the topic, like this one? What should be the result of such a collapse? In my judgement, as far as 90% of these topics relates to the hook script that finally was excluded from the pr completely, it is not a bad idea to make a new pr with only few files in it and with very little changes in them. The hook itself is quite easy, isn't it? It would be easy to review these files and accept the pr.

classabbyamp: you probably shouldn't be using github's web editor to make these changes, but instead clone your fork, commit to the branch locally, and push. `git rebase` is a tool within git that allows you to rewrite the history. After rebasing, you can to `git push --force` to update the PR. There are a number of good tutorials on git concepts like rebasing online.

this is not a "github" thing but a git thing. github is just a place to collaborate using git.

docs/man/zfsbootmenu.7.rst
```diff
@@ -233,6 +233,12 @@ In addition to standard dracut configuration options, the ZFSBootMenu dracut mod
...
 
+**zfsbootmenu_load_key=<executable-list>**
+
+  An optional variable specifying a space-separated list of paths to setup hooks that will be installed in the ZFSBootMenu initramfs. Any path in the list **<executable-list>** that exists and is executable will be installed.
+
+  Any installed hooks are run right before the ZFSBootMenu menu loads key for an encrypted dataset; ZFS pools will generally have been imported and the default boot environment will be available in the *BOOTFS* environment variable. This hooks may provide additional methods of unlocking datasets, for example using clevis. The only argument passed to the hook is dataset name
```
ahesford, line 240: This text isn't correct.

```suggestion
  Any installed hooks are run immediately before ZFSBootMenu menu attempts to load a key for the encryption root of a locked filesystem. Each hook will be run with the name of the encryption root as its sole argument. If any hook exits with a non-zero return code, the load attempt will be aborted and the encryption root will not be unlocked. Hooks can be used to prepare keys needed to unlock the encryption root or may even attempt to unlock the root directly; a filesystem that is unlocked after the hooks have run will be considered to have had its keys successfully loaded. **Note:** The hooks may be run multiple times if keys for a given encryption root need to be loaded multiple times, e.g., by triggering a re-import of the pool. If a script should only run once, the script is responsible for keeping track of this.
```

ahesford (changes requested):

docs/man/zfsbootmenu.7.rst
```diff
@@ -233,6 +233,12 @@ In addition to standard dracut configuration options, the ZFSBootMenu dracut mod
 
   Any installed hooks are run right before the ZFSBootMenu menu will be presented; ZFS pools will generally have been imported and the default boot environment will be available in the *BOOTFS* environment variable. Hooks will not be run if the countdown timer expires (or was set to zero) and the default boot environment is automatically selected. **Note:** The hooks may be run multiple times if the menu is invoked multiple times, e.g., by dropping to an emergency shell and then returning to the menu. If a script should only run once, the script is responsible for keeping track of this.
 
+**zfsbootmenu_load_key=<executable-list>**
+
+  An optional variable specifying a space-separated list of paths to setup hooks that will be installed in the ZFSBootMenu initramfs. Any path in the list **<executable-list>** that exists and is executable will be installed.
```
ahesford, line 238: Missed this one...

```suggestion
  An optional variable specifying a space-separated list of paths to load-key hooks that will be installed in the ZFSBootMenu initramfs. Any path in the list **<executable-list>** that exists and is executable will be installed.
```

zfsbootmenu/help-files/132/zfsbootmenu.7.ansi
rdmitry0911: Ok, got it

zfsbootmenu/lib/zfsbootmenu-core.sh
rdmitry0911: Ok, finally I've managed to make a `git rebase` and  then`git push --force` Hope, the result is what you had in mind

zdykstra:
There should be a single commit in this PR. You've added previous commits to this.

zdykstra:
Feel free to re-open this PR once the acceptance criteria have been met.

PR #453, opened 2023-07-22
refactor remote access guide

classabbyamp:
- docs/guides/general: move mkinitcpio setup to its own guide
- docs/guides/general/remote-access: refactor for easier reading

hopefully this reads better than the old guide. made some stuff consistent between dracut and mkinitcpio to make organisation easier

### DIFF
--- a/docs/guides/general/mkinitcpio.rst
+++ b/docs/guides/general/mkinitcpio.rst
@@ -1,0 +1,44 @@
+Building with mkinitcpio
+========================
+
+ZFSBootMenu also supports the `mkinitcpio <https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/>`_ initramfs
+generator used by Arch Linux and available for Void Linux, but it must be configured first.
+
+Since `version 2.0.0 <https://github.com/zbm-dev/zfsbootmenu/releases/tag/v2.0.0>`_, ZFSBootMenu will install a standard
+:zbm:`mkinitcpio.conf <etc/zfsbootmenu/mkinitcpio.conf>` in the ``/etc/zfsbootmenu`` configuration directory. This file
+is generally the same as a standard ``mkinitcpio.conf``, except some additional declarations may be added to control
+aspects of the ``zfsbootmenu`` mkinitcpio module. The configuration file includes extensive inline documentation in the
+form of comments; configuration options specific to ZFSBootMenu are also described in the
+:ref:`zfsbootmenu(7) <zbm-mkinitcpio-options>` manual page.
+
+ZFSBootMenu still expects to use Dracut by default. To override this behavior and instead use mkinitcpio, edit
+``/etc/zfsbootmenu/config.yaml`` and add the following options:
+
+.. code-block:: yaml
+
+  Global:
+    InitCPIO: true
+    ## NOTE: The following three lines are OPTIONAL
+    InitCPIOHookDirs:
+      - /etc/zfsbootmenu/initcpio
+      - /usr/lib/initcpio
+
+.. note::
+
+  In some ZFSBootMenu guides, like :doc:`remote-access`, some mkinitcpio modules will be installed to
+  ``/etc/zfsbootmenu/initcpio`` to keep them isolated from system-installed modules. To accommodate this non-standard
+  installation, ``InitCPIOHookDirs`` must be defined in ``/etc/zfsbootmenu/config.yaml``. Furthermore, because
+  overriding the hook directory causes mkinitcpio to ignore its default module path, the default ``/usr/lib/initcpio``
+  must be manually specified. If all hooks are installed in ``/usr/lib/initcpio`` or ``/etc/initcpio``, the ZFSBootMenu
+  configuration does **not** need to specify ``InitCPIOHookDirs``.
+
+Without further changes, running ``generate-zbm`` should now produce a ZBM image based on mkinitcpio rather than Dracut.
+
+Whenever ``generate-zbm`` is run to generate images based on mkinitcpio, it forcefully adds the the required
+``zfsbootmenu`` hook after any hooks defined in the ``HOOKS`` array of ``/etc/zfsbootmenu/mkinitcpio.conf``. The default
+configuration file explicitly includes the ``zfsbootmenu`` hook in the array as a visual reminder that it will be
+included (strictly speaking, this will cause mkinitcpio to add the hook **twice**, but because the ``zfsbootmenu`` hook
+completely takes over execution of its initramfs image, it will only ever run once). If any custom configuration requires
+additional hooks be added to the ZFSBootMenu initramfs image, make sure that these hooks are included **before** any
+occurrence of ``zfsbootmenu`` in ``HOOKS``. Better still, just remove the ``zfsbootmenu`` hook from
+``/etc/zfsbootmenu/mkinitcpio.conf`` when making any edits to ``HOOKS`` to minimize the chance of configuration errors.
--- a/docs/guides/general/remote-access.rst
+++ b/docs/guides/general/remote-access.rst
@@ -11,47 +11,163 @@ your boot environments reside in encrypted filesystems, SSH access is necessary 
 when you are not physically present. Because ZFSBootMenu supports Dracut and mkinitcpio, any mechanism that can provide
 remote access to a Dracut or mkinitcpio initramfs will work.
 
-Dracut
-------
+Enabling Network Access
+-----------------------
 
-The `dracut-crypt-ssh <https://github.com/dracut-crypt-ssh/dracut-crypt-ssh>`_ provides a straightforward approach to
-configuring and launching an SSH server in Dracut images. The module is packaged in Void and does not rely on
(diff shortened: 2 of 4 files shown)


docs/guides/general/mkinitcpio.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Because further configuration may require additional mkinitcpio modules, and these must be run before the
+``zfsbootmenu`` module in the initramfs, edit ``/etc/zfsbootmenu/mkinitcpio.conf`` and **remove** any ``zfsbootmenu``
+entry in the ``HOOKS`` definition. As the standard configuration file notes, the ``zfsbootmenu`` module is required for
+ZFSBootMenu to function, but ``generate-zbm`` will forcefully add it at this at the end of the module list. Thus, the
+simplest way to ensure that additions to the ``HOOKS`` array occur *before* the ``zfsbootmenu`` module is to omit the
+latter from the configuration. The standard ``HOOKS`` line in ``/etc/zfsbootmenu/mkinitcpio.conf`` should therefore be
+something like::
+
+  HOOKS=(base udev autodetect modconf block filesystems keyboard)
```
classabbyamp, line 45: would it not make sense to leave `zfsbootmenu` out of `HOOKS` by default? or is there a use-case where it needs to be there?

ahesford: The only reason it's in there is to serve as a visual reminder to users that it will be put there.

docs/guides/general/remote-access.rst
```diff
@@ -11,303 +11,262 @@ your boot environments reside in encrypted filesystems, SSH access is necessary
...
+    To inform ``dracut-network`` that it must bring up a network interface, pass the kernel command-line parameters
+    ``ip=dhcp`` and ``rd.neednet=1`` to your ZFSBootMenu image. If you use another boot loader to start ZFSBootMenu, *e.g.*
+    rEFInd or syslinux, this can be accomplished by configuring that loader. However, it may be more convenient to add these
+    parameters directly to the ZFSBootMenu image::
 
-By default, ``dracut-crypt-ssh`` will generate random host keys for your ZFSBootMenu initramfs. This is undesirable
-because SSH will complain about unknown keys every time you reboot. If you wish, you can configure the module to copy
-your regular host keys into the image. However, there are two problems with this:
+      mkdir -p /etc/cmdline.d
+      echo "ip=dhcp rd.neednet=1" > /etc/cmdline.d/dracut-network.conf
```
zdykstra, line 55: Won't this add this to both the ZBM image and to the distro/BE image?

ahesford: I think these files need to be explicitly added to images, so they should be OK here.

docs/guides/general/mkinitcpio.rst
ahesford, line 45: Maybe something like this is better (but it probably needs more refinement):

```suggestion
Whenever ``generate-zbm`` is run to generate images based on mkinitcpio, it forcefully adds the the required ``zfsbootmenu`` hook after any hooks defined in the ``HOOKS`` array of ``/etc/zfsbootmenu/mkinitcpio.conf``. The default configuration file explicitly includes the ``zfsbootmenu`` hook in the array as a visual reminder that it will be included. (Strictly speaking, this will cause mkinitcpio to add the hook **twice**, but because the ``zfsbootmenu`` hook completely takes over execution of its initramfs image, it will only ever run once.) If any custom configuration requires additional hooks be added to the ZFSBootMenu initramfs image, make sure that these hooks are included **before** any occurrence of ``zfsbootmenu`` in ``HOOKS``. Better still, just remove the ``zfsbootmenu`` hook from ``/etc/zfsbootmenu/mkinitcpio.conf`` when making any edits to ``HOOKS`` to minimize the chance of configuration errors.
```

docs/guides/general/remote-access.rst
```diff
@@ -11,303 +11,262 @@ your boot environments reside in encrypted filesystems, SSH access is necessary
...
-   read your private host keys; and
+    It is possible to specify a static IP configuration by replacing ``dhcp`` with a properly formatted configuration
+    string. Consult the `dracut documentation <https://man.voidlinux.org/dracut.cmdline.7#Network>`_ for details about
+    static IP configuration.
 
-2. The ``dropbearconvert`` program may be incapable of converting modern OpenSSH host keys into the required dropbear
-   format.
+    There are methods besides writing to ``/etc/cmdline.d`` or configuring another boot loader to specify kernel
+    command-line arguments that will configure networking in Dracut. However, Dracut uses the ``/etc/cmdline.d`` directory
+    to store "fake" arguments, which it processes directly rather than handing to the kernel. In my tests, using other
```
ahesford, line 63:
```suggestion
    to store "fake" arguments, which it processes directly rather than handing to the kernel. Using other
```

docs/guides/general/remote-access.rst
```diff
@@ -11,303 +11,262 @@ your boot environments reside in encrypted filesystems, SSH access is necessary
...
+          ip addr add 192.168.1.2/24 brd + dev eth0
+          ip route add default via 192.168.1.1
 
-Basic Network Access
-~~~~~~~~~~~~~~~~~~~~
+          # Add some name servers
+          cat > /etc/resolv.conf <<-EOF
+            nameserver 1.1.1.1
+            nameserver 8.8.8.8
+            EOF
```
ahesford, line 135:
```suggestion
          cat > /etc/resolv.conf <<-EOF
          	nameserver 1.1.1.1
          	nameserver 8.8.8.8
          	EOF
```
Not sure if Sphinx will preserve a tab inside this block, but I think the shell will refuse to recognize the indented heredoc unless real tabs are used. That could bite people copying and pasting this block if we can't include tabs.

docs/guides/general/remote-access.rst
```diff
@@ -11,303 +11,262 @@ your boot environments reside in encrypted filesystems, SSH access is necessary
...
+      rm -r /tmp/mkinitcpio-dropbear-master
+
+    Then, enable the ``dropbear`` module in ``/etc/zfsbootmenu/mkinitcpio.conf`` by manually appending ``dropbear`` to
+    the ``HOOKS`` array, or by running::
+
+      sed -e '/HOOKS.*rclocal/a HOOKS+=(dropbear)' -i /etc/zfsbootmenu/mkinitcpio.conf
+
+    .. note::
+
+      The default configuration will start dropbear on TCP port 22. If using ahesford's fork of ``mkinitcpio-dropbear``,
```
ahesford, line 248:
```suggestion
      The default configuration will start dropbear on TCP port 22. If using the ``ahesford/mkinitcpio-dropbear`` fork recommended here,
```

ahesford (commented):
This is definitely an improvement over the current documentation.

docs/guides/general/remote-access.rst
classabbyamp: it's not preserved, I think we can just use a `<<` heredoc instead

docs/guides/general/mkinitcpio.rst
```diff
@@ -0,0 +1,44 @@
...
+  overriding the hook directory causes mkinitcpio to ignore its default module path, the default ``/usr/lib/initcpio``
+  must be manually specified. If all hooks are installed in ``/usr/lib/initcpio`` or ``/etc/initcpio``, the ZFSBootMenu
+  configuration does **not** need to specify ``InitCPIOHookDirs``.
+
+Without further changes, running ``generate-zbm`` should now produce a ZBM image based on mkinitcpio rather than Dracut.
+
+Whenever ``generate-zbm`` is run to generate images based on mkinitcpio, it forcefully adds the the required
+``zfsbootmenu`` hook after any hooks defined in the ``HOOKS`` array of ``/etc/zfsbootmenu/mkinitcpio.conf``. The default
+configuration file explicitly includes the ``zfsbootmenu`` hook in the array as a visual reminder that it will be
+included (Strictly speaking, this will cause mkinitcpio to add the hook **twice**, but because the ``zfsbootmenu`` hook
```
ahesford, line 40:
```suggestion
included (strictly speaking, this will cause mkinitcpio to add the hook **twice**, but because the ``zfsbootmenu`` hook
```

docs/guides/general/remote-access.rst
```diff
@@ -11,303 +11,262 @@ your boot environments reside in encrypted filesystems, SSH access is necessary
...
+      rm -r /tmp/mkinitcpio-dropbear-master
+
+    Then, enable the ``dropbear`` module in ``/etc/zfsbootmenu/mkinitcpio.conf`` by manually appending ``dropbear`` to
+    the ``HOOKS`` array, or by running::
+
+      sed -e '/HOOKS.*rclocal/a HOOKS+=(dropbear)' -i /etc/zfsbootmenu/mkinitcpio.conf
+
+    .. note::
+
+      The default configuration will start dropbear on TCP port 22. If using the ``ahesford/mkinitcpio-dropbear`` for
```
ahesford, line 248:
```suggestion
      The default configuration will start dropbear on TCP port 22. If using the ``ahesford/mkinitcpio-dropbear`` fork
```

ahesford (approved):
Looks good to me, after a couple of typographical fixes.

PR #455, opened 2023-07-24
Update remote-ssh-build.sh

emad-elsaid:


contrib/remote-ssh-build.sh
```diff
@@ -34,6 +34,7 @@
 # zfs change-key \
 #     -o keylocation=file:///etc/zfs/keys/rpool.key \
 #     -o keyformat=passphrase rpool
+# # add /etc/zfs/keys/rpool.key to the FILES array in /etc/mkinitcpio.conf
```
ahesford, line 37:
```suggestion
# sed -i '/FILES=/a FILES+=(/etc/zfs/keys/rpool.key) /etc/mkinitcpio.conf
```
Keeps this in the same spirit as other commands and allows people to copy and paste if they want.

ahesford (changes requested):

zdykstra:
Please use a descriptive commit message in the same format as previous commits.

emad-elsaid:
updated the commit message

PR #458, opened 2023-08-01
releng/docker/image-build.sh: REPO MADNESS

ahesford:
A new `-r` argument to `image-build.sh` allows the specification of (at most one) local directory that will be mounted to `/pkgs` when preparing the container image. This allows, *e.g.*, a local test version of ZFS just by specifying

    ./releng/docker/image-build.sh -r /path/to/host/repo zbm-builder:$(date '+%Y%m%d')

This isn't (yet?) exposed in `releng/make-binary.sh` or `releng/tag-release.sh`. On the one hand, this is a special-enough use case that we can require a manual build of the container image (because `make-binary.sh` will use an existing container image that matches the desired tag) to use the feature. On the other hand, I'm open to ideas on how to expose this up the chain.

While I was at it, I fixed the `-headers` cleanup so the container image won't hang on to `linux{5.15,6.1}-headers`.

### DIFF
--- a/releng/docker/image-build.sh
+++ b/releng/docker/image-build.sh
@@ -1,12 +1,22 @@
-#!/bin/sh
+#!/bin/bash
 # vim: softtabstop=2 shiftwidth=2 expandtab
 
 usage() {
   cat <<-EOF
-	USAGE: $0 [-h] [-p <pkg>] <tag> [zbm-commit-like]
+	USAGE: $0 [-h] [-r <path>] [-R <repo>] [-p <pkg>] <tag> [zbm-commit-like]
 	Create a container image for building ZFSBootMenu
 	
 	-h: Display this message and exit
+	
+	-r <path>:
+	   Mount the specified path into the build container when installing
+	   packages and configure XBPS to use the path as a package repository
+	
+	   (One path per argument; may be repeated as needed)
+	
+	-R <repo>:
+	   Configure XBPS to use the specified Void Linux package repository
+	   (One repository per argument; may be repeated as needed)
 	
 	-p <pkg>:
 	   Include the specified Void Linux package in the image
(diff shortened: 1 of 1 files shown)


classabbyamp:
would be nice if another http repo could be specified too

ahesford:
Any number of web repos can now be configured; only if none are specified will the old default (servercentral.com) be used. The local repo is still limited to one because I don't want to mess around with multiple binds at different locations in the container.

zdykstra (approved):

releng/docker/image-build.sh
```diff
@@ -1,13 +1,25 @@
...
 	
+	-r <repo>:
+	   Mount the specified host directory, which should be a valid XBPS
+	   repo, into the build container when installing packages
+	
+	-R <repo>:
+	   Configure XBPS to use the specified Void Linux package repository
+	
+	   The repo argument must be an absolute path (within the container),
+	   an HTTP or HTTPS URL, or "default" to specify the container default.
```
classabbyamp, line 19: don't think there's a need for this default workaround. any repo specified in the argv will take precedence over the one in the conf file, with the first-specified one having the highest priority. if you specify another mirror manually it will always take that over servercentral

classabbyamp: and similarly, `repository=` entries closer to the top of the list when `cat /etc/xbps.d/*.conf` will have priority

ahesford: The local repo is listed in a separate file that takes precedence, but the remote repos specified with `-R` are written separately to shadow the system default.

Previously, servercentral.com was the only supported repo, but the new logic allows that to be overridden with the builder's preferred remote. However, because users may prefer to augment rather than replace the default selection, the `default` magic word provides a more convenient way to reinstate that option (and even adjust the ordering of the remotes).

ahesford:
This has been completely rethought. Now, users can specify multiple `-r` arguments to mount multiple host directories as repositories that will be available when installing packages into the build-container image. The `-R` remote repositories are also written to a custom XBPS configuration file that should take precedence over the system default but will not mask it, meaning things like

    ./releng/docker/image-build.sh -r $XBPS_DISTDIR/hostdir/binpkgs/branch1 -r $XBPS_DISTDIR/hostdir/binpkgs/branch2 -R https://my-partial-repo.zfsbootmenu.org/current -R https://another-partial-repo.zfsbootmenu.org/current/ zbm-builder:tag

will just do the right thing, falling back to the default XBPS repo for any packages that can't be found elsewhere.

THAT'S RIGHT, WE FSCKING HAVE `-R`!!!!1

PR #459, opened 2023-08-02
docs/guides/: use feature flags when creationg pools

zdykstra:
Start creating pools with feature flags that match the maximum provided by ZFSBootMenu binary releases. This will possibly need some additional language on how to handle upgrades in the future.

### DIFF
--- a/docs/guides/_include/pool-creation.rst
+++ b/docs/guides/_include/pool-creation.rst
@@ -16,6 +16,7 @@ Create the zpool
        -O xattr=sa \
        -O relatime=on \
        -o autotrim=on \
+       -o compatibility=openzfs-2.1-linux \
        -m none zroot "$POOL_DEVICE"
 
   .. group-tab:: Encrypted
@@ -36,6 +37,7 @@ Create the zpool
        -O keylocation=file:///etc/zfs/zroot.key \
        -O keyformat=passphrase \
        -o autotrim=on \
+       -o compatibility=openzfs-2.1-linux \
        -m none zroot "$POOL_DEVICE"
 
     .. note::
@@ -51,3 +53,9 @@ Create the zpool
       * ``keyformat=passphrase`` - By setting the format to ``passphrase``, we can now force a prompt for this in
         ``zfsbootmenu``. It's critical that your passphrase be something you can type on your keyboard, since you will
         need to type it in to unlock the pool on boot.
+
+.. note::
+
+   The option ``-o compatibility=openzfs-2.1-linux`` ensures that the pool is created only with feature flags supported by the current ZFSBootMenu binary release. If you plan on building a custom ZFSBootMenu image that you will keep synchronized with your host, the compatibility option may be omitted.
+
+   Future releases of ZFSBootMenu may ship a newer version of ZFS which will support a new set of feature flags. Check release notes prior to updating your system.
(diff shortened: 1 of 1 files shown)


docs/guides/_include/pool-creation.rst
```diff
@@ -51,3 +53,7 @@ Create the zpool
       * ``keyformat=passphrase`` - By setting the format to ``passphrase``, we can now force a prompt for this in
         ``zfsbootmenu``. It's critical that your passphrase be something you can type on your keyboard, since you will
         need to type it in to unlock the pool on boot.
+
+.. note::
+
+  Pools are created with the feature flags supported by the current ZFSBootMenu binary (EFI) release. If your system uses a newer version of OpenZFS and you plan on building from source, ``-o compatibility=openzfs-2.1-linux`` can be omitted.
```
ahesford, line 59: What if this is part of the above note?

```suggestion

      * ``compatibility=openzfs-2.1-linux`` fixes the feature flags to those supported by the current ZFSBootMenu binary (EFI) release. This will prevent an errant ``zpool upgrade`` from rendering your system unbootable. If you use a newer OpenZFS and intend to keep custom ZFSBootMenu images sychronized with the host, you can omit this option.
```

ahesford: Also, if this is adopted, maybe change "However, the following options need to be addressed" on line 46 above to "However, the following options merit special attention"

zdykstra: The above note section only applies to encrypted pools. Moving it below the tab group ensures that it's shown for both encrypted and un-encrypted pools.

ahesford: This is why I get for trying to garner context from a patch. In that case I guess I'd revise the wording to:
```suggestion

.. note::

  The option ``-o compatibility=openzfs-2.1-linux`` ensures that the pool is created only with feature flags supported by the current ZFSBootMenu binary release. If you plan on building a custom ZFSBootMenu image that you will keep synchronized with your host, the compatibility option may be omitted.
```

docs/guides/_include/pool-creation.rst
```diff
@@ -51,3 +53,9 @@ Create the zpool
       * ``keyformat=passphrase`` - By setting the format to ``passphrase``, we can now force a prompt for this in
         ``zfsbootmenu``. It's critical that your passphrase be something you can type on your keyboard, since you will
         need to type it in to unlock the pool on boot.
+
+.. note::
+
+   The option ``-o compatibility=openzfs-2.1-linux`` ensures that the pool is created only with feature flags supported by the current ZFSBootMenu binary release. If you plan on building a custom ZFSBootMenu image that you will keep synchronized with your host, the compatibility option may be omitted.
+
+   Future releases of ZFSBootMenu may ship a newer version of ZFS which will support a new set of feature flags. Check release notes prior to updating your system.
```
ahesford, line 61:
```suggestion
   The option ``-o compatibility=openzfs-2.1-linux`` ensures that the pool is created only with feature flags supported by the current ZFSBootMenu binary release. If you plan on building a custom ZFSBootMenu image that you will keep synchronized with your host, the compatibility option may be omitted.

   Binary releases of ZFSBootMenu are generally built with the latest stable version of ZFS. Future release of ZFSBootMenu may therefore support newer feature sets. Check project release notes prior to updating or removing `compatibility` options and upgrading your system pool.
```

PR #461, opened 2023-08-09
zfsbootmenu: log errors for unimportable pools

zdykstra:
Before landing at the emergency_shell in the main import loop, log the output of a failed zpool import for every pool that's marked UNAVAIL .

In the emergency_shell function, write any log messages at the 'err' level directly to the active console.

https://asciinema.org/a/5QLE1dKHyeMDdmGQNmjM29yU0

### DIFF
--- a/zfsbootmenu/lib/kmsg-log-lib.sh
+++ b/zfsbootmenu/lib/kmsg-log-lib.sh
@@ -84,3 +84,58 @@ zerror() {
   : > "${BASE}/have_errors"
   echo "<3>ZFSBootMenu: $1" > /dev/kmsg
 }
+
+# arg1: comma-separated log levels to print
+# prints: all logs at that level
+# returns: nothing
+
+print_kmsg_logs() {
+  local levels levels_array grep_args
+
+  levels="${1}"
+  if [ -z "${levels}" ]; then
+    zerror "levels is undefined"
+    return
+  fi
+
+  # Try to use feature flags found in dmesg from util-linux
+  if output="$( dmesg --notime -f user --color=always -l "${levels}" 2>/dev/null )" ; then
+    echo -e "${output}"
+  else
+    # Both util-linux and Busybox dmesg support the -r flag. However, the log level that is
+    # reported by Busybox dmesg is larger than that reported by util-linux dmesg. Busybox dmesg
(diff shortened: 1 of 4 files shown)


zfsbootmenu/lib/kmsg-log-lib.sh
```diff
@@ -84,3 +84,49 @@ zerror() {
...
+print_kmsg_logs() {
+  local levels levels_array grep_args
+
+  levels="${1}"
+  if [ -z "${levels}" ]; then
+    zerror "levels is undefined"
+    return
+  fi
+
+  # Try to use feature flags found in dmesg from util-linux
```
zdykstra, line 101: There should probably be a better busybox vs util-linux dmesg detection routine ... somewhere.

zfsbootmenu/install-helpers.sh
```diff
@@ -100,6 +100,12 @@ create_zbm_conf() {
     has_column=1
   fi
 
+  local busybox_has_dmesg
+  dmesg="$( find "${BUILDROOT}" -name dmesg )"
```
zdykstra, line 104: This feels a bit hackish, but I don't want to assume it's always at `${BUILDROOT}/usr/bin/dmesg`, and we can't use `$PATH` right here.

zdykstra: I also don't want to do a generic "has busybox" test, because the modularity of busybox means that it might not ship dmesg.

zfsbootmenu/lib/kmsg-log-lib.sh
```diff
@@ -84,3 +84,52 @@ zerror() {
...
+  if output="$( dmesg --notime -f user --color=always -l "${levels}" 2>/dev/null )" ; then
+    echo -e "${output}"
+  # fall back to busybox if we have it
+  elif [ -n "${HAS_BUSYBOX_DMESG}" ]; then
```
ahesford, line 105: Is this test even necessary? It already attempts to use util-linux, so maybe this can just be an all-catching `else`. Worse case, maybe just change to
```bash
elif output="$( dmesg -r )"
[...]
```
and grep off `$output` after the level parsing.

PR #472, opened 2023-09-09
zfsbootmenu-core: improve kernel searches

ahesford:
- Remove obsolete default_kernel file
- Avoid reversing kernels list in select_kernel
- Avoid `ls` hack to find kernels in boot environment
- Make initramfs/kernel pairing agnostic of actual kernel path

This seems a little cleaner and can be generalized in the future to finding kernels at arbitrary paths.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -629,8 +629,9 @@ set_default_kernel() {
   set_rw_pool "${pool}" || return 1
   CLEAR_SCREEN=1 load_key "${fs}"
 
-  # Strip /boot/ to list only the file
-  kernel="${2#/boot/}"
+  # Strip leading /boot/ or / to list only the file
+  kernel="${2#/}"
+  kernel="${kernel#boot/}"
 
   # Restore nonspecific default when no kernel specified
   if [ -z "$kernel" ]; then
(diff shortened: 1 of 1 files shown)


zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -698,76 +698,71 @@ find_be_kernels() {
...
-    zdebug "no default kernel found for ${fs}"
-    rm -f "${kernel_records}" "${def_kernel_file}"
-    return 1
-  fi
-
-  zdebug "default kernel set to ${def_kernel}"
-  echo "${def_kernel##*/}" > "${def_kernel_file}"
-  return 0
+  # Remove an invalid kernel record if the search failed
+  zdebug "failed to find kernels on ${fs}"
```
zdykstra, line 763: Should this debug message instead be an error message? Not having any kernels on a filesystem that ZBM expects to find them on is probably worth yelling about.

zdykstra (approved):

ahesford:
1526bb85e99c0069dcc03ac336b49a164ff46c94

zfsbootmenu/lib/zfsbootmenu-core.sh
ahesford: Yes, an error should help leave some breadcrumbs if an expected boot environment is missing kernels. If somebody has a filesystem with `mountpoint=/` that deliberately contains no kernels and unnecessarily triggers the error, `org.zfsbootmenu:active=off` is the fix.

Merged with this change.

PR #473, opened 2023-09-10
zfsbootmenu-lib: show sort key where needed

zdykstra:
When fzf supports a border label (>= 0.35.0), show the current sort key that is being used to sort the list. This is only relevant on the main BE selection screen and the Snapshot selection screen. If fzf does not support a border label, fall back to legacy behavior of not showing the value anywhere.

https://asciinema.org/a/xfK6RvvJRGYLThSkMJgxhnJy0

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-lib.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-lib.sh
@@ -124,7 +124,7 @@ global_header() {
 # returns: 0 on successful selection, 1 if Esc was pressed, 130 if BE list is missing
 
 draw_be() {
-  local env selected header expects kcl_text kcl_bind blank
+  local env selected header expects kcl_text kcl_bind blank sort_key preview_label
 
   env="${1}"
   if [ -z "${env}" ]; then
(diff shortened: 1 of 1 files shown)


zfsbootmenu/lib/zfsbootmenu-lib.sh
```diff
@@ -251,7 +254,34 @@ draw_snapshots() {
...
+  note="Note: for diff viewer, use tab to select/deselect up to two items"
+
+  local LEGACY_CONTEXT
+  if [ -n "${HAS_BORDER}" ] ; then
+
+    # Determine how much space should be between the 'sorted by' text and a centered note
+    # Remove 4 extra characters so that we can put a 1 character pad between strings and 
+    # the horizontal box line
+
+    spacer=$(( ( ( COLUMNS - ${#note} ) / 2 ) - ${#sorted_by} - 4 ))
```
ahesford, line 269: Logic suggestion:

1. If `spacer` (should this be `local`?) is positive, print both sorted key and note as is done here.
2. Otherwise, if `$COLUMNS` is larger than the length of the note (plus some pad), just print the note.
3. Otherwise, if `$COLUMNS` is larger than the length of the sort key, just print the sort key.
4. Otherwise, omit context altogether.

zfsbootmenu/lib/zfsbootmenu-lib.sh
```diff
@@ -157,12 +157,17 @@ draw_be() {
 ^[CTRL+R] recovery shell
 ^[CTRL+H] help" )"
 
+  sort_key="$( get_sort_key )"
+  context="Sorted by: ${sort_key^}"
```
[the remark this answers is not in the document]
zdykstra: yup

zfsbootmenu/lib/zfsbootmenu-lib.sh
zdykstra: I've implemented that. I had thought of handling small screen sizes and then I got distracted.

ahesford (approved):

PR #474, opened 2023-09-11
zfsbootmenu: refactor installation, support system and user runtime hooks

ahesford:
- All core components, docs and runtime hooks are handled by a common install_helper that is aware of both dracut and mkinitcpio methods

- Installation now adds "system" runtime hooks in addition to those configured by the user; user hooks with names that conflict with system hooks will "mask" the system version

- The hooks `10-console-init.sh` and `xhci-teardown.sh` have been elevated to "system" status and are always installed in ZBM images

- The `zfsbootmenu/hook` path has been renamed to `zfsbootmenu/pre-init` to avoid confusion with the new `zfsbootmenu/hooks` directory of system runtime hooks, and the dracut-only `hook` scripts have been moved to the dracut module

- A new `zbm.skip_hooks` command-line argument allows a comma-separated list of hooks to skip (matched on basename only) to provide an easy mechanism for skipping default hooks without `zbm.hookroot`

- Create the `zfsbootmenu-run-hooks` libexec helper, allowing easy addition of future hook points.


zdykstra:
I've tested this in various forms locally, everything looks good to me.

PR #477, opened 2023-09-12
zfsbootmenu-core: add load-key and boot-environment hooks

ahesford:


zdykstra:
Adding a few more hook points shouldn't be a problem, the bare hook loop is reasonably quick. We spend more time sourcing `/etc/zfsbootmenu.conf` then anything else.

zdykstra (approved):

midzelis:
Thanks!

PR #478, opened 2023-09-12
releng/docker: Dockerfile parity with image-build.sh

midzelis:
Update the existing Dockerfile to match the latest from image-build.sh

### DIFF
--- a/releng/docker/Dockerfile
+++ b/releng/docker/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1.4
+#
 # This Dockerfile creates a container that will create an EFI executable and
 # separate kernel/initramfs components from a ZFSBootMenu repository. The
 # container will pre-populate its /zbm directory with a clone of the master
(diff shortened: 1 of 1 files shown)


releng/docker/Dockerfile
```diff
@@ -7,34 +9,100 @@
...
 # Use the official Void Linux container
-FROM ghcr.io/void-linux/void-glibc-full:latest
+FROM ghcr.io/void-linux/void-glibc-full 
+LABEL org.opencontainers.image.authors="ZFSBootMenu Team, https://zfsbootmenu.org"
+
+# 	   ZFSBootMenu commit hash, tag or branch name used by
+#	   default to build ZFSBootMenu images (default: master)
 ARG ZBM_COMMIT_HASH
+#      Include the specified Void Linux package in the image
+#
```
classabbyamp, line 19:
```suggestion
#      Include the specified Void Linux packages in the image
```

releng/docker/Dockerfile
```diff
@@ -7,34 +9,100 @@
...
+done
+EOF
 
 # Ensure everything is up-to-date
-RUN xbps-install -Suy xbps && xbps-install -uy
+RUN xbps-install -Suy xbps && xbps-install -uy bash
+SHELL ["/bin/bash", "-c"]
+
+# Install custom packages
+RUN <<-EOF
```
classabbyamp, line 44: `-` in heredoc syntax should only be necessary if the base indent is at least 1 tab, so i'd either remote it or put a tab indent (and do the same for all heredocs)

releng/docker/Dockerfile
```diff
@@ -7,34 +9,100 @@
...
 
 # Copy the build script
 COPY build-init.sh /build-init.sh
+RUN chmod 755 /build-init.sh
 
+WORKDIR /
 # Run the build script with no arguments by default
 ENTRYPOINT [ "/build-init.sh" ]
-CMD [ ]
+CMD [ ]
```
classabbyamp, line 108: missing newline at the end

classabbyamp (commented):
nit: can all the indents of the comments around the `ARG`s be the same?

releng/docker/build-init.sh
```diff
@@ -109,6 +109,15 @@ if [ "${#PACKAGES[@]}" -gt 0 ]; then
   xbps-install -Sy "${PACKAGES[@]}"
 fi
 
+if [ -d /alt ]; then
```
midzelis, line 112: When bind mounting NFS shares into containers, some initramfs generators complain that they can't copy files because the file source (NFS) has additional permissions that can not be represented in the tmpfs based filesystems within the container. The workaround is to create an alterative root, and anything that is mounted here will be copied to /.

classabbyamp: can we leave this PR at the original scope?

ahesford: It's not clear what should be mounted at `/alt`, it isn't documented anywhere but this comment, and is outside the scope of this PR.

ahesford (commented):
Personally, I'd prefer every RUN heredoc be indented one level to it's immediately clear which commands are lumped together.

Using `<<-` instead of `<<` should really only be necessary if the `EOF` marker is indented along with the body (which might be desirable anyway), because Bash won't care about leading indentation that isn't stripped.

releng/docker/Dockerfile
```diff
@@ -7,34 +9,99 @@
...
+fi
+EOF
 
 # Prefer an LTS version over whatever Void thinks is current
-RUN echo "ignorepkg=linux" > /etc/xbps.d/10-nolinux.conf \
-	&& echo "ignorepkg=linux-headers" >> /etc/xbps.d/10-nolinux.conf
+RUN <<EOF
+echo "ignorepkg=linux" >> /etc/xbps.d/10-nolinux.conf
+echo "ignorepkg=linux-headers" >> /etc/xbps.d/10-nolinux.conf
+EOF
```
ahesford, line 54:
```suggestion
COPY <<-EOF /etc/xbps.d/10-nolinux.conf
	ignorepkg=linux
	ignorepkg=linux-headers
	EOF
```

releng/docker/Dockerfile
```diff
@@ -7,34 +9,99 @@
...
-	zfs gummiboot-efistub curl yq-go bash kbd \
-	terminus-font dracut mkinitcpio cryptsetup
+RUN <<EOF
+if [ -n "${KERNELS}" ]; then
+	for _kern in "${KERNELS}"; do
+		kern_headers+=( "${_kern}-headers" )
+	done
+	echo Installing Kernels: ${KERNELS} ${kern_headers}
+	xbps-install -y ${KERNELS} ${kern_headers}
+fi
```
ahesford, line 71: If `$KERNELS` is empty, it defeats the whole purpose of building the container. This probably doesn't need to be wrapped in a test, although it would be nice to fail the build up front if $KERNELS is unspecified.

Also, `for _kern in "${KERNELS}"` is probably wrong; in regular shell, it would treat `linux6.1 linux5.10 linux5.15` as a single argument instead of three. Unless the build process is running this in some non-standard shell-like environment, it must not be quoted.

releng/docker/Dockerfile
```diff
@@ -7,34 +9,99 @@
...
-RUN xbps-remove -Roy linux5.10-headers dkms && rm -f /var/cache/xbps/*
+RUN xbps-pkgdb -m manual binutils
+
+RUN <<EOF
+if [ -n "${KERNELS}" ]; then
+	for _kern in "${KERNELS}"; do
+		kern_headers+=( "${_kern}-headers" )
+	done
+	xbps-remove -Roy ${kern_headers}
+fi
```
ahesford, line 88: Same as above: `$KERNELS` probably shouldn't be quoted in the loop, and testing that it's non-empty doesn't make sense.

releng/docker/Dockerfile
```diff
@@ -7,34 +9,86 @@
...
 
 # Use the official Void Linux container
-FROM ghcr.io/void-linux/void-glibc-full:latest
+FROM ghcr.io/void-linux/void-glibc-full 
+LABEL org.opencontainers.image.authors="ZFSBootMenu Team, https://zfsbootmenu.org"
+# ZFSBootMenu commit hash, tag or branch name used by
+# default to build ZFSBootMenu images (default: master)
 ARG ZBM_COMMIT_HASH
+# Include the specified Void Linux package in the image
+#
```
ahesford, line 18:
```suggestion

# Include the specified Void Linux package in the image
```

releng/docker/Dockerfile
```diff
@@ -7,34 +9,86 @@
...
+# (multiple entries must be seperated by spaces)
+ARG PACKAGES=
+ARG XBPS_REPOS="https://repo-fastly.voidlinux.org/current"
+# Include the specified Void Linux kernel series in the image; kernel
+# series take the form <major>.<minor> and correspond to Void packages
+# linux<kver> and linux<kver>-headers
+#
+# Default: install 5.10, 5.15 and 6.1
+#	
+# (multiple entries must be seperated by spaces)
```
ahesford, line 28:
```suggestion

# Include the specified Void Linux kernel series in the image; kernel
# series take the form linux<major>.<minor>
# (multiple entries must be seperated by spaces)
#
# Default: linux5.10 linux5.15 linux6.1
```

releng/docker/Dockerfile
```diff
@@ -7,34 +9,86 @@
...
-# Install components necessary to build the image
-RUN xbps-query -Rp run_depends zfsbootmenu | xargs xbps-install -y
-RUN xbps-install -y linux5.10 linux5.10-headers \
-	zfs gummiboot-efistub curl yq-go bash kbd \
-	terminus-font dracut mkinitcpio cryptsetup
+	# Prevent initramfs kernel hooks from being installed
+	cat > /etc/xbps.d/15-noinitramfs.conf <<-CAT
+		noextract=/usr/libexec/mkinitcpio/*
+		noextract=/usr/libexec/dracut/*
+	CAT
```
ahesford, line 61: Nested heredocs are more difficult to read. Is there a reason this is preferred over `COPY <<-EOF` syntax I see in the documentation?

releng/docker/Dockerfile
```diff
@@ -7,34 +9,86 @@
...
+	if [ -n "${PACKAGES}" ]; then
+		echo "INFO: adding extra Void packages: ${PACKAGES}"
+		xbps-install -y $PACKAGES
+	fi
+	EOF
 
-# Ensure everything is up-to-date
-RUN xbps-install -Suy xbps && xbps-install -uy
+# Switch to bash
+SHELL ["/bin/bash", "-c"]
```
ahesford, line 48: I believe the only purpose where `bash` is needed in this file is in the handling of the kernel-headers array, but there isn't a lot of value in handling that as an array when `$KERNELS` isn't an array, so this can be dropped.
```suggestion
```

releng/docker/Dockerfile
```diff
@@ -7,34 +9,86 @@
...
-# provides objcopy, which is necessary for UEFI bundle creation
-RUN xbps-pkgdb -m manual binutils
-RUN echo "ignorepkg=dkms" > /etc/xbps.d/10-nodkms.conf
-RUN xbps-remove -Roy linux5.10-headers dkms && rm -f /var/cache/xbps/*
+	# Install ZFSBootMenu dependencies and components necessary to build images
+	xbps-query -Rp run_depends zfsbootmenu | xargs xbps-install -y
+	
+	for _kern in ${KERNELS}; do
+		kern_headers+=( "${_kern}-headers" )
+	done
```
ahesford, line 68:
```suggestion
	kern_headers=""
	for _kern in ${KERNELS}; do
		kern_headers="${kern_headers} ${_kern}-headers"
	done
```

ahesford (changes requested):

ahesford (changes requested):
Please squash these commits and make some tweaks so that I can test the new dockerfile.

ahesford:
Sorry, somehow I missed the notification that there were changes. Merged with a few minor edits.

PR #479, opened 2023-09-13
zfsbootmenu-core: conditionally disable spl_hostid warnings

zdykstra:
The default behavior of hostid matching is useful to have on in binary releases. However, it generates a warning message that isn't really relevant for users of that build.

It's expected that the spl_hostid value in ZFSBootMenu.EFI won't match the local pool hostid - advising these users to manually set `spl_hostid` on the ZBM KCL runs counter to how we recommend ZBM should be used.

### DIFF
--- a/zfsbootmenu/install-helpers.sh
+++ b/zfsbootmenu/install-helpers.sh
@@ -100,6 +100,15 @@ create_zbm_conf() {
     has_column=1
   fi
 
+  case "${zfsbootmenu_release_build,,}" in
+    yes|y|on|1)
+      zfsbootmenu_release_build=1
+      ;;
+    *)
+      zfsbootmenu_release_build=
+      ;;
+  esac
+
   # Normalize ZBM_BUILDSTYLE, if set
   case "${ZBM_BUILDSTYLE,,}" in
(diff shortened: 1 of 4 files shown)


zfsbootmenu/install-helpers.sh
```diff
@@ -113,12 +113,14 @@ create_zbm_conf() {
...
 	EOF
 
+  # shellcheck disable=SC2154
   cat >> "${BUILDROOT}/etc/zfsbootmenu.conf" <<-EOF
 	export HAS_REFRESH="${has_refresh}"
 	export HAS_DISABLED="${has_disabled}"
 	export HAS_BORDER="${has_border}"
 	export HAS_COLUMN="${has_column}"
 	export ZBM_BUILDSTYLE="${ZBM_BUILDSTYLE}"
+	export ZBM_RELEASE_BUILD="${release_build}"
```
ahesford, line 123:
```suggestion
	export ZBM_RELEASE_BUILD="${zfsbootmenu_release_build}"
```
Then you can just set that argument directly in the modules. Also, maybe normalize the value before writing it, so that `${zfsbootmenu_release_build,,}` maps `y|yes|on|1|t|true` or similar to `yes`.

PR #480, opened 2023-09-13
generate-zbm: Option (--monitor|m) to stream initramfs generator output to terminal

midzelis:


zdykstra (approved):

ahesford:
I'd be OK with a `--verbose` flag that dumps output from the initramfs generator to the console even without `--debug`, but passing stdin to the program so you can use `generate-zbm` as an interactive debugger is a bridge too far.

ahesford:
A further issue I have with attempting to provide interactivity via open3 is that neither dracut nor mkinitcpio rely on interactivity in their normal configurations, so this feature is only ever useful if one hacks some interactive shells into their module installation routines.

We'd be supporting a feature that will essentially only ever be used by you.

midzelis:
Interactive debugger is not the purpose. Its really nice to see the output. You will see exactly which files are being included when it builds, and how much progress is being made. Additionally, you see all the nice colors/formatting when inherting stdio, since to the process, it appears as a full tty instead of a redirection.

midzelis:
If it helps, I could just pass undef as the STDIN, and leave the STDOUT/STDERR redirected. That way no input would ever get  sent to the generator.

midzelis:
Additionally, there are a lot of warnings (and other really useful information) that are printed by the generators, that would otherwise not be seen at all, since the overall process is not failed, and so, will not be logged.

ahesford:
Extra output is fine, but in that case, verbose output can be accomplished just by calling `print` instead of `Log` on the `@output` returned by the standard `execute` subroutine whenever `generate-zbm --verbose` is called. Adding a second `executeStdio` just to dump output just muddies the water.

midzelis:
Closing as I no longer need this

PR #481, opened 2023-09-13
Docker Image: Support NFS bind mounts for build inputs

midzelis:


ahesford:
Are the errors you see fatal, or are they just noise that still ends with a functional initramfs image?

Copying from `/alt` is a lot of non-obvious magic to solve a problem that is probably extremely rare even if it is fatal. It should also be possible to work around this problem with some manipulation in `rc.d`, without exposing additional complexity to all other users.

midzelis:
The errors are fatal, no image is produced. I could not work around. I looked at the source code for dracut and instead of using normal 'cp' command, they have their custom .c binary that copies with permissions - there are no cli options available that can prevent copying permissions.  

It is not possible to workaround this problem in /rc.d, as the /rd.c processing happens too late in the build-init lifecyle. Another way to go is to add a /rc.start.d which runs as soon as the options are processed.

ahesford:
What is the workflow that causes the problem? What are you bind-mounting from NFS?

midzelis:
Well, all my data is stored on ZFS (obviously) - but hosted natively by proxmox. But proxmox really shouldn't be used for anything else except hosting VMs and LXCs. So, I have a bunch of VMs and LXCs where I do most of my dev work. To get the data from proxmox's ZFS to my VMs, I use NFS. 

All of my git repos are accessed via NFS share. I guess I could try SMB to see if its helps. But, I'm also moving away from Dracut which had the problem - if mkinitcpio doesn't have that issue, then I can close this too. 

But, if we change this request from doing the `/alt` copy to just be a startup hook inside the builder, that seems like a small ask and I hope you'd be open to it.

ahesford:
Running hooks from `${BUILDROOT}/rc.pre.d` right after all packages are installed in the container and before it attempts to verify or populate `/zbm` would not be objectionable.

Issue #432, opened 2023-06-14
[Feature suggestion] Inject ZFSBootMenu's interactively prompted passphrase into appended initramfs


chenxiaolong:
The Linux kernel supports loading multiple initramfs archives as long as each part is aligned to a 4-byte boundary. Each part is extracted in order to the initial `/` rootfs. I believe this has been supported since very early kernel versions. Debian, for example, uses this method to add non-free firmware to its netboot initramfs images.

It would be cool if ZFSBootMenu could use this mechanism to provide the dataset encryption passphrase to the OS. This would allow the user to only input the passphrase once and not require it to be stored anywhere on disk.

I made a very rough proof of concept implementation of this:

<details>
<summary>(click to expand)</summary>

```diff
diff --git a/zfsbootmenu/install-helpers.sh b/zfsbootmenu/install-helpers.sh
index 7ac4a72..e10ae93 100644
--- a/zfsbootmenu/install-helpers.sh
+++ b/zfsbootmenu/install-helpers.sh
@@ -38,6 +38,9 @@ zfsbootmenu_essential_binaries=(
   "fzf"
   "setsid"
   "cat"
+  "find"
+  "cpio"
+  "truncate"
 )
 
 # shellcheck disable=SC2034
diff --git a/zfsbootmenu/lib/zfsbootmenu-core.sh b/zfsbootmenu/lib/zfsbootmenu-core.sh
index 65ef783..732d277 100755
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -272,7 +272,7 @@ mount_zfs() {
 # returns: 1 on error, otherwise does not return
 
 kexec_kernel() {
-  local selected fs kernel initramfs tdhook output
+  local selected fs kernel initramfs tdhook output initramfs_size initramfs_full
 
   selected="${1}"
   if [ -z "${selected}" ]; then
@@ -300,8 +300,26 @@ kexec_kernel() {
   cli_args="$( load_be_cmdline "${fs}" )"
   root_prefix="$( find_root_prefix "${fs}" "${mnt}" )"
 
+  initramfs_full="${mnt}${initramfs}"
+
+  if [[ -d /zbm.keys ]]; then
+    # Append key files created from prompted passphrases and inject them into
+    # the initramfs. The appended archive must be aligned to a 4-byte boundary
+    # for the kernel to reliably detect it:
[...]

PR #486, opened 2023-09-15
zfsbootmenu-core: run boot environment hooks after environment mount

agorgl:
Moving boot environment hooks after environment mount will increase the usefulness of the new `boot-sel.d` hooks.
The hooks will not only be able to interact with the newly mounted environment but also shadow files (like the initramfs) in it using bind mounts, thus allowing techniques like the one described in #432 to be implemented.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -290,6 +290,14 @@ kexec_kernel() {
 
   CLEAR_SCREEN=1 load_key "${fs}"
 
+  tput cnorm
+  tput clear
+
+  if ! mnt=$( mount_zfs "${fs}" ); then
+    emergency_shell "unable to mount $( colorize cyan "${fs}" )"
+    return 1
+  fi
+
   # Variables to tell user hooks what BE has been selected
   hook_envs=(
     ZBM_SELECTED_BE="${fs}"
@@ -299,14 +307,6 @@ kexec_kernel() {
 
   # Run boot-environment hooks, if they exist
   env "${hook_envs[@]}" /libexec/zfsbootmenu-run-hooks "boot-sel.d"
-
-  tput cnorm
-  tput clear
-
-  if ! mnt=$( mount_zfs "${fs}" ); then
-    emergency_shell "unable to mount $( colorize cyan "${fs}" )"
-    return 1
-  fi
 
   cli_args="$( load_be_cmdline "${fs}" )"
   root_prefix="$( find_root_prefix "${fs}" "${mnt}" )"


ahesford:
When adding this hook point, I went back and forth on whether to run these hooks before or after the filesystem was mounted. Eventually, I decided they should run in a relatively pristine environment without an active mount.

The major concern here is that, after the `kexec` load, the `kexec_kernel` function unmounts the filesystem and also exports any writable pools to make sure the reboot is clean. Your bind-mount approach is clever, but any mounts atop the ZFS filesystem would prevent the unmount. We can't even rely on `umount -R` because everything should work with busybox whe mkinitcpio is used, so that complicates the unmount logic quite a bit just to support this very special case.

I'm aware of the desire to dynamically rewrite the initramfs, which was the main inspiration for adding the `boot-sel.d` hooks in the first place. Where they are currently run, the solution for those wishing to alter what will be booted is to design a script that essentially replicates the rest of `kexec_kernel`, eventually ending in a `kexec -e` that will never return control to ZFSBootMenu. This might not be ideal, but the other alternative is potentially breaking the cleanup part of `kexec_kernel` and having to replicate much of that *anyway*, but as a teardown hook.

Do have have any defense against these issues?

agorgl:
Thanks for the thorough explanation.

My main defense on hooks in general is that they exist to provide extra/divergent functionality from what vanilla zbm offers, and to do that they need access to more stuff or be able modify parts of the boot process. Without access to these, I cannot see many cases that `boot-sel.d` hooks would be useful in general, as I believe their whole point could be to allow to modify the kernel loading part of the process.

In any case, refactoring the `kexec_kernel` function to three separate functions `kexec_load`, `kexec_cleanup` and `kexec_run` could partially make this more flexible. One, as you said, could write a hook that never returns control to zbm, but also rely on the predefined functionality without copy-pasting the rest of the `kexec_kernel` function and constantly making sure that is in sync with the upstream.

ahesford:
Thanks for your contribution.

After thinking about it some more, I accepted your changes with two additions:
1. A `ZBM_SELECTED_MOUNTPOINT` environment variable is exposed to the hooks, so they know where to find the kernel and initramfs that will be loaded;
2. The documentation reflects this.

In your application, I recommend against bind-mounting over the ZFS filesystem, because it will break our unmount and cleanup prior to boot. Instead, I suggest you:
1. Copy the kernel and initramfs to `/tmp` or some other suitable location.
2. *Unmount* the ZFS filesystem at `${ZBM_SELECTED_MOUNTPOINT}`
3. Mount a new tmpfs filesystem in its place.
4. Move the kernel back to `${ZBM_SELECTED_MOUNTPOINT}/${ZBM_SELECTED_KERNEL}` and do whatever you need to the initramfs before putting it back at `${ZBM_SELECTED_MOUNTPOINT}/${ZBM_SELECTED_KERNEL}`.

At this point, ZFSBootMenu will happily grab the kernel and initramfs, it will then unmount the tmpfs without issue and proceed with the rest of its cleanup and boot process.

PR #487, opened 2023-09-18
zfsbootmenu-core: remove SIGINT trap when spawning e-shell

midzelis:
After emergency_shell() is called, no commands can be interrupted (ctrl-c). 

reproduce:
1) go to e-shell
2) execute `sleep 5`
3) press ctrl-c
4) sleep is not interrupted

fix
untrap SIGINT before launching e-shell

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -1896,7 +1896,7 @@ emergency_shell() {
   command -v efibootmgr >/dev/null 2>&1 && mount_efivarfs "rw" 
 
   # -i (interactive) mode will source $HOME/.bashrc
-  /bin/bash -i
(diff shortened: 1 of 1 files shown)


zdykstra:
The trap needs to be restored after `/bin/bash` exits

zdykstra (changes requested):
The trap needs to be restored after /bin/bash exits

ahesford:
It's probably best to just
```sh
( trap - SIGINT; exec /bin/bash -i )
```
to avoid having to restore it. (Assuming this works as expected.)

ahesford (approved):
The general concept works in a simple test script, but I haven't undertaken any tests in ZFSBootMenu itself.

zdykstra (changes requested):
As near as I can tell this is working as expected. Squash the commits and it's good to go.

PR #488, opened 2023-09-19
zfsbootmenu-core: optimize initramfs search

ahesford:
Walking through a four-level loop to find all possible initramfs images can be painful, and most times the distribution default will be in place. Let's front-load the initramfs search with commonly expected formats for popular distributions and avoid some of the pain of the complex but more exhaustive loop.


zdykstra (approved):
https://somebits.link/u/p/bc852bb11b/flamechart.svg

Nice improvement for both the normal cases the pathological case. Readable, and a net gain in speed.

PR #493, opened 2023-09-20
zfsbootmenu: rename UI library, normalize imports

zdykstra:
Rename zfsbootmenu-lib to zfsbootmenu-ui-lib to more accurately represent which functions are contained therein.

Additionally, normalize how libraries are sourced.


ahesford (approved):
This is fine by me, or it could be shortened to `zfsbootmenu-ui.sh`.

ahesford (commented):
Actually, the include guard should probably be renamed to stay consistent with the library name.

PR #494, opened 2023-09-22
zfsbootmenu: add sysrq-based reboot

zdykstra:
Not much. Super basic reboot command. I'm open to a better way of doing this.


zfsbootmenu/bin/reboot
```diff
@@ -0,0 +1,23 @@
...
+  fi
+done
+
+unset src sources
+
+while read -r _pool; do
+  is_writable "${_pool}" && export_pool "${_pool}"
+done <<<"$( zpool list -H -o name )"
+
+[ -w /proc/sysrq-trigger ] && echo b > /proc/sysrq-trigger
```
ahesford, line 23: If this file doesn't exist or isn't writable, this should print a message instructing the user to hard-reset the system because a soft reboot is not supported.

zdykstra: Good idea!

PR #506, opened 2023-10-22
Vendor the Gummiboot UEFI stub loader

ahesford:
We need to update `generate-zbm` to point to the vendored location before looking for gummiboot or systemd-boot versions, but I'm thinking we should actually look in seven places for the vendored file:
```
./zfsbootmenu/stub
../zfsbootmenu/stub
../share/zfsbootmenu/stub
dirname(__FILE__)/zfsbootmenu/stub
dirname(__FILE__)/../zfsbootmenu/stub
dirname(__FILE__)/../share/zfsbootmenu/stub
/usr/share/zfsbootmenu/stub
```
That covers common cases like running `generate-zbm` directly from the repo, running it in some installed prefix and (maybe) running in our build container. (This needs to be tested.) I'm looking for some suggestions on this search.

Once we have the desired behavior, we should stop installing `gummiboot-efistub` in the build container.


zdykstra:
Before this is vendored, I'd like to check kernel configs for Debian, Arch and maybe Gentoo to confirm that they have the right EFI handoff option enabled. If they don't, this won't get us very far.

zdykstra:
Arch's 6.3.x kernel ships with `CONFIG_EFI_HANDOVER_PROTOCOL` enabled. Ubuntu 23.10's 6.5.x kernel ships with it enabled.

ahesford:
Because `generate-zbm` is now more intelligent about writing EFI bundles, this is less relevant.

PR #514, opened 2023-10-28
generate-zbm: remove integrated syslinux support

zdykstra:
This has been deprecated in master since a3f15ed and in a release since v2.1.0 . It's time to cut it out ahead of v2.3.0. I think I caught all the references to it, but this could use another set of eyes.


ahesford:
We refer to "now-deprecated" support in the Void syslinux guide, although the instructions document the use of a static configuration. We should scrub the sentences that refer to the built-in capabilities in that guide.

zdykstra:
> We refer to "now-deprecated" support in the Void syslinux guide, although the instructions document the use of a static configuration. We should scrub the sentences that refer to the built-in capabilities in that guide.

Removed. I added a brief reference to the syslinux configuration generation hook script as well.

PR #532, opened 2023-11-13
contrib/zbm-sign.pl: respect EFI.ImageDir

Johnnynator:


zdykstra:
Nice quality of life improvement - thanks!

PR #538, opened 2023-11-27
Use zbm-release file, add some color to zreport

zdykstra:
What it says on the tin. Make the system report easier to access, add some color/spice to the output. Since everything is generated live, there's no "clean" line wrapping. The fzf preview window does at least wrap long lines, however.


ahesford (approved):
Should this be a standalone script?

zdykstra:
> Should this be a standalone script?

I'm okay with using `zfunc` as a helper script for now. I could be persuaded with a reasonable argument, though.

PR #540, opened 2023-11-29
docs/guides/: update pool compat to openzfs-2.2

zdykstra:
With ZFSBootMenu binary releases shipping with OpenZFS 2.2.x, we can bump the minimum pool compatibility recommend.  Debian will need to be special cased. _Again_.  I'm open to suggestions on what to do here - maybe we just stop setting pool option at all?

### DIFF
--- a/docs/guides/_include/pool-creation.rst
+++ b/docs/guides/_include/pool-creation.rst
@@ -56,6 +56,6 @@ Create the zpool
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


classabbyamp:
idk I'd probably stick with 2.1 compat for now, until most/all of the bugs are shaken out of 2.2

ahesford:
Maybe sticking with the latest LTS is a reasonable choice. The note below the command could be amended to note that the selection of the most recent LTS is intended to be conservative, and users are free to select a more recent compatibility option (or none at all) as they prefer. The note can conclude with an admonition that users should either plan to build their own images or select a version that is supported by whatever ZFSBootMenu release they intend to use.

PR #541, opened 2023-11-29
docs: switch to sphinx-book-theme

zdykstra:


ahesford (approved):
This is worth it for the better syntax highlighting and right-edge navigation alone.

zdykstra:
The man pages aren't quite as readable as the old documentation style, but I think they're good enough - https://zfsbootmenu--541.org.readthedocs.build/en/541/man/zfsbootmenu.7.html

PR #549, opened 2023-12-05
docs/guides/third-party: remove install scripts

classabbyamp:
Install scripts are liable to become outdated w/r/t our recommended practices, especially when they aren't written by us and the author lets them sit.

It's not hard to install ZBM from a guide, and users can create their own script based on those guides by just copying/adapting the codeblocks.

Not much is lost from removing these, as Void and Ubuntu have official install guides, and Arch has proper documentation linked on the Arch wiki.


zdykstra (approved):
It's reasonable to restrict our 3rd-party links to other primary/authoritative documentation sources - e.g. distribution Wiki articles.

ahesford (approved):

PR #550, opened 2023-12-06
zfsbootmenu-core: show enabled/disabled hooks in zreport

zdykstra:
I realized tonight that the hooks in the pre-built initramfs might not be very discoverable. Showing them in zreport (now exposed in the Help system) might improve that a bit.

Screenshot_20231205_220800

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -1924,6 +1924,8 @@ emergency_shell() {
 # returns: nothing
 
 zreport() {
+  local hook
+
   colorize white "System Report\n\n"
 
   (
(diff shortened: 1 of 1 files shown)


zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1950,7 +1952,17 @@ zreport() {
   colorize orange "\n>> ZFSBootMenu commandline\n"
   get_zbm_kcl | kcl_assemble
 
-  colorize orange "\n\n>> ZFS/SPL module information\n"
+  colorize orange "\n\n>> Enabled hooks\n"
```
classabbyamp, line 1955: i think the first `\n` should be on its own line grouped with the previous stanza just so we don't have to keep moving it around if things get added

zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1950,7 +1952,17 @@ zreport() {
...
+  for hook in /libexec/hooks/*.d/*; do
+    [ -x "${hook}" ] && echo "* $( colorize green "${hook}")"
+  done
+
+  colorize orange "\n>> Disabled hooks\n"
+  for hook in /libexec/hooks/*.d/*; do
+    [ -x "${hook}" ] || echo "* $( colorize red "${hook}")"
```
ahesford, line 1962: This is going to fail when there are no hooks installed.
```suggestion
    [ -f "${hook}" ] || continue
    [ -x "${hook}" ] && continue
    echo "* $( colorize red "${hook}")"
```

zfsbootmenu/lib/zfsbootmenu-core.sh
zdykstra: shifted it around slightly so that the line return is logically tied to `kcl_assemble`

PR #560, opened 2023-12-18
zsnapshots: check if input is a valid ZFS filesystem

zdykstra:
`zsnapshots` is used from the recovery shell, and as such it accepts any input from the user.  Attempt minimal validation that the filesystem name is actually a ZFS _filesystem_ . Errors are logged via zerror, to work in tandem with on-going work to print zerror logs to recovery shell.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -2072,7 +2072,6 @@ is_efi_system() {
   return 1
 }
 
-
 # arg1: 'ro' or 'rw' to mount or remount efivarfs in that desired mode
 # arg2: optional mountpoint
 # prints: nothing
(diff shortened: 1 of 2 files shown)


zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -2096,3 +2095,20 @@ mount_efivarfs() {
...
+  local dataset
+
+  dataset="${1}"
+
+  if [ -z "${dataset}" ]; then
+    zerror "dataset undefined"
+    return 1
+  fi
+
+  [[ "$( zfs get -H -o value type "${dataset}" 2>&1 )" == "filesystem" ]] && return 0
```
ahesford, line 2112:
```suggestion
  zfs list -H -o name -t filesystem "${dataset}" 2>&1 >/dev/null && return 0
```
This seems to work as expected for filesystem, zvol, snapshot and nonexistent arguments, but should be tested a bit more thoroughly.

zdykstra: I want it actually validating that it's a filesystem. zsnapshots can't operate on a zvol, snapshot, etc.

zdykstra: What does this offer over a string comparison test - speed?

ahesford: It doesn't offer much, but doesn't rely on any program output except the error code when the type isn't as expected.

PR #561, opened 2023-12-18
recovery shell: print zerrors since the last command was executed

zdykstra:
I'm not super happy with how `print_kmsg_logs` is shaking out, but there's only so much time I'm willing to spend chasing elegance.

This is modestly tested, and could likely use further refinement after people play with it.

### DIFF
--- a/zfsbootmenu/install-helpers.sh
+++ b/zfsbootmenu/install-helpers.sh
@@ -155,7 +155,8 @@ create_zbm_profiles() {
 	[ -f /etc/profile ] && source /etc/profile
 	[ -f /lib/zfsbootmenu-completions.sh ] && source /lib/zfsbootmenu-completions.sh
 
-	export PS1="\[\033[0;33m\]zfsbootmenu\[\033[0m\] \w > "
+	export PROMPT_COMMAND="/libexec/recovery-error-printer"
+	export PS1='\[\033[0;33m\]zfsbootmenu\[\033[0m\] \w > '
 
 	alias clear="tput clear"
 	alias reset="tput reset"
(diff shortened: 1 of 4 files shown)


zfsbootmenu/install-helpers.sh
```diff
@@ -151,7 +151,7 @@ create_zbm_profiles() {
 	[ -f /etc/profile ] && source /etc/profile
 	[ -f /lib/zfsbootmenu-completions.sh ] && source /lib/zfsbootmenu-completions.sh
 
-	export PS1="\[\033[0;33m\]zfsbootmenu\[\033[0m\] \w > "
+	export PS1='\$( /libexec/recovery-error-printer )\[\033[0;33m\]zfsbootmenu\[\033[0m\] \w > '
```
ahesford, line 154:
```suggestion
	export PS1='\[\033[0;33m\]zfsbootmenu\[\033[0m\] \w > '
	export PROMPT_COMMAND="/libexec/recovery-error-printer"
```

zdykstra: You can tell I don't actually use Bash as an interactive shell. I didn't know `PROMPT_COMMAND` existed.

ahesford: That's fair. We all know bash is and forever will be a programming language first.

ahesford (approved):
This is a good idea. Printing these messages in red might be nice, but isn't necessary.

zdykstra:
This is **good enough** for now. I'll think a bit more about how I want to refine `print_kmsg_logs` - but this simple solution is fine for now.

PR #568, opened 2023-12-28
releng/docker/build-init.sh: add an option to update packages

ahesford:
The moving target that is Void may make some additional packages uninstallable when running the build container with `-p <pkg>`. Even when this is not the case, users may wish to update packages in the container before building a new ZBM image. (This may be useful, for example, to test new versions of key tools.) A new `-u` argument to the container entrypoint can trigger an XBPS update in the container before images are built or other packages are installed.

Generally, container images lack kernel headers, dkms and other build tools, so updating kernel and zfs packages will result in missing kmods for the updated kernels. Therefore, the default update behavior is to hold all kernel and zfs packages. Adding a second `-u` will force a "full" update, which will allow the kernel and zfs packages to be updated as well. I'm not sure the "full" option is particularly useful, but there it is.

I still need to run through tests of these options, but this at least describes my idea.

### DIFF
--- a/releng/docker/build-init.sh
+++ b/releng/docker/build-init.sh
@@ -31,12 +31,15 @@ Usage: $0 [options]
   -p <package>
      Install the named Void Linux package in the container
      before building. May be specified more than once to
-     install more than one package. (This triggers a full
-     XBPS package upgrade before installation.)
+     install more than one package.
 
   -t <tag>
      Specify specific tag or commit hash to fetch
      (Ignored if /zbm already contains a ZFSBootMenu tree)
+
+  -u
+    Update all Void packages, except kernels and zfs, in the container
+    (Specify more than once to attempt kernel and zfs updates as well)
 
   -V
      Do not attempt to update the version recorded in the source tree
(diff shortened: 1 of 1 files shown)


releng/docker/build-init.sh
```diff
@@ -110,10 +150,12 @@ if [ -z "${ZBMTAG}" ]; then
...
-  # Trigger a sync and upgrade to make sure the package is installable
-  xbps-install -Syu xbps
+# Update packages, if desired
+if [ -n "${UPDATE}" ]; then
+  update_packages "${UPDATE}"
+fi
 
+if [ "${#PACKAGES[@]}" -gt 0 ]; then
   # Install all requested packages
   xbps-install -Sy "${PACKAGES[@]}"
```
zdykstra, line 160: Failures in the additional package installation step should be fatal.

releng/docker/build-init.sh
```diff
@@ -110,10 +150,12 @@ if [ -z "${ZBMTAG}" ]; then
   fi
 fi
 
-if [ "${#PACKAGES[@]}" -gt 0 ]; then
-  # Trigger a sync and upgrade to make sure the package is installable
-  xbps-install -Syu xbps
+# Update packages, if desired
+if [ -n "${UPDATE}" ]; then
+  update_packages "${UPDATE}"
```
zdykstra, line 155: Package update failures should _maybe_ be fatal, too.

zdykstra (approved):
Fatal errors aside, this looks fantastic.

PR #573, opened 2024-01-13
zfsbootmenu: add zbm.waitfor argument handler

zdykstra:
With keyfiles being stored on USB drives, the boot process can be non-deterministic due to intermittently missing /dev entries. Users can now provide one or more devices that must be present before the pool import process can start. Unlike dracut, you can simply hit the escape key to cancel this check.

### DIFF
--- a/docs/man/zfsbootmenu.7.rst
+++ b/docs/man/zfsbootmenu.7.rst
@@ -143,6 +143,14 @@ These options are set on the kernel command line when booting the initramfs or U
 
   Enable automatic font resizing of the kernel console to normalize the apparent resolution for both low resolution and high resolution displays. This option is enabled by default.
 
+**zbm.waitfor=device,device,...**
+
+  Ensure that one or more devices are present before starting the pool import process. Devices may be specified as full paths to device nodes (*e.g.*, **/dev/sda** or **/dev/disk/by-id/wwn-0x500a07510ee65912**) or, for convenience, as a typed indicator of the form **TYPE=VALUE**, which will be expanded internally as
+  
+    **/dev/disk/by-TYPE/VALUE**
+
+  The use of full device paths other than descendants of **/dev/disk/** is fragile and should be avoided.
+
 Deprecated Parameters
 ---------------------
 
--- a/zfsbootmenu/libexec/zfsbootmenu-init
+++ b/zfsbootmenu/libexec/zfsbootmenu-init
@@ -78,6 +78,60 @@ elif [ ! -e /etc/hostid ]; then
   write_hostid "${default_hostid}"
 fi
 
+# Wait for devices to show up
+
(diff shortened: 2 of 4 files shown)


docs/man/zfsbootmenu.7.rst
```diff
@@ -143,6 +143,18 @@ These options are set on the kernel command line when booting the initramfs or U
...
+
+  Ensure that one or more devices are present prior to starting the pool import process. Required devices should be either a direct **/dev** device entry or a partial **/dev/disk/by-** entry.
+
+  **zbm.waitfor=/dev/sda**
+
+    Direct device entries are fragile and should generally be avoided.
+
+  **zbm.waitfor=PARTUUID=b0319364-ec65-1840-a02b-2b58caa12ac0,ID=wwn-0x500a07510ee65912**
+
+    Refer to **/dev/disk/** on the host system to find the required device entries.
```
ahesford, line 156: Not sure if this alternative is any clearer:
```suggestion
  Ensure that one or more devices are present before starting the pool import process. Devices may be specified as full paths to device nodes (*e.g.*, **/dev/sda** or **/dev/disk/by-id/wwn-0x500a07510ee65912**) or, for convenience, as a typed indicator of the form **TYPE=VALUE**, which will be expanded internally as
  
    **/dev/disk/by-TYPE/VALUE**

  The use of full device paths other than descendants of **/dev/disk/** is fragile and should be avoided.
```

zfsbootmenu/libexec/zfsbootmenu-init
```diff
@@ -78,6 +78,60 @@ elif [ ! -e /etc/hostid ]; then
...
+        *=*)
+          ((EXPECTED++))
+          path_prefix="/dev/disk/by-${device%=*}"
+          checkfor="${path_prefix,,}/${device##*=}"
+          if [ -e "${checkfor}" ] ; then
+            ((FOUND++))
+          else
+            missing+=( "$device" )
+          fi
+          ;;
```
ahesford, line 109: It occurs to me that devices can be specified just as dynamically with full `/dev/disk/by-*` paths, so this could be retained as a convenience that saves a few characters or discarded to reduce maintenance burden.

ahesford (approved):

zfsbootmenu/libexec/zfsbootmenu-init
zdykstra: I like the more compact form - it's easier to parse when reading through the ZBM KCL.  There's not too much of a maintenance burden with it, so I'll retain it unless you'd like it removed for less code.

ahesford: No, leaving it is fine and mimics fstab entries.

PR #579, opened 2024-01-18
releng/make-binary.sh: specify release tag in builder images by default

ahesford:
Needs testing and backporting to v2.3.x before another release is tagged, but this ensures that we create images during releng that will build the release that is being tagged.


zdykstra (approved):
As near as I can tell, `releng/docker/image-build.sh` just records arg 2 in the container that's built, but doesn't actually do anything else with it. 

As long as the `vX.X` tag is pushed to the repository before `build-init.sh` is run, it's safe to record this tag in the build container.

PR #581, opened 2024-01-23
Update container-example.rst

curiousercreative:
fix doc for download of zbm-builder


zdykstra:
Thanks for catching that!

PR #582, opened 2024-01-26
zfsbootmenu: clean up main loop

zdykstra:
Rather than relying to a call to draw_be to determine if the bootenvs file is empty, the main loop should use the return code of populate_be_list to determine if any boot environments were found.

The sanity checks in draw_be are still present.


ahesford (approved):
Looks reasonable, and makes me wonder why it was structured so strangely before.

PR #583, opened 2024-01-27
'no pools available to import' printed to term if there are no pools

midzelis:


zdykstra:
Thanks for catching that! I cleaned up a number of other zfs/zpool interactions. Please let me know if you see any others.

PR #586, opened 2024-01-30
zsnapshots: allow filesystem selection when no argument is specified

ahesford:
A couple of points:
- There is some duplication of `fzf` arguments with `draw_be`, but I'm not sure how hard we want to try to unify those two.
- To get proper ANSI coloring of the `border-label`, I needed to define `fs_header`; no amount of quoting magic allowed me to specify the header string inline and have color rather than seeing the raw ANSI escape sequence.

### DIFF
--- a/zfsbootmenu/bin/zsnapshots
+++ b/zfsbootmenu/bin/zsnapshots
@@ -1,4 +1,5 @@
 #!/bin/bash
+# vim: softtabstop=2 shiftwidth=2 expandtab
 
 sources=(
   /lib/profiling-lib.sh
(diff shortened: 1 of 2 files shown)


zfsbootmenu/bin/zsnapshots
```diff
@@ -19,18 +20,61 @@ done
 
 unset src sources
 
+list_root_filesystems() {
```
zdykstra, line 23: Is this worth trying to merge with the largely similiar process in `populate_be_list()` in `zfsbootmenu-ui.sh`, so that if the logic for determining a root filesystem changes in the future it doesn't have to be updated in multiple places?

zdykstra (approved):

zfsbootmenu/bin/zsnapshots
ahesford: Good point. I moved the "BE candidate" logic into a common function.

zfsbootmenu/lib/zfsbootmenu-ui.sh
```diff
@@ -623,11 +616,35 @@ populate_be_list() {
...
+  local be_list fs candidates ret
+
+  be_list="${1}"
+  if [ -z "${be_list}" ]; then
+    zerror "be_list is undefined"
+    return 1
+  fi
+  zdebug "be_list set to ${be_list}"
+
+  sort_key="$( get_sort_key )"
```
zdykstra, line 642: I don't think this needs to get a sort key any more

zfsbootmenu/bin/zsnapshots
```diff
@@ -19,18 +20,61 @@ done
...
- exit
-fi
+fs_header() {
+  echo -n -e "\\033[1;33m[ Select a filesystem ]"
+}
 
 fs="${1}"
 
+if [ -z "${fs}" ]; then
+  if ! candidates="$(list_root_filesystems 2>/dev/null)"; then
```
zdykstra, line 54: The new function should be used in place of the one-off `list_root_filesystems`

zdykstra (changes requested):

zfsbootmenu/lib/zfsbootmenu-ui.sh
ahesford: Done

zfsbootmenu/bin/zsnapshots
ahesford: I can't believe I forgot to change this!

zdykstra (approved):

PR #587, opened 2024-02-03
splash image support

classabbyamp:
now we're a real bootloader


etc/zfsbootmenu/release.yaml
```diff
@@ -8,5 +8,6 @@ Components:
 EFI:
   Versions: false
   Enabled: true
+  SplashImage: /etc/zfsbootmenu/splash.bmp
```
zdykstra, line 11: This one took a bit to track down, but `releng/make-binary.sh` mounts the source tree read-only at `/zbm` in the build container. The path for the splash image should be `/zbm/etc/zfsbootmenu/splash.bmp`.

ahesford: I think it would be better to leave this as `/etc/zfsbootmenu/splash.bmp`, but modify [`releng/docker/build-init.sh`](https://github.com/zbm-dev/zfsbootmenu/blob/master/releng/docker/build-init.sh#L247) to link to `splash.bmp` in addition to `config.yaml` and `mkinitcpio.conf`.

Even if the path were to remain unchanged, the container entrypoint should still be modified because the entrypoint is supposed to make all symlinks necessary to make the tree at `/zbm` look like a local install.

ahesford:
Tested successfully on a Dell laptop, looks nice!

zdykstra:
On my hardware, at least, the ZBM splash is jarring/out of place due to the Lenovo logo disappearing and then reappearing before the countdown screen is shown. 

I'd rather not see release/recovery EFIs exhibiting that behavior, even if it is likely nothing under our control.

With that said, it does look good on certain hardware - just not any of mine.

etc/zfsbootmenu/release.yaml
classabbyamp: now that the splash image isn't default (because it's in `/usr/share/examples/zfsbootmenu/splash.bmp` and the config looks at `/etc/zfsbootmenu/splash.bmp` by default), users can just mount the example (or their own) in the container as desired

zdykstra (approved):

PR #590, opened 2024-02-06
reorganise docs somewhat

classabbyamp:
- docs/: always use venv on void
- docs/: reorganise logos/images
- docs/guides/general/: move to docs/general/


zdykstra:
This looks fantastic!

classabbyamp:
redirect test: https://zfsbootmenu--590.org.readthedocs.build/en/590/guides/general/bootenvs-and-you.html

docs/README.md
```diff
@@ -8,42 +8,43 @@ The `Makefile` alongside this document provides several targets to prepare the S
...
-- sphinx_rtd_theme
-- recommonmark
+Some make targets require further programs: `watchexec` for `make serve` and `rst2ansi` for `make gen-man`.
 
-On Void Linux, these can be installed with `make setup-void`. Otherwise, these packages can be installed through your
-system package manager or via pip in a virtual environment with `make setup`.
-
-If using the virtual environment method, you will then need to activate this environment with `source env/bin/activate`
-(may be different for some shells) before running a build command.
+If on Void Linux, `make setup-void` will install these programs and set up the virtual environment.
```
ahesford, line 16:
```suggestion
On Void Linux, `make setup-void` will install these programs and set up the virtual environment.
```

ahesford (approved):
Looks good to me

classabbyamp:
not 100% optimal because it uses [`http-equiv.refresh`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv) redirects. we could use RTD [redirects](https://docs.readthedocs.io/en/stable/user-defined-redirects.html#user-defined-redirects) (which allows for 301 permanent redirects) but those cannot be defined in the config file, only in the dashboard. thoughts?

zdykstra:
> not 100% optimal because it uses [`http-equiv.refresh`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv) redirects. we could use RTD [redirects](https://docs.readthedocs.io/en/stable/user-defined-redirects.html#user-defined-redirects) (which allows for 301 permanent redirects) but those cannot be defined in the config file, only in the dashboard. thoughts?

I'm a big fan of having a single source of truth. Manually updating a web dashboard with redirects outside of git/project history isn't ideal. 

It looks like we can possibly use the [RTD API](https://docs.readthedocs.io/en/stable/api/v3.html#redirect-create) to create redirects, though this might be more trouble than it's worth to do.

PR #607, opened 2024-03-17
zfsbootmenu/bin/reboot: add reboot-to-uefi functionality

### LINKED ISSUE #423: feature request: option to reboot into efi firmware setup
Use case: You want to go to the BIOS/EFI setup menu. But for example didn't press the key in time, or perhaps don't know which key to press for this particular motherboard, etc. What if you could just press a simple hotkey, for example "Ctrl-F" for **f**irmware, and let ZBM reboot into it?

Similar features exist for example in GRUB, the Windows bootloader, systemd (`systemctl reboot --firmware-setup`).

Looks like it should be fairly straightforward to implement, just set a value in the efivarfs and reboot. We can detect if it is supported before showing the hotkey option at all. A quickly thrown together example:

```sh
#!/bin/bash

# args: none
# prints: nothing
# returns: 0 if reboot to EFI firmware UI is supported, 1 if not
efi_supports_reboot_to_fw_ui() {
    if ! is_efi_system ; then
        zdebug "efivarfs unsupported"
        return 1
    elif [ ! -r /sys/firmware/efi/efivars/OsIndicationsSupported-8be4df61-93ca-11d2-aa0d-00e098032b8c ] ; then
        zdebug "OsIndicationsSupported unsupported"
        return 1
    elif [ ! -r /sys/firmware/efi/efivars/OsIndications-8be4df61-93ca-11d2-aa0d-00e098032b8c ] ; then
        zdebug "OsIndications unsupported"
        return 1
    fi

    # Check if the EFI_OS_INDICATIONS_BOOT_TO_FW_UI = 0x01 bit is set
    perl -e 'exit !(unpack("x4Q<", <STDIN>) & 1)' \
        < /sys/firmware/efi/efivars/OsIndicationsSupported-8be4df61-93ca-11d2-aa0d-00e098032b8c
}

# args: none
# prints: nothing
# returns: 1 on error, 2 if unsupported, otherwise does not return
efi_reboot_to_fw_ui() {
    if ! efi_supports_reboot_to_fw_ui ; then
        zdebug "efi reboot to fw unsupported"
        return 2
    fi

    # Remount efivarfs to be writable
    mount_efivarfs rw
    if [ ! -w /sys/firmware/efi/efivars/OsIndications-8be4df61-93ca-11d2-aa0d-00e098032b8c ] ; then
        zdebug "OsIndications not writable"
        return 1
    fi

    # Set the EFI_OS_INDICATIONS_BOOT_TO_FW_UI = 0x01 bit,
    # with required flags set:
    # EFI_VARIABLE_NON_VOLATILE = 0x01
    # EFI_VARIABLE_BOOTSERVICE_ACCESS = 0x02
    # EFI_VARIABLE_RUNTIME_ACCESS = 0x04
    perl -e 'print pack("(LQ)<", 1|2|4, unpack("x4Q<", <STDIN>) | 1)' \
        < /sys/firmware/efi/efivars/OsIndications-8be4df61-93ca-11d2-aa0d-00e098032b8c \
        | sponge /sys/firmware/efi/efivars/OsIndications-8be4df61-93ca-11d2-aa0d-00e098032b8c
    reboot
}
```

This example depends on `perl` (for the unsigned 64 bit LE integers) and `sponge` (from moreutils), but it can probably be done in pure Bash as well.

classabbyamp:
fixes #423

menu without:
image

menu with:
image


zfsbootmenu/lib/zfsbootmenu-ui.sh
```diff
@@ -664,3 +664,53 @@ populate_be_list() {
...
+  LMARGIN=$(( (COLS - maxlen - 6) / 2 ))
+  TMARGIN=$(( (ROWS - nopts - 3) / 2 ))
+
+  if OUTPUT="$(echo -e "${PROMPT}" | \
+    fzf --no-info \
+      --with-nth="2.." \
+      --delimiter=":" \
+      --no-multi \
+      --no-sort \
+      --border=sharp \
```
zdykstra, line 706: The `--border` flag prefixes need to be gated behind the `HAS_BORDER` toggle, since some platforms we build on (looking at you, Debian) ship with an fzf that's too old to support those flags.

```
${HAS_BORDER:+--border=sharp} \
${HAS_BORDER:+--border-label="${TITLE}"} \
```

classabbyamp:
reordered the shortcuts:
image

classabbyamp:
the reboot-to-firmware doesn't work on the testing VMs (even with the edk2-ovmf efi) because the `OsIndications` efivar is missing, but i have successfully tested it on my laptop.

classabbyamp:
image

reordered again to be a bit more functionally cohesive. things that only affect/act on a single bootenv are in the two left columns (except recovery shell). also moved `edit kcl` down so that it's next to `revert kcl` when it shows up

zdykstra (approved):

zdykstra:
The menu changes and the modal dialog function came together really nice. Everything here looks good to go!

PR #609, opened 2024-03-18
Makefile: make zbm-release the default action

classabbyamp:
this is similar to many other projects that are install-only, like xtools

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ EXAMPLES=$(PREFIX)/share/examples/zfsbootmenu
 
 .PHONY: install core dracut initcpio zbm-release show-version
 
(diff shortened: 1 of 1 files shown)


zdykstra (approved):

classabbyamp:
alternatively, we could move `zbm-release` up so it's the default target, as that's what the void template does: https://github.com/void-linux/void-packages/blob/master/srcpkgs/zfsbootmenu/template#L6

ahesford:
I like making the `zbm-release` target up to the default slot. If you want to keep a no-op target, I'd prefer a name like `dummy` or `no-op` over `all`, which implies it will run other functional targets.

PR #612, opened 2024-04-02
Allow to build ZFSBootManager in containerized Debian environment

BohdanTkachenko:
See https://github.com/zbm-dev/zfsbootmenu/issues/614 for motivation.

- Added Dockerfile.debian so it does not conflict with the main Dockerfile
- Adjusted build-init.sh to support `apt` in addition to `xbps`

### DIFF
--- a/releng/docker/Dockerfile.debian
+++ b/releng/docker/Dockerfile.debian
@@ -1,0 +1,86 @@
+# syntax=docker/dockerfile:1.4
+#
+# This Dockerfile creates a container that will create an EFI executable and
+# separate kernel/initramfs components from a ZFSBootMenu repository. The
+# container will pre-populate its /zbm directory with a clone of the master
+# branch of the upstream ZFSBootMenu branch and build the images from that.
+#
+# To use a different ZFSBootMenu repository or version, bind-mound the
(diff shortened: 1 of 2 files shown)


ahesford:
The Dockerfile and corresponding buildah script live in `releng` because they are primarily tools to support the production of our release builds. We expose our build containers for public consumption mostly because we've had trouble supporting Debian's ancient versions of utilities that are essential to ZBM. (We have a bunch of feature checks around things like `fzf` that are necessary *only* because of Debian.) When users have trouble building custom host images because their userspace utilities don't work with ZBM, our response is "build it in a container that gives you the tools you need". Mostly, our expectation is that the `zbm-builder.sh` front-end script will be used to drive the build container.

Our official release engineering relies on buildah, with the Dockerfile a vestigial remnant of the early days of ZBM build containers. It's kind of nice to keep it around in case we ever need to drop buildah for some reason, but the reality is that the Dockerfile doesn't get much official attention as it is, and some day we'll probably just drop it when it diverges too much from the buildah script.

Adding additional Dockerfiles, especially for distributions that may ship problematic versions of tools that we require and that will get essentially zero official testing from us, is not something we are willing to support at this time.

PR #613, opened 2024-04-02
Fix ZFSBootMenu initialization with systemd

BohdanTkachenko:
Fixes most of issues I encountered in https://github.com/zbm-dev/zfsbootmenu/issues/614

I understand that this fix is probably not very good since it possibly breaks something else, but I would be glad to adjust it based on the feedback.


ahesford:
This might make ZFSBootMenu get along with the Dracut event loop, but that is the opposite of what we're trying to accomplish here. The `exec` ensures that our fundamental assumption---that ZFSBootMenu has total control of the system---is satisfied.

Issue #616, opened 2024-04-12
Shortcuts missing from /usr/bin


Sithuk:
### ZFSBootMenu build source

Local build, dracut

### ZFSBootMenu version

master

### Problem description

The symlinks such as "zbm" and "poweroff" are missing from the zfs shell /usr/bin in a fresh install using the master. I am testing remote access to unlock a zfs on luks drive.

### Steps to reproduce

1. Compile from master.
2. Reboot.
3. Type zbm to start zfsbootmenu.

zdykstra:
Are you SSH'ing in to ZFSBootMenu? If so, what SSH server do you have embedded in your image? What's the value of the `PATH` environment variable? What distribution are you building this on?

Sithuk:
I am SSH'ing in, yes. 
Dropbear is embedded (SSH-2.0-dropbear_2020.81).
The PATH environment variable is: /zbm/bin:/usr/sbin:/usr/bin:/sbin:/bin
I built on Ubuntu 22.04.
"zbm" is not present in /usr/sbin of the zfsbootmenu initramfs shell. I didn't find it when I did a search using find either.

zdykstra:
Is this issue still present in a build from the master branch? This was not able to be replicated locally.

Sithuk:
Just carried out a test with the master branch. The zbm shortcut is still missing unfortunately. "zfsbootmenu" works fine.

Sithuk:
I've updated the ubuntu zfsbootmenu install script. You can download the latest from the link below. That should help you duplicate the lack of zbm shortcut that I have experienced.
https://github.com/Sithuk/ubuntu-server-zfsbootmenu

classabbyamp:
could you upload the generated zfsbootmenu files?

ahesford:
This is a problem with the installation script called from the Makefile. I'll propose a fix later today.

ahesford:
If the fix in #628 doesn't solve this problem, please let me know.

PR #618, opened 2024-04-15
zfsbootmenu/install-helpers.sh: preserve relative symlinks

ahesford:
When installing symlinks into their initramfs images, both dracut and mkinitcpio attempt to canonicalize symlink targets, copying the resolved target into place and making a new symlink that points to there. This means that, even when building a local image, symlinks like `poweroff` and `zbm` will point copies of scripts in `/usr/lib/zfsbootmenu/bin` rather than the versions already installed into `/usr/bin`. For container builds or those built from custom installations that override `$zfsbootmenu_module_root`, the target paths get even weirder in the initramfs image. It's also problematic, because people doing testing may attempt to modify things like `/usr/bin/zfsbootmenu` *in situ* and be surprised when they run `zbm` and get a *different* copy of the script than the one they modified.

I've made a simple helper that tries to do the right thing with symlink *as we use them*. When installing a file, if it is a symlink and resides in `$zfsbootmenu_module_root`, then the relative link will be replicated in the initramfs image. This seems to work as intended for the handful of symlinks that we install.

In the original implementation, I required that the symlink point to a file in the same location and did not attempt to install the target, instead relying on a `zbm_install_file` call for the actual target to add the file separately. In the fixup commit, I instead attempt to install the target in a location where the symlink will find it. I'm open to suggestions on how to improve this target installation, and am also willing to just drop the fixup commit.

### DIFF
--- a/zfsbootmenu/install-helpers.sh
+++ b/zfsbootmenu/install-helpers.sh
@@ -170,7 +170,42 @@ create_zbm_profiles() {
   ln -s "/root/.bashrc" "${BUILDROOT}/root/.profile"
 }
 
+zbm_install_symlink() {
+  [ -L "${1}" ] || return 1
+
+  # shellcheck disable=SC2154
+  case "${1}" in
+    "${zfsbootmenu_module_root}"/*) ;;
+    *) return 1 ;;
+  esac
+
+  local dest src
+
+  dest="$(readlink "${1}")" || dest=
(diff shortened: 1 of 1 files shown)


ahesford:
This may or may not be related to #616; I can't reproduce that issue when generating dracut or mkinitcpio images on my host when standing up testbeds.

PR #623, opened 2024-04-30
Fix InitCPIOHookDirs parent key in host contaner build documentation

ALPHA-60:
Build script [expects](https://github.com/zbm-dev/zfsbootmenu/blob/dc755a87a2397383cb8a0304acf446df60530950/bin/generate-zbm#L172) the `InitCPIOHookDirs` to be in the "Global" section, but the docs say to put it "General" section, which makes hooks unavailable when building the image.


ahesford:
Thanks!

PR #661, opened 2024-08-06
Update zfs-config.rst

camckay:
Add zfs-scripts to packages installed. It is listed when setting up live environment but missing here. This leads to zpool errors after installation because /usr/share/zfs/compatibility.d is missing.


zdykstra:
Thank you. Please adjust the commit message to match previous documentation commits and this can be pushed to the website.

PR #665, opened 2024-08-16
docs/man/zfsbootmenu: additional language for bootfs prop

zdykstra:


docs/man/zfsbootmenu.7.rst
```diff
@@ -183,7 +183,7 @@ The following properties can be set at the pool level to control boot behavior.
 
 **bootfs**
 
-  A dataset that will be considered the default boot environment if the pool is the first to be imported by ZFSBootMenu.
+  A dataset that will be considered the default boot environment if the pool is the first to be imported by ZFSBootMenu. This must be set for automatic booting to function.
```
ahesford, line 186:
```suggestion
  A dataset that will be considered the default boot environment if the pool is the first to be imported by ZFSBootMenu.
  
  .. note::

    This must be set for automatic booting to function. When no **bootfs** property is detected, ZFSBootMenu will always display a menu.
```

zdykstra: Yeah, that's better phrasing.

PR #675, opened 2024-09-22
contrib: add rc.d script to install new kernels in build containers

ahesford:
This is a simple `rc.d` script to install new kernels in containers instantiated from older build containers.


zdykstra:
I wouldn't object to this functionality being baked right into the primary build manager script/container.

Otherwise this looks good to me.

classabbyamp:
yeah this seems reasonable to add outright

PR #678, opened 2024-09-25
testing/: add Makefile to install Void deps

zdykstra:
I don't setup a new system often, but when I do I completely forget what the `testing/` run-time requirements are.

### DIFF
--- a/testing/Makefile
+++ b/testing/Makefile
@@ -1,0 +1,18 @@
+.DEFAULT_GOAL := help
+
+SUDO          ?= sudo
+INSTALLOPTS   ?= -Su
+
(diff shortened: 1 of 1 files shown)


testing/Makefile
```diff
@@ -0,0 +1,18 @@
+.DEFAULT_GOAL := help
+
+SUDO          ?= sudo
+INSTALLOPTS   ?= -Su
+
+VOID_PKGS     = perl-boolean perl-Sort-Versions lsof yq-go qemu
```
classabbyamp, line 6: do we need all of qemu? that includes all user and system emulators

zdykstra: Good call. I've narrowed it down to what's strictly needed to create/start an amd64 guest.

PR #694, opened 2024-11-11
dracut: explicitly require the bash module

### LINKED ISSUE #690: dracut fails to boot
### ZFSBootMenu build source

Local build, dracut

### ZFSBootMenu version

2.3.0

### Boot environment distribution

Void Linux

### Problem description

Building zfsbootmenu with dracut results in a kernel panic when attempting to boot. See below:

screenshot

### Steps to reproduce

Here is my config file:

```
Global:
  ManageImages: true
  InitCPIO: false
  PreHooksDir: /etc/zfsbootmenu/generate-zbm.pre.d
  PostHooksDir: /etc/zfsbootmenu/generate-zbm.post.d
Components:
  Enabled: false
EFI:
  Enabled: true
  Versions: false
  ImageDir: /boot/efi/EFI/zbm
Kernel:
  CommandLine: quiet loglevel=0 zbm.show zbm.sort_key=creation zbm.skip_hooks=20-console-autosize.sh video=efifb:1920x1080
```

Changing `InitCPIO` to `true` and thus building with mkinitcpio works fine. This is a dracut specific issue.

zdykstra:
On Void, and possibly other distributions, /bin/sh points to /bin/dash. When no specific shell module (bash, dash, busybox) is requested, the system /bin/sh is used by default. Because ZFSBootMenu blacklists dash, shell-interpreter is unable to be installed, which then blocks multiple other key/core dracut modules.

Fixes #690

### DIFF
--- a/etc/zfsbootmenu/dracut.conf.d/zfsbootmenu.conf
+++ b/etc/zfsbootmenu/dracut.conf.d/zfsbootmenu.conf
@@ -1,3 +1,3 @@
 nofsck="yes"
-add_dracutmodules+=" zfsbootmenu "
+add_dracutmodules+=" zfsbootmenu bash "
 omit_dracutmodules+=" btrfs zfs resume systemd systemd-initrd dracut-systemd plymouth dash "


zdykstra:
This should optionally be cherry-picked into the 2.3.x branch and a new 2.3.1 release tagged / packaged for Void linux users. There's no reason to build new release assets for 2.3.1; this issue only impacts source users.

classabbyamp (approved):

ahesford (approved):
I'm surprised this isn't handled by `depends()` in the module setup, but this seems fine to me.

PR #695, opened 2024-11-13
contrib: add key/cert options and signature check for sbsign

### LINKED ISSUE #693: Improve `contrib/zbm-sign.pl` by checking signiture before signing & allow user to define the cert and key file names
### ZFSBootMenu build source

Local build, dracut

### ZFSBootMenu version

master

### Boot environment distribution

Debian 12

### Problem description

The problem with the current implementation for `sbsign` use case is that it appends the `-signed.efi` to the file and the `generate-zbm` script no longer prunes the files corectly when the `versions > 1`. 

Here are my minimal changes to the [contrib/zbm-sign.pl](https://github.com/zbm-dev/zfsbootmenu/blob/master/contrib/zbm-sign.pl):

```perl
#     SecureBoot:
#       SignBackup: true
#       DeleteUnsigned: false
#       SignMethod: sbctl
#       KeyDir: /etc/sbkeys
#       KeyFileName: DB.key
#       CrtFileName: DB.crt

  if ( $SignMethod eq "sbctl" ) {
    system "sbctl sign $ZBM/$_";
  } elsif ( $SignMethod eq "sbsign" ) {
    my $verify_output = "sbverify --cert $KeyDir/$CrtFileName $ZBM/$_ 2>&1";
    if ( $verify_output =~ /Signature verification OK/ ) {
      say "File $_ is already signed.";
      next;
    }
    system "sbsign --key $KeyDir/$KeyFileName --cert $KeyDir/$CrtFileName $ZBM/$_ --output $ZBM/$_";
  } else {
    die "Sign method $SignMethod not valid.";
  }
  # Remove the deleting part
```

Would you be interested in a PR? Feedback on improving my Perl programming skills is welcome.

### Steps to reproduce

1. Install [sbsign 0.9.4](https://manpages.debian.org/bookworm/sbsigntool/sbsign.1.en.html)
2. Build with the perl script configured
etc...

Adito5393:
Closes #693

### DIFF
--- a/contrib/zbm-sign.pl
+++ b/contrib/zbm-sign.pl
@@ -20,6 +20,8 @@
 #       DeleteUnsigned: false
 #       SignMethod: sbctl
 #       KeyDir: /etc/sbkeys
+#       KeyFileName: DB.key
+#       CrtFileName: DB.crt
 #
 # The configuration keys should be self-explanatory.
 
@@ -48,6 +50,8 @@ my $ESP    = $Global->{BootMountPoint};
 
 my $SecureBoot     = $config->{SecureBoot} or die "No config found, please edit /etc/zfsbootmenu/config.yaml";
 my $KeyDir         = $SecureBoot->{KeyDir};
+my $KeyFileName    = $SecureBoot->{KeyFileName};
+my $CrtFileName    = $SecureBoot->{CrtFileName};
 my $DeleteUnsigned = $SecureBoot->{DeleteUnsigned};
 my $SignBackups    = $SecureBoot->{SignBackup};
 $SignMethod = $SecureBoot->{SignMethod};
@@ -56,9 +60,9 @@ opendir my $ZBM_dir, $ZBM
   or die "Cannot open ZBM dir: $ZBM";
 
 if ($SignBackups) {
-  @EFIBins = grep { !/signed\.efi$/i and /\.efi/i } readdir $ZBM_dir;
+  @EFIBins = sort grep { !/signed\.efi$/i and /\.efi/i } readdir $ZBM_dir;
 } else {
-  @EFIBins = grep { !/signed\.efi$/i and !/backup/i and /\.efi/i } readdir $ZBM_dir;
+  @EFIBins = sort grep { !/signed\.efi$/i and !/backup/i and /\.efi/i } readdir $ZBM_dir;
 }
 
 say "Found: @EFIBins";
@@ -72,17 +76,14 @@ for (@EFIBins) {
   if ( $SignMethod eq "sbctl" ) {
     system "sbctl sign $ZBM/$_";
   } elsif ( $SignMethod eq "sbsign" ) {
-    $Unsigned = substr( $_, 0, -4 );
-    system "sbsign --key $KeyDir/DB.key --cert $KeyDir/DB.crt $ZBM/$_ --output $ZBM/$Unsigned-signed.efi";
+    my $verify_output = "sbverify --cert $KeyDir/$CrtFileName $ZBM/$_ 2>&1";
+    if ( $verify_output =~ /Signature verification OK/ ) {
+      say "File $_ is already signed.";
+      next;
+    }
+    system "sbsign --key $KeyDir/$KeyFileName --cert $KeyDir/$CrtFileName $ZBM/$_ --output $ZBM/$_";
   } else {
     die "Sign method $SignMethod not valid.";
-  }
-
-  if ( $DeleteUnsigned && $SignMethod eq "sbctl" ) {
-    say "sbctl signs in place, not deleting $_";
-  } elsif ( $DeleteUnsigned && $SignMethod ne "sbctl" ) {
-    say "Deleting unsigned $_";
-    system "rm $ZBM/$_";
   }
 }
 print "---------- FINISHED ----------\n";
--- a/contrib/zbm-sign.pl
+++ b/contrib/zbm-sign.pl
@@ -18,9 +18,8 @@
 #     SecureBoot:
 #       SignBackup: true
 #       SignMethod: sbctl
-#       KeyDir: /etc/sbkeys
-#       KeyFileName: DB.key
-#       CrtFileName: DB.crt
+#       KeyFileName: /etc/sbkeys/DB.key
+#       CrtFileName: /etc/sbkeys/DB.crt
 #
 # The configuration keys should be self-explanatory.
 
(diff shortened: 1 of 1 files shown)


contrib/zbm-sign.pl
```diff
@@ -72,17 +76,14 @@
...
+    if ( $verify_output =~ /Signature verification OK/ ) {
+      say "File $_ is already signed.";
+      next;
+    }
+    system "sbsign --key $KeyDir/$KeyFileName --cert $KeyDir/$CrtFileName $ZBM/$_ --output $ZBM/$_";
   } else {
     die "Sign method $SignMethod not valid.";
   }
-
-  if ( $DeleteUnsigned && $SignMethod eq "sbctl" ) {
```
zdykstra, line 81: Why was this whole block removed? It seems like it will fundamentally change the behavior of this script.

Adito5393: Really? It does only 1 of the 2 things:

1. Print a message if `sbctl` is used
2. Clean up after the inability of `sbsign` to sign in place... with the new code, that is no longer the case

Maybe I should remove the option `DeleteUnsigned` since it is no longer needed.

What am I missing? I see this block of code as no longer needed.

Adito5393:
I did my best with the PR title. Feel free to edit it to match the project standards.

contrib/zbm-sign.pl
```diff
@@ -20,6 +20,8 @@
 #       DeleteUnsigned: false
 #       SignMethod: sbctl
 #       KeyDir: /etc/sbkeys
+#       KeyFileName: DB.key
+#       CrtFileName: DB.crt
```
ahesford, line 23: The purpose of `KeyDir` is to provide a single configuration option that provides an access to both the cert and key. Adding separate name fields for each now requires setting three options to refer to two values. This is bad UX.

Either keep the `KeyDir` as is, or require `KeyFile` and `CertFile` as independent, full paths.

contrib/zbm-sign.pl
```diff
@@ -56,9 +60,9 @@
   or die "Cannot open ZBM dir: $ZBM";
 
 if ($SignBackups) {
-  @EFIBins = grep { !/signed\.efi$/i and /\.efi/i } readdir $ZBM_dir;
+  @EFIBins = sort grep { !/signed\.efi$/i and /\.efi/i } readdir $ZBM_dir;
```
ahesford, line 61: Why sort?

contrib/zbm-sign.pl
```diff
@@ -56,9 +60,9 @@
...
 } else {
-  @EFIBins = grep { !/signed\.efi$/i and !/backup/i and /\.efi/i } readdir $ZBM_dir;
+  @EFIBins = sort grep { !/signed\.efi$/i and !/backup/i and /\.efi/i } readdir $ZBM_dir;
```
ahesford, line 63: As above, why sort?

contrib/zbm-sign.pl
```diff
@@ -72,17 +76,14 @@
   if ( $SignMethod eq "sbctl" ) {
     system "sbctl sign $ZBM/$_";
   } elsif ( $SignMethod eq "sbsign" ) {
-    $Unsigned = substr( $_, 0, -4 );
-    system "sbsign --key $KeyDir/DB.key --cert $KeyDir/DB.crt $ZBM/$_ --output $ZBM/$Unsigned-signed.efi";
+    my $verify_output = "sbverify --cert $KeyDir/$CrtFileName $ZBM/$_ 2>&1";
```
ahesford, line 77: You aren't actually running this command, so the test on the output you are expecting will always fail.

Furthermore, if possible, it would be better to distinguish whether the input is already signed by the exit code, rather than grepping output for a magic string.

contrib/zbm-sign.pl
```diff
@@ -72,17 +76,14 @@
...
+    if ( $verify_output =~ /Signature verification OK/ ) {
+      say "File $_ is already signed.";
+      next;
+    }
+    system "sbsign --key $KeyDir/$KeyFileName --cert $KeyDir/$CrtFileName $ZBM/$_ --output $ZBM/$_";
```
ahesford, line 82: Replacing the unsigned input with the signed output is destructive and probably shouldn't be the default behavior. It seems better to leave this as it was, creating a separate signed file.

contrib/zbm-sign.pl
ahesford: Are you sure that sbsign won't corrupt its input if you try to overwrite it with output? Is that the case for several prior versions of sbsign?

If it's safe, the better approach would be to replace `DeleteUnsigned` with something a bit more meaningful like `Overwrite` and handle that above by just changing what you're passing as the output name to sbsign.

ahesford (changes requested):

contrib/zbm-sign.pl
Adito5393: Glad you asked. The `sbverify` command modifies the efi files last modified timestamp resulting in refind showing as the default ZBM the last of the this array (last of the array is first in rEFInd list). This `sort` solution allows the user to controll the rEFInd order via naming.

Do you see an issue with using sort?

contrib/zbm-sign.pl
Adito5393: Off-topic: I'm glad I'm invested time into this PR.

On-topic: This exact line of code behaves as expected. I can see it only signs the files that need signing. How do you explain it?
Should I replace the double qoutes with qx or backticks? What do you prefer?

I will look into the exit codes and see if that is an option.

contrib/zbm-sign.pl
Adito5393: This is exactly the default behavior of `sbctl`. The [Arch wiki](https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface/Secure_Boot#With_sbsigntools) shows an example using the same approach:
> sbsign --key db.key --cert db.crt --output /boot/vmlinuz-linux /boot/vmlinuz-linux

Have you had bad experiences with `sbsign` leaving you with a corrupt file?

contrib/zbm-sign.pl
Adito5393: Then I propose `UseOriginalFileName`:
- true -> *.efi = stays the same.
- false (default) -> *.signed.efi

Or you can think of a better boolean var name?

contrib/zbm-sign.pl
ahesford: Yes, this is not the place to hack around rEFInd sorting specifics. There is no need to sort unless this script depends on the order of the array, which it does not.

contrib/zbm-sign.pl
ahesford: Never assume that a process will complete successfully, especially when you are changing existing behavior that is less destructive. As I noted earlier, replacing the `DeleteUnsigned` option with an `Overwrite` and using that to control whether `--output` is passed at all is a better approach than always clobbering inputs.

contrib/zbm-sign.pl
ahesford: I proposed, and still prefer, `Overwrite`, which is both shorter and more descriptive.

PR #698, opened 2024-11-25
bin/generate-zbm: remove gummiboot from default stub list

zdykstra:
gummiboot doesn't work with newer (6.6+?) kernels, so lets stop recommending it as a default. If you still want to use it, set EFI.Stub in your configuration file.

The testing scripts still reference the old gummiboot location as a fall back, but that's not unreasonable that we leave that in our testing infra.


zdykstra:
Looks like this will need a new build container that has `systemd-boot` embedded in it.

PR #700, opened 2024-12-01
Run zgenhostid without a static value

Tookmund:
When generating a new host id, a specific value should not be provided so that each host has a different unique host id.

https://openzfs.github.io/openzfs-docs/man/master/8/zgenhostid.8.html


zdykstra:
The specific value is used because it's the default value in our prebuilt binaries. Unique hostids are pointless unless you're doing multipath/HA SAS.

PR #710, opened 2025-01-20
releng/make-binary.sh: build images for all installed kernels

ahesford:
We have multiple kernels in the build image, so we might as well spit out ZBM images for each one. Tested (for basic image production, not that the images are actually function) with

    ./releng/make-binary.sh v3.0.0 ghcr.io/zbm-dev/zbm-builder:development

### DIFF
--- a/releng/make-binary.sh
+++ b/releng/make-binary.sh
@@ -59,52 +59,71 @@ else
   mkdir -p "${assets}"
 fi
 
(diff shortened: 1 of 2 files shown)


zdykstra:
That all looks pretty good to me. Re-reading the source for get.zfsbootmenu.org, I think I'll need to patch it to understand that there are multiple EFI / .tar.gz files available for each release. I'll need to publish a release with multiple assets on my personal fork / point a copy of get.zfsbootmenu.org against that to see how it behaves.

PR #716, opened 2025-01-28
zfsbootmenu-core.sh: fix instant timed prompts

zdykstra:
bin/zfsbootmenu cheats and passes a timeout of 0 seconds, which is an instant timeout / success. Protect that value and restrict infinite timeouts to negative numbers entirely.

This fixes `zbm.skip`.

### DIFF
--- a/zfsbootmenu/lib/zfsbootmenu-core.sh
+++ b/zfsbootmenu/lib/zfsbootmenu-core.sh
@@ -1263,7 +1263,7 @@ has_resume_device() {
 }
 
 # getopts arguments:
-# -d  prompt countdown/delay; non-positive values will wait forever
+# -d  prompt countdown/delay; negative values will wait forever
 # -p  prompt with countdown timer (use %0.Xd as a placeholder for time)
 # -m+ message to be printed above the prompt, usable multiple times
 # -r  message to be prefixed with [RETURN] (accept)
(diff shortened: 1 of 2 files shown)


zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -1281,7 +1281,7 @@ timed_prompt() {
         delay="${OPTARG:-0}"
         if [ "${delay}" -gt 0 ] >/dev/null 2>&1; then
           :
-        elif [ "${delay}" -le 0 ] >/dev/null 2>&1; then
+        elif [ "${delay}" -lt 0 ] >/dev/null 2>&1; then
```
ahesford, line 1284: Maybe add an `elif [ "${delay}" -eq 0 ]` block to just `return 0` without printing a debug message.

Also, `s/non-positive/negative/` in the comment on line 1266.

fharbe:
I came across this issue because some of my machines would sit stuck forever at the menu / prompt with the new ZFSBootMenu v3.0.0 EFI binary. After investigating, I realized that I had configured `zbm.timeout=0` for those machines.

@zdykstra @ahesford I'd really appreciate a quick merge and release of updated EFI binaries, as this effects not only my setup but many others with the  `zbm.timeout` option set to `0` which causes the affected machines to hang forever at the menu / prompt.

ahesford:
I added a commit that incorporates my suggestions and also fixes the intentionally infinite prompt for initialization in `bin/zfsbootmenu`. Without that change, `bin/zfsbootmenu` will repeatedly try to initialize the bootloader if it is launched before the regular initialization sequence completes successfully. This can happen when launching zbm from a remote shell, either by winning a race or if initialization somehow fails to complete successfully. In either case, repeated rapid-fire initialization attempts are undesirable. Explicit input from the active user in these exceptional cases (and a prompt that explains the situation) is important to allow discover of configuration problems.

If you're happy with these fixes, squash and merge at will.

PR #720, opened 2025-01-30
sign-assets.sh: add manual passphrase input option

zdykstra:
If you store your passphrases somewhere else (Vaultwarden, etc), this provides a reasonable way to get them into the signing process.


ahesford:
How about dropping the `pass` invocation altogether? We can look for something like a non-empty `$ZBM_SIGN_PASS` and use that, or if it's empty, populate it from a prompt.

PR #722, opened 2025-02-05
aarch64 support: efi stub and devicetree selection

zdykstra:
generate-zbm now understands the `EFI.DeviceTree` key. This key can contain one of three possible values:

1) /absolute/path/%{kernel}/to/device-tree-file.dtb 
2) /absolute/path/to/device-tree-file.dtb
3) partial/path/device-tree-file.dtb

If found, %{kernel} is replaced with the kernel version being built in to the EFI asset. Partial paths are appended to `/boot/dtbs/dtb-$kernel/`, a well-known path for device tree files used by Void Linux.

Inside ZFSBootMenu, the `org.zfsbootmenu:devicetree` property is read. The values of this property are handled in the same order/priority as `EFI.DeviceTree`.

Documentation will be written and appended to this PR once the overall design of this support is accepted.

### DIFF
--- a/bin/generate-zbm
+++ b/bin/generate-zbm
@@ -15,6 +15,7 @@ use File::stat;
 use File::Path qw(make_path remove_tree);
 use File::Glob qw(:globally :nocase);
 use Sort::Versions;
+use Config;
 use bigint qw(hex);
 
 use Pod::Usage qw(pod2usage);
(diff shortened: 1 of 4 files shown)


zdykstra:
Go go gadget force push

ahesford (approved):
This works exactly as I'd expect. I offer only some trivial comments, take or leave them as you like.

We need documentation on the `config.yaml` options and the ZFS property.

classabbyamp (approved):
worked for me, pending docs ofc

ahesford:
In the fixup, I reworded the docs to replace sample paths with generic prose. If you think this improves clarity, squash it; if not, just stick with your original.

PR #725, opened 2025-02-16
docs: copy dropbear keys and config instead of symlink

aptalca:
In the container build, copy the contents of the dropbear folder into the expected folder instead of symlinking as the mkinitcpio function `add_full_dir` does not handle symlinked folders properly

Full discussion in https://github.com/zbm-dev/zfsbootmenu/discussions/724

Also fix quick typo: missing quote


zdykstra:
Thank you for the documentation improvements!

PR #733, opened 2025-03-23
docs: fix error on host-specific container build dropbear path

ckenna:
[The Container Build docs](https://docs.zfsbootmenu.org/en/latest/general/container-building/example.html#configuring-dropbear) page probably copied some code snippet from [this other page](https://docs.zfsbootmenu.org/en/latest/general/remote-access.html#configuring-dropbear) without updating the path for the container build.

The containerized build uses a slightly different directory. This makes a minor update to fix the code snippet that works on my machine.


aptalca:
Good call. I didn't notice it before because I skipped that step and let the container generate the keys during build.

PR #735, opened 2025-03-25
Add info on mkinitcpio-nfs-utils to container build example in docs

aptalca:
It is simpler to set up networking via dhcp with mkinitcpio-nfs-utils and this PR adds info on how to set it up in the container build example alongside the existing rclocal method.


ahesford:
This example is illustrative and intended to follow a specific path rather than provide alternatives. While the example may benefit from some clarifying edits, discussing other configuration options is better reserved for the more general-purpose pages and muddies the example.

PR #736, opened 2025-03-26
Add dhcp info to mkinitcpio-rclocal

aptalca:
Currently the instructions for the mkinitcpio-rclocal method for enabling remote access only contains static IP info.

This PR adds DHCP configuration instructions via dhclient to bring it more in line with the `Simple Host-Specific Container Build` instructions that suggest adding dhclient binaries and mentions DHCP: https://docs.zfsbootmenu.org/en/v3.0.x/general/container-building/example.html#configuring-basic-network-access


ahesford:
In 54bbde2b5e89119b9473a2e118f07fa46b373b35, I added mention of DHCP and the use of `dhclient` in the note about customizing `rc.local`, but don't think it merits breaks in flow with extra conditionals to document the matter further.

PR #741, opened 2025-04-16
zfsbootmenu: extend zbm.prefer to set bootfs

zdykstra:
Other bootloaders, such as `rEFInd`, can now be used to drive the boot process by specifying a dataset from which to boot. `zbm.prefer` now accepts a full dataset path, with the previous `!` and `!!` modifiers still applying to the pool portion of the dataset.

This closes https://github.com/zbm-dev/zfsbootmenu/discussions/738

### DIFF
--- a/zfsbootmenu/pre-init/zfsbootmenu-parse-commandline.sh
+++ b/zfsbootmenu/pre-init/zfsbootmenu-parse-commandline.sh
@@ -219,35 +219,41 @@ if kcl_override=$( get_zbm_arg zbm.kcl_override ) ; then
   zinfo "overriding all BE KCLs with: '$( kcl_assemble < "${BASE}/cmdline" )'"
 fi
 
-zbm_prefer_pool=
-if zbm_prefer_pool=$( get_zbm_arg zbm.prefer ) ; then
-  # shellcheck disable=SC2034
-  zbm_prefer_pool="${zbm_prefer_pool%%/*}"
-  zinfo "preferring ${zbm_prefer_pool} for bootfs"
-fi
-
 zbm_wait_for_devices=
 if zbm_wait_for_devices=$( get_zbm_arg zbm.wait_for ) ; then
   zinfo "system will wait for ${zbm_wait_for_devices}"
 fi
 
-# pool! : this pool must be imported before all others
-# pool!!: this pool, and no others, must be imported
+zbm_prefer_bootfs=
+zbm_require_pool=
+if zbm_prefer=$( get_zbm_arg zbm.prefer ) ; then
 
-# shellcheck disable=SC2034
-case "${zbm_prefer_pool}" in
-  *!!)
-    zbm_require_pool="only"
-    zbm_prefer_pool="${zbm_prefer_pool%!!}"
-    ;;
-  *!)
-    zbm_require_pool="yes"
-    zbm_prefer_pool="${zbm_prefer_pool%!}"
-    ;;
-  *)
-    zbm_require_pool=""
-    ;;
-esac
+  # strip the modifiers and set zbm_require_pool as needed
+  # shellcheck disable=SC2034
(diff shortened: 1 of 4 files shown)


zfsbootmenu/pre-init/zfsbootmenu-parse-commandline.sh
```diff
@@ -219,30 +219,40 @@ if kcl_override=$( get_zbm_arg zbm.kcl_override ) ; then
...
   *!!)
     zbm_require_pool="only"
     zbm_prefer_pool="${zbm_prefer_pool%!!}"
+    zinfo "will only attempt to import ${zbm_prefer_pool}"
+
+    if [ -n "${zbm_prefer_bootfs}" ] && [ "${zbm_prefer_pool}" != "${zbm_prefer_bootfs%%/*}" ]; then
+      zerror "zbm.bootfs '${zbm_prefer_bootfs}' not on the exclusive pool import"
+      unset zbm_prefer_bootfs
+    fi
+
```
ahesford, line 250: I wouldn't resolve this here and, more generally, wouldn't treat a missing `zbm.bootfs` as something to ignore. If somebody specifies a non-existent or improper filesystem for `zbm.bootfs`, the intent is unclear. This should trigger a drop to the menu rather than moving on to reading the `bootfs` pool property.

I think removing this block and relying on your existing changes in `50-import-pools` will achieve this effect.

ahesford (commented):
I wonder if it makes sense to extend the `zbm.prefer` syntax to accommodate this capability without an additional option. The argument to `zbm.prefer` could be the desired filesystem, with the pool and bootfs extracted directly. The only drawback is ambiguity when the user might wish to specify a `bootfs` pointing to the top-level filesystem. I think this would be an unlikely edge case; most sources don't recommend putting any content on the root-level filesystem. A trailing slash on a top-level filesystem could be used to disambiguate in this corner case (*e.g.*, `zbm.prefer=pool/`.

zdykstra:
> I wonder if it makes sense to extend the `zbm.prefer` syntax to accommodate this capability without an additional option. The argument to `zbm.prefer` could be the desired filesystem, with the pool and bootfs extracted directly. The only drawback is ambiguity when the user might wish to specify a `bootfs` pointing to the top-level filesystem. I think this would be an unlikely edge case; most sources don't recommend putting any content on the root-level filesystem. A trailing slash on a top-level filesystem could be used to disambiguate in this corner case (_e.g._, `zbm.prefer=pool/`.

Extending `zbm.prefer` to also accept a dataset is an interesting idea.  

- `zbm.prefer=zroot` - look to this pool for the `bootfs` value
- `zbm.prefer=zroot/ROOT/void` - explicitly use that as the `bootfs` value

- `zbm.prefer=zroot!` - make sure this pool is imported before proceeding, use the `bootfs` value set on the pool
- `zbm.prefer=zroot/ROOT/void!` - make sure this pool is imported before proceeding, explicitly use that as the `bootfs` value

- `zbm.prefer=zroot!!` - import only that pool, use the `bootfs` value set on the pool
- `zbm.prefer=zroot/ROOT/void!!` - import only that pool, explicitly use that value as the `bootfs` value

That seems like it could work, and wouldn't change any existing behavior of the `zbm.prefer` KCL option.

zfsbootmenu/pre-init/zfsbootmenu-parse-commandline.sh
zdykstra: no longer needed now that `zbm.prefer` has been extended

Muaytie666 (commented):
https://h20000.www2.hp.com/bizsupport/TechSupport/Document.jsp?objectID=c01732801

Muaytie666 (commented):
https://repo.cloudlinux.com"}/

PR #746, opened 2025-05-08
Install linux-lts-zfs-bin before modifying initramfs umask.conf

azulbeae:
/etc/initramfs-tools/conf.d/umask.conf doesn't exist until linux-stable-zfs-bin or linux-lts-zfs-bin are installed.

### DIFF
--- a/docs/guides/chimera/_include/zfs-config.rst
+++ b/docs/guides/chimera/_include/zfs-config.rst
@@ -1,6 +1,13 @@
 ZFS Configuration
 -----------------
 
+Install ZFS and kernel
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block::
+
+  apk add --no-interactive linux-lts-zfs-bin
+
 Configure initramfs-tools
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


zdykstra:
Thanks. Please see other documentation commits to construct a similar commit message.

zdykstra:
You've changed the title of the PR, not of the commit itself.

ahesford:
A better fix would be ensuring the directory exists rather than relying on some unrelated package to create it.

azulbeae:
Is anything else needed here? Like adding a tab for alternative package linux-stable-zfs-bin? These packages are just the linux kernel with zfs support, of lts and stable flavours.

ahesford:
Yes, drop the reliance on package installation order by just making the directory as I suggested.

azulbeae:
Is there a reason to add a mkdir besides preference? Why have more commands than necessary for installing?

ahesford:
Relying on the ordering of installation is fragile and may break in the future should the kernel package change its contents. If our guide requires that a directory exist, our guide should ensure that the directory exists.

azulbeae:
Should the kernel package changing its contents happen, won't that likely be the cause of /etc/initramfs-tools/conf.d/ ceasing to be the valid place for umask.conf, meaning breakage regardless?

ahesford:
No. The umask setting always belongs here, and is independent of what the kernel package installs. The matter is resolved, so there's no sense debating the right fix further.

PR #754, opened 2025-06-25
Update uefi.rst

psy0rz:
mention alpinebox


zdykstra:
It'd be okay to add it to https://docs.zfsbootmenu.org/en/v3.0.x/guides/third-party.html - but the commit message would need to match our style.

PR #756, opened 2025-07-09
docs/guides/: recommend openzfs-2.2-linux pool compatibility

zdykstra:
With OpenZFS 2.2 as the current LTS, lets set our minimum pool feature set to that release. All distros we have guides for should be on at least that version by now.


classabbyamp (approved):
:shipit:

PR #757, opened 2025-07-12
AlmaLinux 9

FingerlessGlov3s:
I've gone through and made a copy of the fedora guide but for AlmaLinux 9, I would also added AlmaLinux 10 but the ZFS RPMs aren't ready yet, but I think the process will effectively the same.

I've tested
 - Unencrypted
 - Encrypted
 - Prebuilt ZFSBootMenu
 - Compiled ZFSBootMenu

### DIFF
--- a/docs/guides/almalinux/_include/alma-config.rst
+++ b/docs/guides/almalinux/_include/alma-config.rst
@@ -1,0 +1,25 @@
+Configure Alma
+--------------------------
+
+Set hostname
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block::
+
+  hostname AlmaRocks
+  hostname > /etc/hostname
+
+Fix SELinux filesystem labels
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block::
+
+  fixfiles -F -f relabel
+
+Create initial admin user
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block::
+
+  adduser -G wheel fred
+  passwd fred
--- a/docs/guides/almalinux/_include/configure-gen-zbm.rst
+++ b/docs/guides/almalinux/_include/configure-gen-zbm.rst
@@ -1,0 +1,17 @@
+Configure :doc:`generate-zbm(5) </man/generate-zbm.5>` by ensuring that the following keys appear in
+``/etc/zfsbootmenu/config.yaml``:
+
+.. code-block:: yaml
+
+  Global:
+    ManageImages: true
+    BootMountPoint: /boot/efi
+  Components:
+     Enabled: false
+     Versions: false
+  EFI:
+    ImageDir: /boot/efi/EFI/zbm
+    Enabled: true
+  Kernel:
+    CommandLine: quiet loglevel=0
+    Version: "*.x86_64"
--- a/docs/guides/almalinux/_include/zbm-install.rst
+++ b/docs/guides/almalinux/_include/zbm-install.rst
@@ -1,0 +1,35 @@
+Install ZFSBootMenu
+~~~~~~~~~~~~~~~~~~~
+
+.. tabs::
+
+  .. group-tab:: Prebuilt
+
+    .. include:: ../_include/zbm-install-prebuilt.rst
+
+  .. group-tab:: Source
+
+    .. include:: _include/zbm-install-deps.rst
+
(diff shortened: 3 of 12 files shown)


docs/guides/almalinux/_include/zfs-config.rst
```diff
@@ -0,0 +1,58 @@
...
+
+  dnf install -y https://zfsonlinux.org/epel/zfs-release-2-3$(rpm --eval "%{dist}").noarch.rpm
+  dnf install -y epel-release
+  dnf install -y zfs zfs-dracut sudo console-setup efibootmgr langpacks-en dosfstools
+  dnf install -y kernel kernel-devel
+
+
+.. note::
+
+  Ignore missing kernel messages, we haven't installed the kernel yet, when we do install the kernel after ZFS, dkms will install the module as it should.
```
zdykstra, line 45: This can be simplified to something like `Missing kernel messages can be ignored during this step.`

docs/guides/almalinux/_include/zfs-config.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Regenerate initramfs
+~~~~~~~~~~~~~~~~~~~~
+
+.. code-block::
```
zdykstra, line 50: Will `dracut --force --regenerate-all` not work? Avoiding having to type in a long kernel version string would be good. The `note` can also be removed entirely.

docs/guides/almalinux/_include/configure-gen-zbm.rst
zdykstra, line 1: This seems to be a dead file that's not included by anything. There's an identical block in [docs/guides/almalinux/_include/zbm-install.rst](https://github.com/zbm-dev/zfsbootmenu/pull/757/files#diff-e8e4556007d0c767c80887eb17131b7736e17c34d56b7ca1a854bee6aa91e32d)

docs/guides/almalinux/_include/alma-config.rst
```diff
@@ -0,0 +1,25 @@
...
+  hostname > /etc/hostname
+
+Fix SELinux filesystem labels
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block::
+
+  fixfiles -F -f relabel
+
+Create initial admin user
```
zdykstra, line 19: User creation is typically left as an exercise to the end user. Are root logins entirely disabled, such that even setting a password for that account won't let you login?

docs/guides/almalinux/_include/zbm-install.rst
```diff
@@ -0,0 +1,35 @@
...
+            BootMountPoint: /boot/efi
+        Components:
+            Enabled: false
+            Versions: false
+        EFI:
+            ImageDir: /boot/efi/EFI/ZBM
+            Enabled: true
+        Kernel:
+            CommandLine: quiet loglevel=0
+            Version: "*.x86_64"
```
zdykstra, line 32: Why does the `Version` key need to be specified? Is `generate-zbm` doing the wrong thing by default on Alma?

zdykstra (changes requested):

docs/guides/almalinux/_include/zfs-config.rst
FingerlessGlov3s: It errors trying to build for the +debug kernel which isn't installed, but ZFS package installs apart of it.

docs/guides/almalinux/_include/zfs-config.rst
FingerlessGlov3s: Please see new commit

docs/guides/almalinux/_include/alma-config.rst
```diff
@@ -1,6 +1,24 @@
...
+
 .. code-block::
 
   fixfiles -F -f relabel
+
+Set a root password
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block::
+=
```
zdykstra, line 23: Typo here - `passwd` isn't in the code-block.

FingerlessGlov3s:
I've now consolidated the guide down as proposed. Thoughts?

zdykstra:
I've done a quick run-through of the guide following the easy installation path (no encryption, pre-built EFI binaries) and everything works as expected. If you can rebase this branch on master / fix the conflicts with `docs/index.rst` this is ready for a merge!

zdykstra (approved):

PR #758, opened 2025-07-15
docs/guides: collapse layers of indirection for single-purpose guides

ahesford:
I brought this up in #757. Let's drop the layers of indirection for distributions with only single guides, and add some redirects to avoid breaking links.

I've also removed version numbers from page titles, but we can add them back (or find another way to clearly note the applicable version at the start of each document) if desired.


zdykstra:
:+1:

PR #768, opened 2025-08-21
docs/guides/debian: Debian Trixie

FingerlessGlov3s:
Debian Trixie tested for prebuilt ZBM and compiling from source using Debian Trixie


FingerlessGlov3s:
Changes made, also just made `/etc/apt/sources.list` effectively mirror default file, and added contrib too it.

zdykstra:
Thanks!

PR #773, opened 2025-09-10
zfsbootmenu: add and use recursive_umount function

zdykstra:
Add a `recursive_umount` function and use in the chroot helper and when unmounting filesysems just before kexec. This is in place for https://github.com/zbm-dev/zfsbootmenu/discussions/770 .


zfsbootmenu/lib/zfsbootmenu-core.sh
```diff
@@ -2272,3 +2272,40 @@ is_zfs_filesystem() {
...
+  local mounts mountpoint mp skip mp_re depth filesystem ret
+
+  mountpoint="${1}"
+  if [ -z "${mountpoint}" ]; then
+    zerror "mountpoint undefined"
+    return 1
+  fi
+
+  ret=0
+  if ! umount -R "${mountpoint}" >/dev/null 2>&1 ; then
```
ahesford:
```suggestion
  umount -R "${mountpoint}" >/dev/null 2>&1 && return
  
  ret=0
```
Then you can drop the big `if` indent.

PR #774, opened 2025-09-18
dracut: exclude clevis due to missing systemd module

zipcode:
If clevis is available, dracut will add it to the image. However, as dracut-systemd is already omitted, clevis will crash complaining of a missing environment variable.

### DIFF
--- a/etc/zfsbootmenu/dracut.conf.d/zfsbootmenu.conf
+++ b/etc/zfsbootmenu/dracut.conf.d/zfsbootmenu.conf
@@ -1,3 +1,3 @@
 nofsck="yes"
 add_dracutmodules+=" zfsbootmenu bash "
-omit_dracutmodules+=" btrfs zfs resume systemd systemd-initrd dracut-systemd plymouth dash "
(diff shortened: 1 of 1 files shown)


zipcode:
Huh, I guess I should have checked if upstream could do that

zdykstra:
I guess `clevis` provides their own dracut module - looks like this might be the setup script for it: https://github.com/latchset/clevis/blob/master/src/luks/dracut/clevis/module-setup.sh.in

zipcode:
Okay, it looks like in theory this should happen [here](https://github.com/latchset/clevis/blob/2ad7027b7182b8f46404f8ec0735a83738f7ca7f/src/luks/dracut/clevis/module-setup.sh.in#L23) but it's not doing the right thing.

zdykstra:
If you want change the short commit message to something along the lines of:

> dracut: exclude clevis due to missing systemd module

We can merge that in, as I don't expect the clevis project to get a meaningful fix out in a reasonable time frame.

zipcode:
This could be a wacky interaction with my distribution (void), but I think I'm done digging into it here.

PR #785, opened 2025-11-15
zfsbootmenu-core.sh: do not cache unencrypted keysources

ahesford:
This needs testing.

I can't think of a single legitimate use for storing ZFS encryption keys on an unencrypted ZFS filesystem. People can do it if they want, but we don't need to help them make bad choices by copying their unencrypted keys via our caching mechanism.


ahesford:
Moved from `zdebug` to `zwarn`.

zdykstra:
Tested locally with `org.zfsbootmenu:keysource` pointing to an un-encrypted dataset; works as expected. No keys are cached and the main menu forces a passphrase input.

PR #800, opened 2026-01-06
feat: Add remote SSH access with RFC 3442 DHCP fix and SSH timeout

terem42:
This commit extends remote SSH access support for ZFSBootMenu and fixes some DHCP bugs.

New Features:
- SSH timeout support (zbm.ssh_timeout=N): Wait N seconds for SSH login before auto-boot
- Network readiness check before SSH timeout starts
- Race condition prevention: Console waits when SSH user is connected
- Dropbear SSH on port 22 (configurable)

RFC 3442 DHCP Fix:
- Patched dhclient-script with robust parse_option_121() function
- Fixes 'integer expression expected' and 'shift count out of range' errors
- Works correctly with Hetzner and other providers using classless static routes
- Uses dracut module prefix 30 to load before 35network-legacy


terem42:
Unfortunately, the very nature of DHCP bug in dracut and combined nature of zbm being able to be accessible simultaneosly via KVM console and SSH leave not many other choices.

Thanks for your feedback, I will keep this fork anyway, it's proved itself very useful for situations where one need to removely manage systems where KVM console is not available.

classabbyamp:
if it's a bug in dracut, make an issue/PR to https://github.com/dracut-ng/dracut-ng

PR #806, opened 2026-02-03
docs: Fix nested list syntax in container-building

noelleleigh:
In reStructuredText markup, nested lists are created by aligning the bullet/enumerator with the left edge of the text in the list item above it. For [enumerated lists][0], that means an indent of at least **3 spaces** is required, otherwise the indent is treated as a [block quote][1].

This PR fixes the indentation of emphemeral container steps in the `zbm-builder.sh` behavior description in
`docs/general/container-building.rst`, which removes the (presumably) unintended block quote.

[0]: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#enumerated-lists
[1]: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#block-quotes


zdykstra:
Good catch, thank you!

PR #808, opened 2026-02-05
zbm-builder.sh: Use -d for custom container CLI

noelleleigh:
This PR changes the `-d` argument of `zbm-builder.sh` to accept an executable argument, which will be used in place of `podman` or `docker`. The container-building docs have been updated to mention the ability to use a `docker`-compatible CLI.

For example, to run the script using [nerdctl][0]:

```shell
./zbm-builder.sh -d nerdctl
```

The choice of container CLI follows this priority list:

1. `-d` argument
2. `PODMAN` variable in `zbm-builder.conf`
3. `podman`
4. `docker`

Specifying the `PODMAN` variable with `env` isn't supported, because the value would be overwritten by the value from `zbm-builder.conf` when it gets sourced.

[0]: https://github.com/containerd/nerdctl

### DIFF
--- a/docs/general/container-building.rst
+++ b/docs/general/container-building.rst
@@ -37,8 +37,8 @@ Dependencies
 ------------
 
 To build ZFSBootMenu images from a build container, either `podman <https://podman.io>`_ or
-`docker <https://www.docker.com>`_ is required. The development team prefers ``podman``, but ``docker`` may generally be
-substituted without consequence.
+`docker <https://www.docker.com>`_ is required. The development team prefers ``podman``, but ``docker`` or a
+compatible container utility may generally be substituted without consequence.
 
 If a custom build container is desired, `buildah <https://buildah.io>`_ and ``podman`` are generally required. A
 :zbm:`Dockerfile <releng/docker/Dockerfile>` is provided for convenience, but feature parity with the ``buildah``
--- a/zbm-builder.sh
+++ b/zbm-builder.sh
@@ -38,7 +38,9 @@ OPTIONS:
      Specify the path to a configuration file that will be sourced
      (Default: \${BUILD_DIRECTORY}/zbm-builder.conf, if it exists)
 
-  -d Force use of docker instead of podman
+  -d <container-frontend>
+     Use the specified front-end utility to launch the build container
+     (Default: podman if available, docker otherwise)
 
   -M <argument>
      Provide a comma-separated list of options to use for volume
@@ -119,7 +121,7 @@ else
   PODMAN="docker"
 fi
 
-CMDOPTS="b:dhi:l:L:c:M:O:HR"
+CMDOPTS="b:d:hi:l:L:c:M:O:HR"
 
 # First pass to get build directory and configuration file
 while getopts "${CMDOPTS}" opt; do
@@ -166,7 +168,7 @@ while getopts "${CMDOPTS}" opt; do
     b|c|h)
       ;;
     d)
-      PODMAN=docker
+      PODMAN="${OPTARG}"
       ;;
     i)
       BUILD_IMG="${OPTARG}"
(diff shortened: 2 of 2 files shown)


docs/general/container-building.rst
```diff
@@ -37,8 +37,8 @@ Dependencies
 ------------
 
 To build ZFSBootMenu images from a build container, either `podman <https://podman.io>`_ or
-`docker <https://www.docker.com>`_ is required. The development team prefers ``podman``, but ``docker`` may generally be
-substituted without consequence.
+`docker <https://www.docker.com>`_ is required. The development team prefers ``podman``, but ``docker`` (or other
+compatible container CLI) may generally be substituted without consequence.
```
ahesford:
```suggestion
`docker <https://www.docker.com>`_ is required. The development team prefers ``podman``, but ``docker`` or a
compatible container utility may generally be substituted without consequence.
```

zbm-builder.sh
```diff
@@ -38,7 +38,8 @@ OPTIONS:
      Specify the path to a configuration file that will be sourced
      (Default: \${BUILD_DIRECTORY}/zbm-builder.conf, if it exists)
 
-  -d Force use of docker instead of podman
+  -d <executable>
+     Force the use of a container CLI other than podman or docker
```
ahesford:
```suggestion
  -d <container-frontend>
     Use the specified front-end utility to launch the build container
     (Default: podman if available, docker otherwise)
```

zbm-builder.sh
```diff
@@ -202,7 +203,7 @@ done
 shift $((OPTIND-1))
 
 if ! command -v "${PODMAN}" >/dev/null 2>&1; then
-  echo "ERROR: this script requires podman or docker"
+  echo "ERROR: '${PODMAN}' not found. Pass -d to specify a different container CLI"
```
ahesford:
```suggestion
  echo "ERROR: ${PODMAN} not found; use '-d' to specify a container front-end"
```

ahesford (changes requested):
I'm vacillating between wanting to see this change respect `$PODMAN` in the environment and not really caring. We could guard the [current default assignment](https://github.com/zbm-dev/zfsbootmenu/blob/master/zbm-builder.sh#L116-L120) with `if [ -z "${PODMAN}" ]` and move it right after the configuration file is sourced (and right before the second option parsing) and get this for free. On the other hand, one could just do

    zbm-builder.sh -d "${PODMAN}"

and be done.

If you are inclined to make this change, I'll accept it; if not, I won't bother adding it independently.

noelleleigh:
> [...] We could guard the [current default assignment](https://github.com/zbm-dev/zfsbootmenu/blob/master/zbm-builder.sh#L116-L120) with `if [ -z "${PODMAN}" ]` and move it right after the configuration file is sourced (and right before the second option parsing) and get this for free.

I tried this while implementing this feature, and it didn't work the way I wanted. To me, this is the ideal order of precedence for controlling this setting:

1. `-d nerdctl`
2. `env PODMAN=nerdctl`
3. `PODMAN=nerdctl` in `zbm-builder.conf`
4. Search for `podman`
5. Search for `docker`

However, to allow `env PODMAN` to override the contents of `zbm-builder.conf`, the environment variable would need to be captured _before_ the config is sourced, then _re-applied afterwards_. I found possible solutions in [this StackOverflow question](https://stackoverflow.com/questions/69127550/is-it-possible-to-make-the-bash-source-command-keep-environment-variables-pass), but at that point it felt overengineered, especially for a single variable, and I wasn't sure what side effects it might have, so I decided to scrap the idea.

Given that, do you still think it's worth adding the extra logic required for supporting `env PODMAN` precedence?

ahesford:
I'm sorry, this slipped past me. I agree that it isn't worthwhile to delve further into supporting the environment variable. The changes are fine as they are now; please squash all the commits down (you can leave out the explanatory text in favor of a one-liner commit message).

noelleleigh:
> I'm sorry, this slipped past me. I agree that it isn't worthwhile to delve further into supporting the environment variable. The changes are fine as they are now; please squash all the commits down (you can leave out the explanatory text in favor of a one-liner commit message).

Done :+1: https://github.com/zbm-dev/zfsbootmenu/pull/808/commits/f62b24722bb8a97468039bacd5eed5f2d7a05a13

PR #809, opened 2026-02-19
docs: Fix default UEFI Executable location

yushisora:
With Commit 9c91afa, the default UEFI executable should be located at `/boot/efi/EFI/zbm` instead of `/boot/efi/EFI/void`.

Also a minor grammatical fix for native encryption documentation.


zdykstra:
Good catch, thank you.

PR #817, opened 2026-06-18
Make it possible to replace sed

rbgarga:
On systems like Chimera Linux, GNU sed is available as gsed.  Make it
possible to set an environment variable and use it instead of sed.

No functional changes intended

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ INITCPIODIR=$(PREFIX)/lib/initcpio
 MANDIR=$(PREFIX)/share/man
 BINDIR=$(PREFIX)/bin
 EXAMPLES=$(PREFIX)/share/examples/zfsbootmenu
+SED?=sed
+
+export SED
 
 .PHONY: install core dracut initcpio zbm-release show-version
 
--- a/releng/tag-release.sh
+++ b/releng/tag-release.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 # vim: softtabstop=2 shiftwidth=2 expandtab
 
+: ${SED:=sed}
+
 error () {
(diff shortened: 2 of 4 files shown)


ahesford:
Does the BSD sed misbehave?

I don't generally like constructs like this because it makes things less readable. The purpose of things like POSIX and the Single UNIX Specification is to define interfaces that allow us to not care which provider sits behind them. If we are using GNU-specific extensions, then either we should fix that or make an accommodation like this. Otherwise, if you just want to accommodate a personal preference, symlink `/some/path/to/sed -> /usr/bin/gsed` and set your search path accordingly.

rbgarga:
> Does the BSD sed misbehave?
> 
> I don't generally like constructs like this because it makes things less readable. The purpose of things like POSIX and the Single UNIX Specification is to define interfaces that allow us to not care which provider sits behind them. If we are using GNU-specific extensions, then either we should fix that or make an accommodation like this. Otherwise, if you just want to accommodate a personal preference, symlink `/some/path/to/sed -> /usr/bin/gsed` and set your search path accordingly.

The common issue is related to `-i` parameter.  On BSD sed, -i requires to get a parameter with the suffix that will be added to backup file, if you don't want a backup file you must pass an empty string `-i ''` and that syntax doesn't work with GNU sed.  I also see the following error that I didn't investigate

```
$ make core
./releng/version.sh -u
Updating zfsbootmenu/zbm-release
Updating bin/generate-zbm
sed: -I or -i may not be used with stdin
```

rbgarga:
@ahesford oh and BTW, I agree with you that adjusting all sed calls to be agnostic and work on both sed implementations would be the best option

PR #824, opened 2026-08-09
bin/generate-zbm: fix globbing in versionedKernel

ahesford:
If a specific kernel version is requested on aarch64, where the kernel is named `vmlinux` rather than `vmlinuz`, the `glob` function returns, *e.g.*,

    /boot/vmlinuz-6.18.43_1

even when that file doesn't exist. This shorts the loop over other potential prefixes and returns an invalid result. Rather than grabbing the first glob result blindly, it's better to verify that *some* valid result exists in this list, or move on to the next prefix.


zdykstra:
Good catch. 👍

